PR #154, opened 2015-02-11
just making some japanese titles more natural  

everpeace:
"<some noun>をはじめましょう" is much more natural in Japanese than "<some noun>はじめましょう"

this relates to SkymindIO/nd4j#18


chrisvnicholson:
thanks!

Issue #225, opened 2015-04-12
RNTN fit function does not work


ts335793:
getValueGradient function in RNTN does not calculate gradient. Code inside if statements in lines 933 - 967 is never reached.

Fit function creates threads and does not wait for them to finish. It also does not return any thread handlers.

Possible solution:
https://github.com/ts335793/deeplearning4j/commit/617754401f34c24b30c6e631151fb20c71939b70

agibsonccc:
Thanks for the headsup. Would you mind submitting a pull request? I can work on it a bit more there.

Bam4d:
This issue is also causing me some troubles. I'll get a PR done ASAP when I understand whats going on better

Bam4d:
51065  [ka.actor.default-dispatcher-26] WARN  Parallelization           - Error occurred processing data 
java.lang.IllegalStateException: Given ndarray does not have data type double
	at org.nd4j.linalg.factory.DataTypeValidation.assertDouble(DataTypeValidation.java:40) ~[nd4j-api-0.0.3.5.5.3-SNAPSHOT.jar:na]

Getting this error now... any ideas why ?

agibsonccc:
Your data types need to line up. For now at the beginning of your program just put:

Nd4j.dtype = DataBuffer.FLOAT;

The root cause of the problem is arrays being created that are inconsistent with the data type for nd4j.

Bam4d:
The offending line that causes the above exception is: ``` Double nodeWeight = classWeights.get(goldClass); ```

if I set dtype to FLOAT then ```nodeWeight``` is still a double and causes the same exception...

Bam4d:
changing nodeWeight to a float and forcing various things to be floats I then start getting this error:

```
o.d.p.Parallelization - Error occurred processing data
org.jblas.exceptions.LapackException: LAPACK Java.raxpy: Parameters for x aren't valid! (n = 10000, dx.length = 500000, dxIdx = 0, incx = 5000)
	at org.jblas.JavaBlas.raxpy(JavaBlas.java:213) ~[jblas-1.2.3.jar:na]
	at org.nd4j.linalg.jblas.BlasWrapper.axpy(BlasWrapper.java:195) ~[nd4j-jblas-0.0.3.5.5.3-SNAPSHOT.jar:na]
[...]

PR #227, opened 2015-04-14
225: have two options for fitting, one that is async and one not.

Bam4d:
user can choose how to process. and a second blocking call that implements a countdown latch for each batch. fixes parts of #225


Bam4d:
Incorrect implementation, please ignore

agibsonccc:
Updates on this? I'm getting ready for a release here in the next week or so (the last feature I need is opencl support in nd4j) and checking some of the other models. It'd be great to get this mainlined.

Bam4d:
As far as I'm concerned the implementation works for me, I don't know why the build is failing however.

If you're happy with it go for it.

Bam4d:
The only thing I have to say is that I couldn't test it working with GPU due to the bug I posted in the google group https://groups.google.com/forum/#!topic/deeplearning4j/0m5s4Qkjc-I. I wouldn't be related to this PR however.

PR #233, opened 2015-05-03
configuration serialization - iteration listener

EronWright:
Here is the next step in the series of improvements to the fidelity of NeuralNetworkConfiguration serialization.   The iteration listener is removed from the configuration object altogether.   Listeners are rather provided during layer construction (i.e. to the set of objects concerned with executing a network).

If we imagine a network configuration shared between any two applications, it cannot be assumed that the listeners on one side would be the same as those on the other side.    One might also argue that a configuration with executable elements is less safe security-wise (like a Word doc with scripts).


deeplearning4j-core/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
```diff
@@ -92,6 +93,7 @@ public MultiLayerNetwork(MultiLayerConfiguration conf) {
      * @param conf   the configuration json
      * @param params the parameters
      */
+    @Deprecated
```
EronWright: I marked these deprecated because they offer no opportunity to add a listener.   Listeners must be added before init().

PR #235, opened 2015-05-04
Break up convolution and subsampling layers.

agibsonccc:


deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -79,9 +79,11 @@
     //whether to constrain the gradient to unit norm or not
     protected boolean constrainGradientToUnitNorm = false;
     /* RNG for sampling. */
-    protected Random rng;
+    protected long seed = 123;
```
EronWright: stuff is regressing...

agibsonccc: I haven't merged your PR yet.
ᐧ

PR #236, opened 2015-05-05
Undo regressions from previous merge

EronWright:


EronWright:
Note that the listener change is already on master.  My justification from the original PR:    

The iteration listener is removed from the configuration object altogether. Listeners are rather provided during layer construction (i.e. to the set of objects concerned with executing a network).

If we imagine a network configuration shared between any two applications, it cannot be assumed that the listeners on one side would be the same as those on the other side. One might also argue that a configuration with executable elements is less safe security-wise (like a Word doc with scripts).

EronWright:
The overall idea of this series of improvements is to cleanly separate the 'configuration' and 'execution' elements, based on the insight that network configuration is shared across the cluster and between applications, whereas the execution elements are local to a node.   Without this you're needlessly forced to serialize execution elements or make them transient.

Secondly it allows the api to be factored out of the implementation.

agibsonccc:
I can't argue with that :(. I'm more in favor of merging this. Let me draft up a migration email. This is more than just one or 2 changes.

EronWright:
Since you merged the newmultilayer branch already, please merge this into newmultilayer then immediately merge to master.  Master is presently in an unusable state.

PR #248, opened 2015-05-13
custom loss functions

agibsonccc:
Allow users to create custom loss functions.


EronWright:
ServiceLoader.

agibsonccc:
Feel free to submit a POC for this. I'm going to do this via reflections. Pulling in all classes that extend BaseAccumulation on the classpath is trivial and doesn't require the user to know java.

The service loader makes sense for you and I not as a user level thing though.

EronWright:
I played with the code while on the bus yesterday.   Adding a proper extension framework to DL4J would take some planning but the outcome could be awesome!

The main impediment I encountered is the lack of a context object (like ObjectMapper in Jackson or Environment in Dropwizard), to which an extension point could be registered.   I hesitated to make registration be static.

Long story short I opened issue #251.

EronWright:
OOPS I clicked Close when I meant Comment.   My bad!

PR #272, opened 2015-05-21
cli improvements

agibsonccc:
Command Line Interface improvements: Actually trains a model based on the config


deeplearning4j-cli/deeplearning4j-cli-api/src/main/java/org/deeplearning4j/cli/subcommands/Train.java
```diff
@@ -59,294 +61,229 @@
...
+            e.printStackTrace();
+        }
+        try {
+            this.configProps.load(in);
+            in.close();
+        } catch (IOException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
+        }
+
```
smarthi: Replace this entire try {} by 
{Code}

try (InputStream in = new FileInputStream(this.configurationFile)) {
  this.configProps.load(in);
}

agibsonccc: @smarthi Send me a PR. You're right though. That's not habit for me yet.

deeplearning4j-cli/deeplearning4j-cli-api/src/main/java/org/deeplearning4j/cli/subcommands/Train.java
```diff
@@ -59,294 +61,229 @@
...
-
-		while (e.hasMoreElements()) {
-			String key = (String) e.nextElement();
-			log.warn(key + " - " + props.getProperty(key));
-		}
-
-		log.warn("-- DL4J Configuration --\n");
-	}
-
+        if ( null != this.configProps.get( INPUT_DATA_FILENAME_KEY ))
```
smarthi: I think it would be better to replace references like this with guava equivalents String.isNullOrEmpty(<arg>). Will make a PR

Issue #259, opened 2015-05-19
NeuralNet serialization exception


raver119:
On current stable release of dl4j and nd4j there's an issue: serialization called on MultiLayerNetwork causes serialization exception:

Call to serialization: 

SerializationUtils.saveObject(d,new File("test.neuro"));

Exception:

Exception in thread "main" java.lang.reflect.InvocationTargetException
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at com.intellij.rt.execution.CommandLineWrapper.main(CommandLineWrapper.java:130)
Caused by: java.lang.RuntimeException: java.io.NotSerializableException: org.nd4j.linalg.api.rng.DefaultRandom
    at org.deeplearning4j.util.SerializationUtils.saveObject(SerializationUtils.java:110)
    at org.hive.cloud.neurality.networks.POSNetwork.main(POSNetwork.java:67)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at com.intellij.rt.execution.application.AppMain.main(AppMain.java:140)
    ... 5 more
Caused by: java.io.NotSerializableException: org.nd4j.linalg.api.rng.DefaultRandom
    at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
    at java.io.ObjectOutputStream.defaultWriteFields(ObjectOutputStream.java:1548)
    at java.io.ObjectOutputStream.writeSerialData(ObjectOutputStream.java:1509)
    at java.io.ObjectOutputStream.writeOrdinaryObject(ObjectOutputStream.java:1432)
    at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1178)
    at java.io.ObjectOutputStream.writeArray(ObjectOutputStream.java:1378)
[...]

PR #279, opened 2015-05-23
undo regression from d27f640264a650a4b59c50a53de64c0e50fb0eee

EronWright:


EronWright:
You misunderstand, the RNG object is from conf.rng, not from ND4J.    It is fully serializable and JSON friendly...that'y why we changed the API a couple weeks ago.

agibsonccc:
Ooh got it. I wonder why we were having serde issue with the conf? I'll take a closer look at that.

EronWright:
I don't think there is a serde issue with the conf.  I believe that bug report #259 motivated this change, and the user was attempting to serialize the MultiLayerNetwork itself (which does have issues).

PR #280, opened 2015-05-25
Smaller/faster Word2VecTests added.

carschno:
Adding some unit tests for Word2Vec for testing nearestWords() on previously built models.


carschno:
The test suit fails, but I actually think that there is something wrong with the LookupTable rather than with these tests.

PR #291, opened 2015-05-26
Fixes WeightInitTests error. Fixing bugs in other tests.

sonaliii:


smarthi:
u may want to squash ur commits, before merging this into trunk

nyghtowl:
@smarthi agree that squash is good to use and thanks for the input. We are working to keep that as a best practice when we can. This PR is a work in progress right now and since there are more than one of us on it that will create limits to our ability to squash. We'll definitely keep it in our pocket and tackle it when we can.

PR #316, opened 2015-06-05
dl4j-spark-ml

EronWright:
I think this feature branch is ready to merge.   Introduces a new module called dl4j-spark-ml.  

The module depends on Spark 1.4.0 which is not yet available in Maven Central.  I introduced a Maven profile ("spark-1.4") to selectively activate the module.   To compile the module, you must have Spark 1.4RC4 installed into your local Maven repository.   In a few days I believe that 1.4.0 will be announced, and the special-case shall be removed.

Note that the module does not inherit the parent 'spark' module (which is fixed at Spark 1.2).


agibsonccc:
I'll wait till its in maven central then

PR #341, opened 2015-06-15
use new implicit conversion methods instead of MLLib

EronWright:
I remember now why I placed a .java file into main/scala - intellij doesn't import mixed projects properly (if both main/java and main/scala exist, the latter is ignored).   Placing .java in main/scala works.

I simply deleted MLLib and switched to using implicit conversion methods I authored earlier.


agibsonccc:
That's all I know how to build with. Maven supports mixed projects just fine. A bit dubious there.

EronWright:
Maven is fine, intellij is the problem.  I meant to use the implicit conversions anyway, so this was an inevitable change.

Sent from my phone.
________________________________

agibsonccc:
That's fine. And what I'm saying is intellij with the right maven plugins figures it out for me. I'll go ahead and merge this but would still love to see how you setup your env.

Issue #355, opened 2015-06-18
Backpropagation algorithm implementation


AlexDBlack:
OK, so I'm going through the back-prop implementation, which is implemented in MultiLayerNetwork.doBackward() and BaseLayer.backwardGradient().
There are a number of significant issues here.

First I'm going to define notation and the back-prop algorithm for the sake of discussion/debugging/my own sanity:

Denote:
![](http://mathurl.com/q8nt7yw.png) as weight from unit i (prev. layer) to unit j (output layer).
![](http://mathurl.com/nr9pktp.png) is input to unit j
![](http://mathurl.com/osglszn.png) is target/label
![eqn](http://mathurl.com/oopsolc.png) is output
L loss function
L (superscript): layer number.
![eqn](http://mathurl.com/ojcy77o.png) is element-wise nonlinearity/activation function
![](http://mathurl.com/ooyn3vk.png) is the activation of the jth unit in layer L. Note that ![](http://mathurl.com/pl79ogo.png)

For the output layer, define the error/delta:
![](http://mathurl.com/ngokjpv.png)
(See https://github.com/deeplearning4j/deeplearning4j/issues/349)

For lower layers, we have (with ![](http://mathurl.com/nq6x75r.png) being element-wise multiplication):
![](http://mathurl.com/ovapmc2.png)

(References for above equations:)
(1) http://neuralnetworksanddeeplearning.com/chap2.html (eq. 29, BP1, BP2)
(2) Bishop (1995) Neural Networks for Pattern Recognition (pg144/eq4.36)

-----
-----
MultiLayerNetwork.doBackWard(): Javadoc: "Calculate the gradient relative to the error in the next layer"
Given the context and how the results are used, I'm going to assume this means calculate ![](http://mathurl.com/qx7u6px.png) given ![](http://mathurl.com/on7zv3k.png)

The error/delta/ix for the output layer
```java
INDArray ixInitial = labels.sub(outputActivate)
                    .subi(Nd4j.getExecutioner()
                            .execAndReturn(Nd4j.getOpFactory()
                                    .createTransform(getOutputLayer().conf().getActivationFunction(),outputActivate).derivative()));
```
should simply be:
```java
INDArray ixInitial = outputActivate.sub(labels);
```

Furthermore,
```for(int j = getnLayers() - 1; j >= 0; j--) { ```
becomes
```for(int j = getnLayers() - 2; j >= 0; j--) {```
[...]

PR #366, opened 2015-06-24
Mw gradient fix

nyghtowl:
Update and main fixes:
- Added conditional for how outputlayer gradient is calculated based on loss function. Opening enhancement to move this back into the Outputlayer gradient at a later date.
- There is a derivative calculation error that is coming from ND4J. Issue currently logged there
- Changed code in feedforward to return a pair of activations and derivatives. The original code was returning zs and applying derivative later. There were direct reference and order of storing the data issues that were throwing off calculations. Key fix was to ensure the derivative was provided for the output layer gradient calculation if it is needed.
- Changed BaseLayer update to sum the gradient vs. take the mean
- Condensed gradient adjustment code and simplified multiple calls into the same function in several places. Also shifted some loops and parameter passing in the gradient adjustment and update
- MLP test fixes - setting configuration outputlayer to softmax and taking derivative on post sigmoid activation.
- Made lots of name changes for clarification

Merging code and will open new branches for other enhancements and fixes.

Original:
@AlexBlack
Heads up I've started making some changes to the backprop calculations. I added back in the application of derivatives to the first delta that is calculated to align to the equations at http://neuralnetworksanddeeplearning.com/chap2.html. Also, I adjusted how the derivatives are tracked and applied. This branch is a work in progress and I'm still digging through to confirm approach. Sharing in case you are digging into it now.

One modification on your test is to change the output layer so it initializes values for the weights. This is included on this branch. If you have questions or additional input please send our way.


AlexDBlack:
So I cloned the mw_gradient_fix branch, but the unit tests aren't passing. (I think I cloned everything correctly, looking at the code. But maybe not?). TestMLP and TestMLP2 fail outright (see stack trace below), whereas testSingleExampleWeightUpdates and testMiniBatchWeightUpdates fail on expected vs. actual tests. testMLPTrivial passes. I think @agibsonccc said these tests were now passing?

Partial stack trace for TestMLP:
java.lang.IllegalStateException: Column of left array 3 != rows of right 10 or rows of left array 10 != columns of right 3
	at org.nd4j.linalg.util.LinAlgExceptions.assertMultiplies(LinAlgExceptions.java:75)
	at org.nd4j.linalg.api.ndarray.BaseNDArray.mmuli(BaseNDArray.java:2527)
	at org.nd4j.linalg.api.ndarray.BaseNDArray.mmul(BaseNDArray.java:2384)
	at org.deeplearning4j.nn.layers.BaseLayer.backwardGradient(BaseLayer.java:121)
	at org.deeplearning4j.nn.multilayer.MultiLayerNetwork.doBackWard(MultiLayerNetwork.java:1061)
	at org.deeplearning4j.nn.multilayer.MultiLayerNetwork.fit(MultiLayerNetwork.java:1215)
	at org.deeplearning4j.nn.multilayer.MultiLayerNetwork.fit(MultiLayerNetwork.java:1238)
	at org.deeplearning4j.nn.multilayer.BackPropMLPTest.testMLP(BackPropMLPTest.java:50)

AlexDBlack:
So I went through the old and your new/proposed version. I'm not convinced that either is correct at this point.

Anyway, have a look at the following. I'm happy to discuss this more.
https://www.dropbox.com/s/wcxqqgfxtsj3zbx/DL4J_BackpropDerivation.pdf?dl=0

nyghtowl:
Yeah to clarify the tests were not working when I pushed this yesterday which is why I have it down to not merge and WIP. Shared the current status (esp about the non-zeros in the output layer weight initialization for the tests you put together) because you've been great about digging into it lately. Will continue to update here as I'm working through it.

Thanks for sharing the link.

nyghtowl:
Key changes so far:

-  Added conditional for output change calculation based on the loss function. Alex is recommending the cross-entropy approach vs the squared error approach we originally used. Granted it appears that Bishop (1995) Neural Networks for Pattern Recognition is recommending that approach in general but need to dig in further
- Fixed feed-forward so it only track activation results and derivative results and ignores original input. Previously it was capturing the activation single value before it was applied to features as a item in the list which threw off the flow for backprop calculations
- Fixed actually capturing adjusted gradient results. Change consolidated AdjustedGradients, moved loop into MultiLayer and changed parameters that are passed. This required some changes in other files that referenced the class.
- Fixed actually applying the gradient updates to the NN parameters.  Moved loop into MultiLayer and changed parameters that are passed. This required some changes in other files that referenced the method.
- Made lots of name changes to clarify and to protect the innocent.

- BackPropMLPTest currently failing on testMPL and testMPL at the 3rd layer gradient application. Something in the shapes is throwing it off. The single and mini-batch weigh update tests are back to having actual and expected very close. Need to investigate further.

- There are several checks that need to be run to verify the algorithm approach and that it is working as expected. A couple outstanding points to work out such some authors recommend using current weights in the hidden layer delta calculations while other authors recommend using previous layer weights. We are currently going with the later. 

More to come

AlexDBlack:
Had a very quick scan over your recent changes. One thing that I noticed is biases are still divided by batch size twice: i.e., BaseLayer.update(Gradient) does a mean(0) on the biases (i.e., row sum + divide by batch size), but GradientAdjustment already does 'gradient.divi(batchSize);'. Thus, I think BaseLayer.update(Gradient) should do a sum(0) operation, not a mean(0) operation for biases.
I noted this in my pdf linked earlier, but it was in a footnote so it was easy to miss.

nyghtowl:
I did see your comment about the batch size division twice and talked with Adam briefly about it but we got side tracked on some of the other stuff. I'll look into that more. 

And no worries on the two approaches. Just digging through it today I realized most of your references for it explicitly talk about it being for the cross-entropy cost function. Although, Bishop may be recommending across the board.

Your links are great. How are you throwing together pdfs with such official-looking equations so quickly? They look like you are pulling pages from a book. I ask because if it's an easy technique to replicate I may want to use that at some point.

And good question on doBackWard and gradient initialization. I'll definitely look into that. Probably tomorrow because running low on steam for deep thoughts.

Thanks for helping to dig through this.

AlexDBlack:
Re: two approaches: yeah, I have been focusing on the classification/cross-entropy approach pretty much exclusively up to this point.

So the pdfs are just LaTeX. One of the many skills I've picked up along the way.
Not sure how familiar you are with it, but for example the latex source for the most recently linked PDF is here: https://gist.github.com/AlexDBlack/5a9df9a97b74b9eeca89
For compiling this I use TeXstudio, but there's a million different programs out there that essentially do the same thing. Lyx is another good one, a little more user friendly.

Also if you want to do LaTeX in github, you can use http://mathurl.com/. That's how I did the inline stuff (they are just images) here: https://github.com/deeplearning4j/deeplearning4j/issues/355

"Thanks for helping to dig through this."
Happy to help. The sooner it gets done, the sooner I can use it. :)

nyghtowl:
@AlexDBlack 
 - Ok great catch on the OutputLayer gradient method. Currently working through the existing approach to just get it to work. Will look at switching to that section at some point.
- I see where you are going with the mean and batchsize divide. Summing it isn't the same. I've tried a couple things there and currently one approach is just dividing weight gradients by batch size in the adjustment section. Not ideal and still doesn't give results like your test. 
- Thanks for the rec on LaTeX.
- You and me both on getting it working as expected sooner rather than later.

nyghtowl:
@agibsonccc & @AlexDBlack 

Two key points to note:

1) For some reason the different sigmoid equations you are both using produced slightly different results even though the are derived of each other as per (http://mathworld.wolfram.com/SigmoidFunction.html)
 
ND4J uses = pre-sig (1-pre-sig)
MLPtest used = exp(pre-sig) / (1+exp(pre-sig))^2

I changed MLP to follow ND4J for now

2) More importantly both equations should be calculating on post activation application as per the reference above. Meaning post sigmoid. I fixed MLPtest so it's using post sigmoid and currently failing.

ND4J's transformation is not applying the activation function prior to taking the derivative. See code [here](https://github.com/deeplearning4j/deeplearning4j/blob/mw_gradient_fix/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java#L572). I didn't see a quick fix on that when looking through it. Will post an issue in ND4J to track.

PR #385, opened 2015-06-30
Mw add score plotter

nyghtowl:
Hold off on merging

Fixes:
Added iterator to track and graph scores
Finally saves activation probabilities
Added test 
Made histograms dynamic in number of graphs to account for structures that do not have visual bias

Need to design all plotter iterators to store final graphs for each layer
Investigating adding training accuracy graph and ratio of weight updates


nyghtowl:
Found a couple key points around listeners:

- Only running on a single layer which limits application to model wide results. Looking into how to change especially to add graph like accuracy during training.
- Not running on OutputLayer
- On one test was not able to get the listener to run on multiple hidden layer model. Only ran on the first layer of the model - need to investigate further.
- Runs with feedforward but does not work with back prop. Investigate how to add.

nyghtowl:
Confirmed output layer is running and setup so that the different types of layers will save graphs to separate files. 
Need to ensure if more than one layer of the same type saves to a separate folder.

nyghtowl:
- RenderFilter left as is with commented out code. Expecting to link in with ui updates on separate branch
- MLP Backprop works
- Removing graph folder is manual. DeleteOnExit does not work unless the folder is empty. Need to explore functionality to make this easier to automate.

PR #394, opened 2015-07-03
Update WordVectorImp wordsNearest by vector

randombishop:
Added a version of wordsNearest that takes a vector as input.
Created a new private function to factorize the common part between the two versions of wordsNearest


agibsonccc:
https://github.com/deeplearning4j/deeplearning4j/commit/829a3b24c99fe8d33a286510cb10adb383af4d09

PR #424, opened 2015-07-11
Dl4jcaffe

jyt109:


dl4j-caffe/src/main/java/org/deeplearning4j/translate/CaffeModelToJavaClass.java
```diff
@@ -0,0 +1,23 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Created by jeffreytang on 7/9/15.
+ */
+public class CaffeModelToJavaClass {
+
+    public static NetParameter readCaffeModel(String caffeModelPath, int sizeLimitMb) throws IOException {
+        InputStream is = new FileInputStream(caffeModelPath);
```
EronWright: Ensure the stream is closed.

agibsonccc: @jyt109 

@EronWright Thanks

PR #430, opened 2015-07-12
Fix not found companion object

Lewuathe:
When we use NeuralNetwork* classes from spark ml (such as spark-shell), they cannot be found due to missing companion object. Companion object should retain constructor and training API at least.
```scala
scala> org.deeplearning4j.spark.ml.classification.NeuralNetworkClassification
<console>:20: error: object NeuralNetworkClassification is not a member of package org.deeplearning4j.spark.ml.classification
Note: class NeuralNetworkClassification exists, but it has no companion object.
              org.deeplearning4j.spark.ml.classification.NeuralNetworkClassification
                                                         ^
```


Lewuathe:
Yes, thank you for comment. I agree with your option about `train` and `learn` method. We cannot set hyper parameters. However instantiate with `object` class is a general pattern to create a object. So how about removing `train` and `learn` method? I think it is good to create `NeuralNetwork*` class with `apply` method.

Issue #487, opened 2015-07-20
BarnesHutTsne.plot's parameter nDims is not used; can't set BarnesHutTsne.numDimensions


SebNickel:
The method BarnesHutTsne.plot has a parameter nDims, which I assume is meant to correspond to the number of dimensions that the input data will be reduced to.

However, this parameter is not actually used in the method. The data is always reduced to the number of dimensions corresponding to the numDimensions property of the BarnesHutTsne object.

More specifically, the method plot calls the method fit(INDArray data), which calls the method fit() ... which calls the methods Tsne.calculate and Tsne.randn, and passes the value of the BarnesHutTsne object's numDimensions property to the respective nDims parameters of those methods.

Further, the numDimensions property is private and does not have a setter method. Nor does the nested Builder class provide a method for setting the number of dimensions.

So currently, fitting and plotting data to vectors of any dimension other than 2 requires more work than it should, and is quite constraining.

An easy way of fixing this, I think, would be add the following overloaded fit method, and have the plot method call this one instead of calling fit(INDArray data):

public void fit(INDArray data, int nDims) {
&nbsp;&nbsp;&nbsp;&nbsp; this.x  = data;
&nbsp;&nbsp;&nbsp;&nbsp; this.numDimensions = nDims;
&nbsp;&nbsp;&nbsp;&nbsp; fit();
}

PR #488, opened 2015-07-20
Fix for issue 487 with method BarnesHutTsne.plot

SebNickel:
See the issue I've opened [here](https://github.com/deeplearning4j/deeplearning4j/issues/487).

Added the overloaded method fit(INDArray data, int nDims) to BarnesHutTsne.
Changed the method BarnesHutTsne.plot so that it now calls this new fit method instead of fit(INDArray data). 
-> The parameter nDims of the method plot is no longer unused.
-> non-2D tSNE is now possible.

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/plot/BarnesHutTsne.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/plot/BarnesHutTsne.java
@@ -506,7 +506,7 @@ public class BarnesHutTsne extends Tsne implements Model {
      */
     public void plot(INDArray matrix,int nDims,List<String> labels,String path) throws IOException {
 
-        fit(matrix);
+        fit(matrix, nDims);
 
         BufferedWriter write = new BufferedWriter(new FileWriter(new File(path)));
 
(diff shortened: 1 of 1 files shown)


agibsonccc:
Could you also add in the old signature with a fallback value of 2? This preserves old behavior and avoid compilation errors while still allowing for the extended feature.

SebNickel:
@agibsonccc 
I haven't removed any signatures.

The fit(INDArray data) method is still there, unchanged. This uses the number of dimensions corresponding to the current value of BarnesHutTsne.numDimensions. Since I've added a new way of changing the value of numDimensions, perhaps that will cause trouble for you?

The signature of the plot method has not changed, either. It has always had an nDims parameter.

If you would like a plot method that always defaults to nDims = 2, we could *add* a new signature without an nDims parameter. But that would probably not preserve old behaviour.

I could also change either the fit(INDArray data, int nDims) method or the plot method (or both) to always set numDimensions back to 2 after it's done.
The documentation should then reflect this behaviour.

I suspect this last option would be the best way to fix the issues my changes may cause you.
Let me know.

agibsonccc:
Merged. I can deal with that. No need to make anything complicated.

PR #493, opened 2015-07-21
Mnist enhancement

EronWright:
@Lewuathe please have a look at this PR.  Am striving to align all datasources with what I perceive as the standard approach, and fix a couple of bugs.

- pruned columns
- hadoop paths (not local paths) as input
- camel case for relation params
- unit test

I removed numExamples because it seemed unnecessary to me.


deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/sql/sources/mnist/MnistRelation.scala
```diff
@@ -31,18 +30,33 @@ case class MnistRelation(imagesFile: String, labelsFile: String, numExamples: In
...
-    val rowBuilders = requiredColumns.map {
-      case "label" => (pt: (Int, Array[Int])) => Seq(pt._1.toDouble)
-      case "features" => (pt: (Int, Array[Int])) => Seq(Vectors.dense(pt._2.map(_.toDouble)))
-    }
+      val numExamples = manager.getLabels.getCount
 
-    sc.makeRDD(baseData.map(pt => {
-      Row.fromSeq(rowBuilders.map(_(pt)).reduceOption(_ ++ _).getOrElse(Seq.empty))
-    }))
+      sc.makeRDD((0 until numExamples).map(pt => {
```
Lewuathe: Why is `numExample` necessary here? In this code, `numExamples` always seems to represent all data. So I think that mapping simply sufficient in this use case.

EronWright: MnistManager doesn't provide a convenient iterator, that i know of.   Your numExamples technique seemed fine.

EronWright:
@Lewuathe thanks for the mnist datasource and for reviewing this PR.

PR #497, opened 2015-07-21
Ab layerbackprop

nyghtowl:
Needs ND4J fixes and resolved to dos below to merge

Main Changes
- Rework backprop (was backward) to take in epislon, gradient and previous layer instance and return  Pair <Gradient, epsilon> . Epsilon is the multiplication of the previous layer weight and delta which is used in the current layer delta calculation. Futher adjusts may be made with pulling out passing previous layer but will be investigated after this branch is collapsed into master
- Renamed score method that was actually just returning f1 to f1score to help differentiate from score
- Consolidated the processors to input preProcessor
- Changed processor backprop API to just pass epsilon
- Update just updates params and doesn't do any further calculations
- Removed deprecated content related to backward gradient calcs, filter and featureMap
- Cleared up OutputLayer calcuations esp loss function and backpropGradient
- Added backprop data structure transformations to processors
- Added processors to backprop method
- Added tests for processors
- Test updates (api related as well as backprop calc related)

CNN Specific Changes:
- Updated WeightInitUtil to provide multiple kernel weights base on number of outputs
- Updated ConvlutionLayer activate to use output num to represent num of kernels (filters) and utilmately num of feature maps that are expected.
- Still needs Nd4j fix to enable the convolution to apply the weights to the right revised shapes of the inputs. That dependes on kernel size, stride and padding.
- Put in minor fixes around names to get consistency (esp filter vs kernel)
- Fixed how test functions because nIn and nOut were not being set
- Added test of the resulting CNN structure dimensions
- Added padding parameter to configuration.This needs to be hooked into calculations if used

Outstanding to dos for this branch:
- Finish setting CNN backprop after ND4J fixes
- Verify CNN and Graves preProcessors work
- Test backprop for all layers
- CNN activate is not working yet


deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/recurrent/GravesLSTM.java
```diff
@@ -187,9 +188,12 @@ public Gradient backwardGradient(INDArray derivative, Layer nextLayer, Gradient
...
 			//Shape: [m,n^L]
 
-			INDArray prevLayerActivationSlice = activation;
+
+			// TODO this was added to master assuming activation was still passed in but change on this branch is epsilon - needs verification
+
+			INDArray prevLayerActivationSlice = epsilon;
 			if (!is2dInput) {
-				prevLayerActivationSlice = activation.slice(t, 2);
+				prevLayerActivationSlice = epsilon.slice(t, 2);
```
nyghtowl: @AlexDBlack I merged in master with the new Graves change but it uses activation which we are no longer pulling into this method as a parameter. Looked like you had switched it to epsilon before. Can you take a look at this change and see if it sounds correct. On master and this branch 2 Grave' tests fail regarding linear index on weights; thus, doesn't seem related.

PR #501, opened 2015-07-22
2d case handling for activation in GravesLSTM.

drlebedev:
Activation should produce same-shaped results in case input is 2-dimessional


agibsonccc:
CC @AlexDBlack

PR #519, opened 2015-07-27
Fix apply prevMemCellActivation derivative for memory cell error.

nyghtowl:
@AlexDBlack - Please take a look

- Found one gap in the calculations compared to notes on getting prevMemCellActivation derivative for the memory cell error
- Also seeking clarity on what prevLayerActivations should be and whether that should be weights for each of the different gates?
- I've fixed some tests to work with new api and added some additional componants to help test when passing in a layer since backprop does depend on a layer. 
- Tests still failing are biasGradient shape and epsilon shape. Probably due to new Output Layer passed in


AlexDBlack:
So this is a bit of a mess. A whole bunch of previous changes I had made went missing somewhere along the line. Will take a while to go through this and fix.
- Re: line 138 "//TODO is this still needed?" Yes. Very much so. This is for a necessary future feature that we need just can't do right now only because of it would require yet another API change (and it's not worth worrying about right now). But otherwise that commented-out code is good to go.
(More to come. I'll update this as I go)

PR #525, opened 2015-07-29
Fix dead link

Lewuathe:
Fix wrong link to LFW example.


nyghtowl:
Thanks for catching that

PR #549, opened 2015-08-08
remove experimental dsl elements

EronWright:
Spark 1.5/1.6 will introduce a standard dsl for pipeline components (SPARK-7434).   Until then I don't want to introduce any such syntax.


EronWright:
@Lewuathe are you OK with this?

Lewuathe:
@EronWright OK. Thank you for catching up!

PR #581, opened 2015-08-16
#421 - rework test-resources dependency

### LINKED ISSUE #421: "mvn test" is broken
Many of the unit tests rely on resources from the dl4j-test-resources module, and locate the resource with `ClassPathResource::getFile()`.   That approach works in the IntelliJ IDE but doesn't work on the command-line.  The reason is, in the latter case the test-resources dependency is provided as a JAR on the classpath (and `getFile` cannot formulate a path to an enclosed resource).

```
$ cd deeplearning4j-core
$ mvn -Dtest=CSVDataSetIteratorTest test

java.io.FileNotFoundException: class path resource [csv-example.csv] cannot be resolved to absolute file path because it does not reside in the file system: jar:file:/Users/eron/.m2/repository/org/deeplearning4j/dl4j-test-resources/0.0.3.3.5.alpha2-SNAPSHOT/dl4j-test-resources-0.0.3.3.5.alpha2-SNAPSHOT.jar!/csv-example.csv
	at org.springframework.util.ResourceUtils.getFile(ResourceUtils.java:205)
	at org.springframework.core.io.AbstractFileResolvingResource.getFile(AbstractFileResolvingResource.java:52)
	at org.deeplearning4j.datasets.iterator.CSVDataSetIteratorTest.testCSV(CSVDataSetIteratorTest.java:32)
```

This problem will preclude running the unit tests in travis-ci.

EronWright:
This changes the way that modules make use of test-resources.   The
goal was to facilitate both ‘mvn test’ and intellij test.   With this
patch, resources continue to be located via the classpath, as local
files.

Closes #421.


EronWright:
To test this change, I ran a few unit tests from each of the affected modules, in both intellij and with `mvn test`.

@agibsonccc you mentioned that you don't intend for dl4j-test-resources module to be published.  I think this patch gets us closer to that, because the dl4j-test-resources module need no longer produce a jar.   But I didn't make that particular change yet.

agibsonccc:
That's fine. This looks great.

PR #626, opened 2015-08-26
Add NetworkOutputProcessor interface, RnnOutputProcessor class

AlexDBlack:
Added a NetworkOutputProcessor
Basic need:
When using RNNs + OutputLayer, we previously needed to manually do two things:
(a) reshape labels 3d->2d (normally 3 dimensions: [miniBatchSize,outputSize,timeSeriesLength] to 2 dimensions)
(b) reshape outputs 2d->3d
Problem: we had no way of doing this automatically. We have preprocessors but these work before each layer, not after.
Now we can simply add a .outputProcessor(new RnnOutputProcessor()) to the configuration, and the reshaping of labels (3d->2d) and output (2d->3d) with OutputLayer is handled automatically.

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/MultiLayerConfiguration.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/MultiLayerConfiguration.java
@@ -21,11 +21,14 @@
 package org.deeplearning4j.nn.conf;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+
 import org.deeplearning4j.nn.conf.override.ConfOverride;
+import org.deeplearning4j.nn.conf.preprocessor.output.NetworkOutputProcessor;
 import org.nd4j.linalg.factory.Nd4j;
 
 import java.io.IOException;
@@ -49,6 +52,7 @@ public class MultiLayerConfiguration implements Serializable, Cloneable {
     protected double dampingFactor = 100;
     protected Map<Integer,InputPreProcessor> inputPreProcessors = new HashMap<>();
     protected boolean backprop = false;
+    protected NetworkOutputProcessor outputProcessor;
 
     /**
      *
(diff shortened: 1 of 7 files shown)


deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/preprocessor/output/NetworkOutputProcessor.java
```diff
@@ -0,0 +1,31 @@
...
+
+/** NetworkOutputProcessor processes both final network output
+ * (during forward pass) and the labels (during backprop).<br>
+ * Can be used to do things like reshape output and labels, for
+ * example.
+ * @author Alex Black
+ */
+@JsonTypeInfo(use= JsonTypeInfo.Id.NAME, include= JsonTypeInfo.As.WRAPPER_OBJECT)
+@JsonSubTypes(value={
+        @JsonSubTypes.Type(value = RnnOutputProcessor.class, name = "rnnOutputProcessor"),
```
EronWright: To be consistent with other aspects of the serialized output, please change  `name="rnnOutputProcessor"` to `name="rnn"`.  The term "outputProcessor" is redundant.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/preprocessor/output/RnnOutputProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+
+/** RnnOutputProcessor does two things:<br>
+ * (a) reshapes 2d output from OutputLayer to 3d (time series) output,
+ * in the same manner FeedForwardToRnnPreProcessor<br>
+ * (b) reshapes 3d (time series) labels for use with OutputLayer
+ * (similar to what FeedForwardToRnnPreprocessor.backprop() does)
+ * @author Alex Black
+ *
+ */
+public class RnnOutputProcessor implements NetworkOutputProcessor {
```
EronWright: You must implement `clone` here - see `conf.layers.Layer` for an example.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/MultiLayerConfiguration.java
```diff
@@ -49,6 +52,7 @@
     protected double dampingFactor = 100;
     protected Map<Integer,InputPreProcessor> inputPreProcessors = new HashMap<>();
     protected boolean backprop = false;
+    protected NetworkOutputProcessor outputProcessor;
```
EronWright: Be sure to adjust the `clone` method to clone the outputProcessor.

EronWright:
At first glance, creating a subclass of OutputLayer may be a more succinct way to achieve this.     It seems that the approach here is causing some undue conditional logic.

AlexDBlack:
Will be replaced by RnnOutputLayer instead

PR #649, opened 2015-08-29
#641 - rework MnistRelation for scalability

### LINKED ISSUE #641: Spark ML - MNIST data source suffers from memory exhaustion
The MNIST data source throws out-of-memory errors in trying to push the entire MNIST dataset from driver to executor.   

Spark assumes that the amount of data transferred from driver to executor is small.   The data source needn't load the dataset in the driver, by returning an RDD that loads it on-demand.

EronWright:
The MNIST data is now loaded by the executor, not the driver, and partitioned into reasonable batch sizes.

Closes #641.

### DIFF
--- a/deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/sql/sources/mnist/MnistRelation.scala
+++ b/deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/sql/sources/mnist/MnistRelation.scala
@@ -1,10 +1,11 @@
 package org.deeplearning4j.spark.sql.sources.mnist
 
 
+import java.net.URI
 import java.nio.file.Paths
 
 import org.apache.hadoop.fs.Path
-import org.apache.spark.Logging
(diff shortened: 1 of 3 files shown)


deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/sql/sources/mnist/MnistRelation.scala
```diff
@@ -30,36 +31,75 @@ case class MnistRelation(imagesPath: Path, labelsPath: Path)
...
+    private val labelsPath: URI,
+    private val requiredColumns: Array[String],
+    recordsPerPartition: Int = 1000
+  )
+  extends RDD[Row](sc, Nil) {
 
-      val numExamples = manager.getLabels.getCount
+  override val partitioner = None
 
-      sc.makeRDD((0 until numExamples).map(pt => {
```
EronWright: In future we must be wary of 'makeRDD', which pushes data from driver to executor.  Spark recommends that such a dataset be less than 100KB.

PR #683, opened 2015-09-06
Convolutionlayerautopilot

agibsonccc:
Convolution layer auto pilot. Allows for automatic setting of values wrt the values currently set in the network.

This minimizes the amount of effort required on the user of the conv nets when trying to setup the architecture.


EronWright:
Consider moving `ConvolutionLayerSetup` to `conf` package.

agibsonccc:
Will do.

PR #705, opened 2015-09-12
Additional layer configuration options

AlexDBlack:
- GravesLSTM forget gate bias option
- L1/L2/learning rate can be set on a per-layer basis


EronWright:
We recently moved a lot of properties from the NeuralNetConfiguration to the Layer.   The approach was to remove it from the conf, not to leave it in two places (like one overrides the other).   Now this PR is inconsistent with that.  Can we keep it consistent and remove l1/l2 from NeuralNetConfig?

agibsonccc:
Yeah go ahead.
ᐧ

PR #706, opened 2015-09-13
Setting up capability to configure global and layer

nyghtowl:
Returning functionality to either set a parameter once across all layers or modify to a specific layer.

- When creating a 20+ layer network that all use relu or set bias to zero, its easier to set it once 
- Some layers require specialized modifications on parameters like momentum and probably needed for any parameters related to updaters

Config listing duplication removed and tests updated


agibsonccc:
@EronWright

EronWright:
Wow this is a huge change to make at this time.  It is a big design change, treating layer values as overrides.    Reverses course on reducing NeuralNetConfig bloat.  Lacks the sophistication of per-layer-type defaults.   To reduce duplication (the impetus for this PR), it might be wiser to design a dsl for cloning the previous layer, or a templating scheme.    My 2 cents.

Specifics: 
- default values appear to be regressing in a few cases
- deep cloning must be updated (e.g. dist)
- JSON/YAML serialization of NaN should be validated

nyghtowl:
Its a cleaner user experience to have the option to set it globally and/or locally. In a 20 layer structure, having to specificy in every layer 4 parameters that are the same and then 2 or 3 that are different makes it difficult to build the structure and work with when tuning. Better to only set in the layer the parameters that are explicitly different for that layer.

Agree the json format that is built out with duplicate values is clunky and bloated as noted above. Is there a way to allow us to set globally and locally while keeping the configuration json that's built streamlined?

agibsonccc:
I think there's a good middle ground here. We need to keep specific
parameters to layers but general to layers/global. We need better unit
tests around this though to make sure local over rides global.

EronWright:
How about:  a) the general properties are defined on the nnconf.builder (not on nnconf), and b) the layer builder's constructor is given a reference to the nnconf builder and sets the layer's initial values based on that.   This would retain the existing JSON format.   The dsl would resemble:

```
        MultiLayerConfiguration.Builder builder = new NeuralNetConfiguration.Builder()
                .seed(seed)
                .batchSize(batchSize)
                .iterations(iterations)
                .constrainGradientToUnitNorm(true)
                .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
                .l1(.9)
                .list(3)
                .layer(0, builder.convolution()
                        .l1(.8)
                        .kernelSize(10, 10)
                        .nIn(nChannels)
                        .nOut(6)
                        .weightInit(WeightInit.XAVIER)
                        .activation("relu")
                        .build())
```

agibsonccc:
I like this idea.

AlexDBlack:
Sounds good to me. FYI a basic unit test recently added https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/conf/layers/LayerConfigTest.java

EronWright:
@nyghtowl please revert the merge, the PR is not ready.   Layers mustn't have NaN values, that was the main thrust of my complaint.  Please give me a chance to fully review this.

PR #721, opened 2015-09-16
make dropout consistent

agibsonccc:
Dropout was inconsistent for some activations (namely train/test)

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
@@ -175,6 +175,20 @@ public abstract class BaseLayer<LayerConfT extends org.deeplearning4j.nn.conf.la
     }
 
 
+    @Override
+    public INDArray preOutput(INDArray x, TrainingMode training) {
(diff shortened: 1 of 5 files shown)


deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
```diff
@@ -300,7 +314,7 @@ public INDArray preOutput(INDArray x, boolean training) {
         INDArray b = getParam(DefaultParamInitializer.BIAS_KEY);
         INDArray W = getParam(DefaultParamInitializer.WEIGHT_KEY);
         if(conf.isUseDropConnect() && training) {
-            if (conf.getLayer().getDropOut() > 0) {
+            if (conf.getLayer().getDropOut() > 0 && training) {
```
justhalf: `training` is already checked in line 302, I think it's unnecessary to put it here also?

agibsonccc: https://github.com/deeplearning4j/deeplearning4j/commit/a395a1ce11d58570b2897acf5971b105cc4edb2a

PR #736, opened 2015-09-23
x86 must be a test dependency

EronWright:
Without the test scope, the netlib dependency becomes harder to exclude.   And I don't think we intended to remove the test scope, was just an accidental commit.


EronWright:
@agibsonccc We may need this fix to make the spark package work.

PR #754, opened 2015-09-28
Adding back config changes 

nyghtowl:
These changes were lost in a revert 2 weeks ago. Goal is to make configuration streamlined when there is several layers and configuration replication (esp. when making updates during training)

- Changes are primarily in NeuralNetConfig and Layer to enable parameter configuration globally for all layers or for specific layers
- Added test
- Updated api calls across other files

This configruation is leveraging the existing application already on master of Double.NaN to initialize attributes on Layer.


agibsonccc:
Again let me review this first before merging.  Thanks!

PR #812, opened 2015-10-26
Compilation problems

Shmuma:
CSVSequenceRecordReader and NumberedFileInputSplit are not present in trunk canova api. Remove the test to make this compilable again.


nyghtowl:
@Shmuma Thanks for the submission. These files are in the current Canova repo. Pull from that repo, do a mvn clean install in Canova then pull from this repo and compile again. Closing this issue because the current dl4j does compile with the latest Canova.

Shmuma:
Right. Sorry for bothering, I've totally overlooked import of those files in canova master.

nyghtowl:
Not a bother at all. We are definitely happy to have the help.

PR #817, opened 2015-10-27
new branch for bleeding spark revised

tripiana:
As requested by @agibsonccc bleeding spark version.

I recommend checking the diff between dl4j-spark/dl4j-spark-bleeding and the diff between dl4j-spark-ml/dl4j-spark-ml-bleeding before accepting the commit. Just to see some code modifications I've done so as to be able to compile.

Compiled with java 1.8, Spark 1.5.1, Hadoop 2.7.1.


tripiana:
Checks have failed because it cannot download bleeding versions of nd4j and Canova.

@agibsonccc, as you requested, Spark cluster works (can be set up) with this version. If anyone has any simple example to make a small test it will be superb.

agibsonccc:
So I'm wondering if there's anything we could do here with generating multiple jars for different versions of spark?

tripiana:
Well if I have time I want to test the cluster with the classification scene tomorrow (un a few hours in Spain) but I do not know what to do to make bleeding and standard branches sync as the people improves standard version. That's just a nightmare.

agibsonccc:
Welcome to my world :).  Let me think on this one a bit. I think there can be 2 versions. It might be something like maven profiles. Not sure yet...I think I understand why people compile spark now.

EronWright:
I'd vote for dl4j-spark-ml to track Spark 1.5. The two 'spark' modules should track 1.3 and 1.5 respectively.   If we assume binary compatibility of Spark 1.3->1.5, then the dl4j-spark-ml module could depend on the dl4j-spark module. That said, we had that arrangement earlier and there was a problem (I don't remember what).  Don't worry about the module names; just place code in one or the other based on API compatibility. 

Maybe I don't understand the goal, but at first glance, the option of duplicating all the code should be a last resort.

agibsonccc:
I agree completely. Thanks for your input here. I was hoping you'd chime in here.Spark.ml should definitely track bleeding edge. We can have the rdd version support distros.

I actually like the idea of the clear segmentation there. Let me think on this a bit more.

tripiana:
Spark rdd version compiles against 1.5 without any problems. So adding a
profile is just the right direction.

I had to modify three files at the ML version to be able to compile with
1.5 so please review my changes because I'm not familiar with the code.

Thanks!

tripiana:
Any update on this? @agibsonccc, I need the s3 bucket to execute the test but I cannot donwload it. I've filed and issue.

agibsonccc:
Limited bandwidth on my end atm. Ping me on gitter about the s3 bucket.

PR #834, opened 2015-11-02
MultiLayerNetwork.activateSelectedLayers method

raver119:
Simple routine for activation few layers in cycle. 
That could be useful in tasks like AutoEncoder, where you dont need full calculation cycle like feedForward for encoding. Or you have only encoded values, and need to decode them.


agibsonccc:
Clean it up a bit and squash the commits and I think we'll be good. Look at some of my inline comments as well.

agibsonccc:
Closed in favor of #835

PR #835, opened 2015-11-02
activateSelectedLayers

raver119:
Calculate activation for few layers at once. Suitable for autoencoder partial activation.

As for comments:

Q. What if input is already set?
A. As i can see, activationFromPrevLayer(int curr, INDArray input,boolean training) method, used in activateSelectedLayers fills input, regardless of its previous state, and feedForward(INDArray input) works in the same way. So i dont see problems here - similar functions are declared and work in the same way.

Q. Use slf4j for output.
A. Fixed.

Q. Squash the commit.
A. Done, sorry, i'm still new to git mechanics :)


agibsonccc:
Done.

PR #919, opened 2015-12-01
Doc2Vec rework

raver119:
This is pre-final pr on w2v/d2v fixes.

ParagraphVectors now have two training modes: with and without parallel word vectors training. 

Few sentence iterators were added, etc.

There's still some work should be done, especially with iterators, ParagraphVectors last training mode, subsampling etc, but to do it i need to have current version of InMemoryLookupTable, with all latest fixes included.


raver119:
As soon as i'll finish with it, i'll post detailed list of changes, sure.

PR #956, opened 2015-12-14
AbstractVectors

raver119:
- AbstractVectors implementation - algorithm and all dependant classes were transformed to become generic and defined as <T extends SequeneElement>.

- ParagraphVectors and Word2Vec was migrated for AbstractVectors mechanic, and now became wrappers over AbstractVectors

- Proper tests for ParagraphVectors were added

- New implementation of VocabCache was introduced, AbstractCache<T extends SequenceElement>. Current implementation, InMemoryLookupCache was left intact mostly, as static implementation of VocabCache<VocabWord>

- Abstract VocabConstructor was introduced, entity for vocabulary mastering from multiple sources with different settings being applied.

- Bunch of minor fixes over existing codebase: tokenizers, iterators, etc

- deeplearning4j-nlp and dl4j-spark-nlp tests are passing


raver119:
Sorry, i can't squash all these commits into 1 again. 

But there's no conflicts as i see.

Issue #958, opened 2015-12-15
 NeuralNetPlotterIterationListener does not work with MultiLayerNetwork


Ingwar:
Hello, 

I am trying to use `NeuralnetPlotIterationListener` in a following way:
```scala
model.setListeners(new ScoreIterationListener(), new NeuralNetPlotterIterationListener(10, true))
```

I got this error:
```
16:43:13.334 [main] INFO  o.d.plot.NeuralNetPlotter - Rendering histogram graphs for data analysis...
16:43:37.856 [main] INFO  o.d.plot.NeuralNetPlotter - Std out [Loading matrix 0_W, , Loading matrix 0_b, , Loading matrix 1_W, , Loading matrix 1_b, , Loading matrix 3_W, , Loading matrix 3_b, , Loading matrix 4_W, , Loading matrix 4_b, , Loading matrix 6_W, , Loading matrix 6_b, , Loading matrix 7_W, , Loading matrix 7_b, , Loading matrix 0_W-gradient, , Loading matrix 0_b-gradient, , Loading matrix 1_W-gradient, , Loading matrix 1_b-gradient, , Loading matrix 3_W-gradient, , Loading matrix 3_b-gradient, , Loading matrix 4_W-gradient, , Loading matrix 4_b-gradient, , Loading matrix 6_W-gradient, , Loading matrix 6_b-gradient, , Loading matrix 7_W-gradient, , Loading matrix 7_b-gradient, ]
16:43:37.856 [main] ERROR o.d.plot.NeuralNetPlotter - Std error [/usr/lib/python2.7/dist-packages/matplotlib/__init__.py:874: UserWarning: axes.color_cycle is deprecated and replaced with axes.prop_cycle; please use the latter.,   warnings.warn(self.msg_depr % (key, alt_key))]
Exception in thread "main" java.lang.UnsupportedOperationException
	at org.deeplearning4j.nn.multilayer.MultiLayerNetwork.activationMean(MultiLayerNetwork.java:410)
	at org.deeplearning4j.plot.NeuralNetPlotter.plotActivations(NeuralNetPlotter.java:275)
	at org.deeplearning4j.plot.NeuralNetPlotter.plotNetworkGradient(NeuralNetPlotter.java:326)
	at org.deeplearning4j.plot.iterationlistener.NeuralNetPlotterIterationListener.iterationDone(NeuralNetPlotterIterationListener.java:111)
[...]

PR #979, opened 2015-12-19
Deprecate `GradientPlotterIterationListener` and NeuralNetPlotterIterationListener

Ingwar:
According to the #958, they are not supported anymore.

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/plot/NeuralNetPlotter.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/plot/NeuralNetPlotter.java
@@ -45,7 +45,10 @@ import org.springframework.core.io.ClassPathResource;
  * for visualizations
  * @author Adam Gibson
  *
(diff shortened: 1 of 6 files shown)


Ingwar:
Actually, are others iteration listeners in the `org.deeplearning4j.plot.iterationlistener` still supported or they should be deprecated too?

agibsonccc:
Anything to do with the neuralnetplotter more or less. There's some other
ones in there.

Thanks!

Ingwar:
OK, I'll take a closer looks at them and add suspicious ones to this PR.

Ingwar:
Also, I am wondering, why Travis build have failed? I haven't touched the code of the listeners, so the build shouldn't be affected.

AlexDBlack:
@Ingwar I wouldn't worry about Travis right at the moment. It used to work, but there's some issues with it after CodeHaus shut down that we haven't got around to fixing yet.

Ingwar:
I've marked all other literation listeners that use `NeuralNetPlotter` as well as `NeuralNetPlotter` itself as deprecated.

PR #1025, opened 2016-01-07
Bidirectional LSTM

bennythedataguy:
I'm a man of few words.

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/layers/GravesBidirectionalLSTM.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/layers/GravesBidirectionalLSTM.java
@@ -1,0 +1,62 @@
+/*
+ *
+ *  * Copyright 2015 Skymind,Inc.
+ *  *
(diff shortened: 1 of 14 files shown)


agibsonccc:
Could you add a gradient check test? https://github.com/deeplearning4j/deeplearning4j/blob/18fd44b9e5b4ee6817b45b260abec9898ccbb1e1/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNNGradientCheckTest.java#L55

bennythedataguy:
Yes, I meant to but forgot.

AlexDBlack:
Final comments and changes:
- rnnActivateUsingStaredState: we'll make this an unsupported op (used for truncated BPTT). This means we can remove the state keys too
- Placeholder masking array comments in LSTMHelpers: can be removed
- Javadoc for LSTMHelpers: briefly add details of why functionality was split out into that class (1-2 sentences) + add author tags
- forwardEpsilon.add(backwardsEpsilon) in GravesBidirectionalLSTM -> addi

As noted: we'll need to handle variable length time series stuff at some point, but we'll open a separate issue for that.

PR #1157, opened 2016-02-18
Issue - 1153 MultiLayerNetwork score computation - output layer forward pass done twice

konradkalita:
I hope it fixes issue 1153, but it is my first PR to deeplearning4j so I might have misunderstood something.


AlexDBlack:
Thanks for this.
It's pretty much exactly what I would have done, though I did make one minor change for correctness:  https://github.com/deeplearning4j/deeplearning4j/pull/1158
For future reference, running the gradient checks is a good idea when making changes to scoring/gradients/backprop.
(We will have Travis set up again in the future, which will automatically run those tests on all PRs)

Issue #1253, opened 2016-03-10
Epoch listener for SequenceVectors


LukasMasuch:
Feature request: Could you consider adding a simple epoch listener to sequence vectors. That would be really helpful for evaluating the accuracy of the vector space after every epoch. I'm dealing with 3GB of textual data and it would take too long to wait for all epochs to see if the configuration made a positive impact on the accuracy. I have currently implemented this in my dl4j version in this way:
```Java
private EpochListener epochListener;

public void setEpochListener(EpochListener epochListener){
    this.epochListener = epochListener;
}

public interface EpochListener {
    void epochCompleted(int epoch,WordVectors wordVectors);
}
```
An I call it on the completion of every epoch:
```Java
if (epochListener != null){
      epochListener.epochCompleted(currentEpoch,this);
 }
```

raver119:
Code will be refactored and included in the master branch.

raver119:
After refactoring it's implemented in the following way:

There's 2 possible reasons to activate listener: end of epoch, or sequence passed. Like every 2nd epoch, or every 100000th sequence passed.

As example i've provided SerializingListener, that can serialize model after X epochs or lines.

LukasMasuch:
Thanks @raver119! The refactored listener is very helpful.

PR #1255, opened 2016-03-10
Added epoch listener to SequenceVectors

LukasMasuch:
Feature further described here: #1253


raver119:
Let's do that in better way. Let's define that interface externally, and make it callable not only at different epochs, but via different triggers.

Most of w2v tasks do not require multiple epochs, and that makes this iterator useless. But if we'll provide hooks for execution, i.e. on every 1m sequences processed - that'll be better.

However, for GloVe tuning epoch listener will definitely help alot.

raver119:
So, i'll merge this pr now, but will refactor it.

PR #1257, opened 2016-03-11
word2vec fixes

darrenfoong:
Fix bugs in WordVectors/Word2Vec.


raver119:
Sorry, that protected method isn't a mistake. It's protected for a reason: it's a not finished implementation. Don't make it public please.

raver119:
The same for UNK word. It's not guaranteed to be existent in the vocab, and that's yet to be reflected.

raver119:
@darrenfoong could you please come to gitter and poke me there, so we could discuss these fixes?

PR #1264, opened 2016-03-11
Replace UNK with DEFAULT_UNK

darrenfoong:
Missed these ones in previous PR.


raver119:
Thats wrong. Because at runtime there can be another UNK, due to your setUnk() call.

raver119:
I've done that using VectorsConfigurations, as we've discussed earlier, so i'll close this PR. 

Thanks for your help there.

PR #1338, opened 2016-04-01
new leakyreluAlpha can be set by user in NeuralNetConfiguration 

eraly:


eraly:
Generalized structure

PR #1422, opened 2016-04-21
Add ArrayWritable to optimize loading of images

saudet:
Please merge if this looks alright...

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/canova/RecordReaderDataSetIterator.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/canova/RecordReaderDataSetIterator.java
@@ -25,6 +25,7 @@ import org.canova.api.io.converters.WritableConverterException;
 import org.canova.api.records.reader.RecordReader;
 import org.canova.api.records.reader.SequenceRecordReader;
 import org.canova.api.writable.Writable;
+import org.canova.common.data.NDArrayWritable;
 import org.deeplearning4j.datasets.iterator.DataSetIterator;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.dataset.DataSet;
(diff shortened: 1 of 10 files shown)


nyghtowl:
Put questions in the code. So the general idea is to have an INDArray Writable for any input object in regards to what you are building? And is it expected to replace all RecordReader objects?

saudet:
Thanks for the review! Yes, the general idea is to replace the feature portion of `Collection<Writable>` with an array, to avoid creating a lot of wrapper `Writable` objects. I don't think we need to do much for existing `RecordReader` classes, but it does affect what they return, and I've updated `DataSetIterator` classes for that. It seems to be working fine. Hopefully, it won't break anything :)

agibsonccc:
Updates on this?

saudet:
Apparently it breaks the labels for some loaders, but I'm not able to reproduce the issue here. Waiting for @nyghtowl on this...

nyghtowl:
Regarding examples to run: Cifar, LFW, ImageNet - reality ongoing changes since end of Feb have made CNNs a moving target every day on whether its working. One current issue affecting master is: https://github.com/deeplearning4j/nd4j/issues/858

Label issue: (sharing what I sent Sat) the code from [here](https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/canova/RecordReaderDataSetIterator.java#L223) to [here] (https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/canova/RecordReaderDataSetIterator.java#L243) is defining the label to include in the DataSet. It's expecting a labelIndex on where in the list of values that are passed in did we append | store the numerical value of the label. A change like [this](https://github.com/deeplearning4j/deeplearning4j/pull/1422/files#diff-9d6a99bc16fffd413a2de45d49325b24L39) assumes that the labelIndex where the label value was appended is in position 1 of that array vs at the end of the array which would be equivalent to (height \* width \* channels).

treo:
@nyghtowl the issue affecting master has been fixed.

nyghtowl:
@treo that was just an example - main point is there are ongoing issues that have been getting in the way of CNN dev in regards to Sam's request for a working example.

nyghtowl:
@saudet verified code is working as expected - good to go for merge

saudet:
Great! Thanks for the review

PR #1530, opened 2016-05-12
Add identity weight init

atollFP:
To enable fake identity layer.


treo:
Would be great if you could add a test to it. Just to make sure that future changes will not change the behavior of this.

atollFP:
Will do when I have more time. I feel like it's too simple to break though.

PR #1586, opened 2016-05-23
Fix CifarDataSetIterator and renable tests using it and CifarLoader

saudet:
For https://github.com/deeplearning4j/Canova/issues/156

Also, CifarLoader doesn't support image scaling, so test using the data's native size (32x32x3).


agibsonccc:
Could we enable that with the image loader maybe?

saudet:
I'm thinking we'll eventually be able to just have it go through whatever I'll do for image transformation...

saudet:
But it looks easy enough to switch BaseImageLoader to NativeImageLoader for now so let me do that

PR #1600, opened 2016-05-26
Run specified number of iterations in Spark impl

crockpotveggies:
Both Spark multi-layer and computation graph do not run the specified number of iterations when `AVERAGE_EACH_ITERATION` is set to false. Fix this.


crockpotveggies:
@AlexDBlack @agibsonccc can you have a quick look to ensure I've understood the logic properly?

crockpotveggies:
I've made a serious error without realizing that iterations is in the conf itself. Closing.

PR #1611, opened 2016-05-29
Refactor mnist data fetcher to avoid putScalar (better for cuda)

AlexDBlack:


saudet:
So we're removing putScalar() calls and replacing that with temporary buffers? Although it doesn't matter for this small dataset, I'm not sure that in general this a step in the right direction...

AlexDBlack:
Well, if you have a better idea - I'm open to suggestions. We have to do the conversion from byte[] to float or double at some point.

PR #1612, opened 2016-05-29
Update image-data-pipeline.md

plankes-projects:
Added link for a simplified custom image Eclipse/Maven project.

### DIFF
--- a/image-data-pipeline.md
+++ b/image-data-pipeline.md
@@ -5,6 +5,8 @@ layout: default
 
 # Customized Data Pipelines for Images etc.
 
+Have a look at following Eclipse/Maven project for a simple quick start of how to use DL4J with custom images: https://github.com/plankes-projects/DL4JCustomImage
+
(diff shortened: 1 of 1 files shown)


agibsonccc:
Before I merge this, would you mind maybe adding what we discussed on gitter last night? eg: merging datasets and the like? Keep in mind that it'd be better if we could use the image record reader here. If there's a problem with that record reader, let's get that sorted out here before we point to this.

plankes-projects:
The merge is already there, but I'd like to not include the iterator.
Without it, it is more transparent and easier to start with.

agibsonccc:
You're trying to circumvent using a whole library. The whole point of the canova project is to be the defacto way any ETL is done. The problem with ETL in python and really any other machine learning framework out there is the lack of an actual data pipeline abstraction that also works at scale. Regardless of whether you think it's confusing or not, as the maintainer of the project. I'm going to get 100x the requests (eg: after you're gone) being confused about the right way to do things with the library. If it's not consistent with how the maintainers are going to be doing it, I'm closing this.

PR #1640, opened 2016-06-04
canDoBackward() in BaseGrahVertex: check for epsilons

StefanoWoerner:
The default canDoBackward() implementation did not check for the errors to be present. This pull request changes it, as both the inputs and the epsilons are required for the backward pass.


agibsonccc:
Thanks!

Issue #1614, opened 2016-05-30
KernelValidationUtil Error Checking


UnyieldingOrca:
I'm trying to use the Convolution Layer on time series data(i.e. an image with height 1) and I had it working for 0.4-rc3.8. When i upgraded to 0.4-rc3.9 i got this exception.

```
Exception in thread "main" org.deeplearning4j.nn.conf.inputs.InvalidInputTypeException: Invalid input: activations into layer are h=1 but kernel size is 1 with padding 0
	at org.deeplearning4j.nn.layers.convolution.KernelValidationUtil.validateShapes(KernelValidationUtil.java:16)
	at org.deeplearning4j.nn.conf.layers.setup.ConvolutionLayerSetup.getConvolutionOutputSize(ConvolutionLayerSetup.java:402)
	at org.deeplearning4j.nn.conf.layers.setup.ConvolutionLayerSetup.updateLayerInputs(ConvolutionLayerSetup.java:97)
	at org.deeplearning4j.nn.conf.layers.setup.ConvolutionLayerSetup.<init>(ConvolutionLayerSetup.java:74)
	at com.dap.dl4j.DLMomentumNet.<init>(DLMomentumNet.java:69)
	at com.dap.strategy.dl_momentum_en.DLMomentumEn.TrainNet(DLMomentumEn.java:203)
	at com.dap.strategy.dl_momentum_en.DLMomentumEn.main(DLMomentumEn.java:520)

```

My network setup code:

```
MultiLayerConfiguration.Builder builder = new NeuralNetConfiguration.Builder()
				.seed(seed)
				.iterations(1)
				.regularization(true).l2(0.0005)
				.learningRate(0.01)
				.weightInit(WeightInit.XAVIER)
				.optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
				.updater(Updater.NESTEROVS).momentum(0.9)
				.list()
				.layer(0, new ConvolutionLayer.Builder(1, 5)
						.nIn(1)
						.padding(0,0)
						.stride(1, 5)
						.nOut(20)
						.activation("identity")
						.build())
				.layer(1, new SubsamplingLayer.Builder(SubsamplingLayer.PoolingType.MAX)
						.kernelSize(1,2)
						.stride(1,2)
						.build())
				.layer(2, new DenseLayer.Builder().activation("relu")
						.nOut(512).build())
[...]

PR #1672, opened 2016-06-12
Update KernalValidationUtil

UnyieldingOrca:
I believe the corner case this file was checking was unnecessary. I submitted an issue about this a couple weeks ago (#1614) and this is my suggested fix.


agibsonccc:
LGTM

PR #1681, opened 2016-06-14
Slightly improve readability of new Customized Data Pipelines tutorial

treo:
Replace quoting blocks with language annotated source code for slightly better readability.


eraly:
Thanks treo!

PR #1698, opened 2016-06-15
Will jenkins trigger?

SusanBot:
please o please


eraly:
Will this trigger another build? Please don't. It's the same sha.

eraly:
And it didn't.

PR #1700, opened 2016-06-16
Introduce support for cuDNN starting with ConvolutionLayer and SubsamplingLayer

saudet:
This is far from complete, but I'd like some early feedback on if this is going the right way, or if we should do things differently. If it looks alright though, it can be merged.

To use, we only need to add the following dependency in the `pom.xml` file:
```xml
        <dependency>
            <groupId>org.deeplearning4j</groupId>
            <artifactId>deeplearning4j-cuda-7.5</artifactId>
            <version>${dl4j.version}</version>
        </dependency>
```
I've tested with LenetMnistExample only for now, but training time is cut by more than 50% on that sample, so it looks promising.

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionHelper.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionHelper.java
@@ -1,0 +1,34 @@
+/*
+ *
+ *  * Copyright 2016 Skymind,Inc.
+ *  *
+ *  *    Licensed under the Apache License, Version 2.0 (the "License");
+ *  *    you may not use this file except in compliance with the License.
+ *  *    You may obtain a copy of the License at
+ *  *
+ *  *        http://www.apache.org/licenses/LICENSE-2.0
+ *  *
+ *  *    Unless required by applicable law or agreed to in writing, software
+ *  *    distributed under the License is distributed on an "AS IS" BASIS,
+ *  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  *    See the License for the specific language governing permissions and
+ *  *    limitations under the License.
+ *
+ */
+package org.deeplearning4j.nn.layers.convolution;
+
+import org.deeplearning4j.berkeley.Pair;
+import org.deeplearning4j.nn.gradient.Gradient;
+import org.nd4j.linalg.api.ndarray.INDArray;
+
+/**
+ * Helper for the convolution layer.
+ *
+ * @author saudet
+ */
+public interface ConvolutionHelper {
+    Pair<Gradient, INDArray> backpropGradient(INDArray input, INDArray weights, INDArray delta,
+            int[] kernel, int[] strides, int[] pad, INDArray biasGradView, INDArray weightGradView, String afn);
+    INDArray preOutput(INDArray input, INDArray weights, INDArray bias, int[] kernel, int[] strides, int[] pad);
+    INDArray activate(INDArray z, String afn);
+}
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionLayer.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionLayer.java
@@ -35,6 +35,8 @@ import org.nd4j.linalg.convolution.Convolution;
 import org.nd4j.linalg.factory.Nd4j;
 import org.nd4j.linalg.indexing.NDArrayIndex;
 import org.nd4j.linalg.ops.transforms.Transforms;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Arrays;
 
@@ -45,12 +47,29 @@ import java.util.Arrays;
  * @author Adam Gibson (original impl), Alex Black (current version)
  */
 public class ConvolutionLayer extends BaseLayer<org.deeplearning4j.nn.conf.layers.ConvolutionLayer> {
+    protected static final Logger log = LoggerFactory.getLogger(ConvolutionLayer.class);
+
+    ConvolutionHelper helper = null;
+
     public ConvolutionLayer(NeuralNetConfiguration conf) {
         super(conf);
+        initializeHelper();
     }
 
     public ConvolutionLayer(NeuralNetConfiguration conf, INDArray input) {
         super(conf, input);
+        initializeHelper();
+    }
+
+    void initializeHelper() {
+        try {
+            helper = Class.forName("org.deeplearning4j.nn.layers.convolution.CudnnConvolutionHelper")
+                    .asSubclass(ConvolutionHelper.class).newInstance();
+        } catch (Throwable t) {
+            if (!(t instanceof ClassNotFoundException)) {
+                log.warn("Could not load CudnnConvolutionHelper", t);
+            }
+        }
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


saudet:
BTW, here are what the hot spots of LenetMnistExample.java look like on my laptop:
dl4j1700profile
It takes about 8:20 minutes with my GPU (GeForce GT 640M), but the equivalent examples/mnist/train_lenet.sh from Caffe takes only 2:57 minutes with cuDUNN (and 9:23 without cuDNN but with CUDA), so there's still work to do here. Getting close to respectable numbers though :)

AlexDBlack:
@saudet btw, ping me in dev channel if you ask for feedback and don't hear from me more promptly :)

anyway, how difficult would it be to have just one conv layer type, rather than "ConvolutionLayer" and "CudnnConvolutionLayer"? Is it even possible?
Ideally I'd want them to be interchangeable (from a configuration point of view) - just say you train on cudnn but then want to deploy on CPU... can't do that with 2 different implementations.

saudet:
Why not? CudnnConvolutionLayer is just a subclass of ConvolutionLayer so they already are interchangeable, or am I missing something? To deploy without cuDNN, the way I did it, we just need to not include the JAR in their class path, and that's it. If we were to make it part of deeplearning4j-core, all users would have to build against CUDA and runtime would be problematic, not impossible, but I'm not sure I understand the advantages of doing that.

In any case, if we want to be able to use the CPU backend simultaneously we'll also need to create subclasses for NativeOps as well. Right now it binds either to the CUDA binaries or the CPU binaries. The same JNI methods can't link to both at the same time.

AlexDBlack:
ok, right. looking at the code again: the switch statements in the ConvolutionLayerSetup had (has) me confused.
"switch (inputLayer.getClass().getSimpleName()){"
inputLayer is a org.deeplearning4j.nn.conf.layers.Layer (i.e., configuration), and there is no org.deeplearning4j.nn.conf.layers.CudnnConvolutionLayer, so you'll never hit the case "CudnnConvolutionLayer":, right?

I had assumed you were defining a separate configuration class for cudnn as a result.

saudet:
That's correct, there's no new "conf.layers". Sorry for the confusion :(

The thing is, ideally cuDNN should do everything and stand on its own, but it doesn't. It doesn't do everything. For things that are missing, we either have to copy/paste the missing features from the existing classes (if you really want to keep the exact same class name), or we subclass. One "drawback" of subclassing is with ConvolutionLayerSetup. It checks for the name of the class, instead of using the "instanceof" operator. But is this the only benefit of having two different classes, with the same exact name? Or is there something else that would outweigh the cost of maintaining copy/pasted code?

saudet:
Or we could move everything out of the layer classes into helper classes like it's done for RNNs?

saudet:
It might also be possible to have a cuDNN helper class, and when it loads successfully, use it inside the existing layers. If that sounds like what we want, I'll give it a try.

AlexDBlack:
I'm happy with subclassing. My main concern initially was the configuration side, not the implementation side. Now that I properly understand what you've done with configuration, it all looks OK to me.

And yeah, we can do helper classes and the like, but it's probably not necessary (not sure we gain much by doing so).

saudet:
Ok, thanks. There's also RNNs to consider. For CNNs it's general enough to be useful as a real backend, but for RNNs, from what I understand right now, cuDNN offers only a very high-level specialized implementation, so we might have to come up with a new layer type entirely...

AlexDBlack:
CuDNN has LSTMs
- However: They are using a variant that's slightly different to ours (ours has peephole connections, theirs doesn't; not much practical/performance difference between the two) https://devblogs.nvidia.com/parallelforall/optimizing-recurrent-neural-networks-cudnn-5/
    - I'm not opposed to adding this variant (it can share a lot of code with the current one). It might have to be a separate layer type, however.
- It looks like parallelizing across multiple stacked layers helps a lot (see previous link). However, at first glance this doesn't really match our current design (forward/backward pass done separately for each layer). Few different ways to handle that. Plus we can add it further down the line, too.
- Some stuff I'd have to look into::
    - How (or, if) CuDNN handles bidirectional RNNs
	- How we'd support 'one step at a time' forward pass (need access to internal state, etc) using CuDNN

They also have GRU and (vanilla) RNN implementations. Latter is easy to add a CPU version for (but few people use them); former I tried to implement at one point but didn't get gradient checks passing (and haven't looked at it since).

AlexDBlack:
Oh, and they have batch norm too, apparently: https://developer.nvidia.com/cudnn
Batch norm is used a lot in modern CNN architectures (see http://josephpcohen.com/w/visualizing-cnn-architectures-side-by-side-with-mxnet/ for example), so we'll also want in cudnn.

agibsonccc:
I'd be ok with the inheritance version of this for now...

saudet:
I've refactored them into helper classes, less intrusive that way alright. If that's more like it, I'll squash that before we merge into master, and I'll do the same for BatchNormalization, LocalResponseNormalization I believe as well, and maybe some others?

AlexDBlack:
I pulled this down on my machine (Win 10, 2 x GTX 970s)
Running on the LeNet MNIST example (set to 1 epoch only):
https://gist.github.com/AlexDBlack/fc115f1addc135641b0549df6ae8e837

AlexDBlack:
As for LRN: it's much lower priority than batch norm (and, RNNs too I'd say). It doesn't seem to be used much anymore. But if it's easy enough to add - sure, go ahead.

saudet:
cudaErrorMemoryAllocation? I'm allocating memory outside of ND4J's allocator, maybe that's causing problems?

AlexDBlack:
Could be. More your domain than mine. :)
Let me know if there's any other info I can provide that could make this easier to debug.

saudet:
I've updated it so that `cudaMalloc()` shouldn't get called more than a couple of times. That should take care of the error you're encountering.

AlexDBlack:
No crash anymore, ran through 10 epochs, twice.

Periodically getting the following, however:
o.d.o.l.ScoreIterationListener - Score at iteration 201 is 0.5221968203848764
o.n.j.h.i.CudaZeroHandler - Out of [DEVICE] memory, host memory will be used instead: deviceId: [0], requested bytes: [1600000]

o.d.o.l.ScoreIterationListener - Score at iteration 324 is 0.051025763885376
o.n.j.h.i.CudaZeroHandler - Out of [DEVICE] memory, host memory will be used instead: deviceId: [0], requested bytes: [2949120]
o.n.j.h.i.CudaZeroHandler - Out of [DEVICE] memory, host memory will be used instead: deviceId: [0], requested bytes: [200704]

o.d.o.l.ScoreIterationListener - Score at iteration 443 is 0.10356423587524023
o.n.j.h.i.CudaZeroHandler - Out of [DEVICE] memory, host memory will be used instead: deviceId: [0], requested bytes: [2949120]

o.d.o.l.ScoreIterationListener - Score at iteration 567 is 0.05991974060641414
o.n.j.h.i.CudaZeroHandler - Out of [DEVICE] memory, host memory will be used instead: deviceId: [0], requested bytes: [2949120]

o.d.o.l.ScoreIterationListener - Score at iteration 691 is 0.02591174758222174
o.n.j.h.i.CudaZeroHandler - Out of [DEVICE] memory, host memory will be used instead: deviceId: [0], requested bytes: [128000]

image

Memory use is consistently high it seems.

agibsonccc:
Updates on this? We'll be doing a merge within the next few days here. It'd be great to have this in there.

AlexDBlack:
Adding RNN/LSTM support probably won't make it into the next release - cudnn isn't a drop-in replacement for GravesLSTM as the cudnn LSTM variant doesn't have peephole connections. I'd like to have it (as a separate LSTM variant) in the release after the next one though.

@raver119 has been working on memory issues, which may solve those issues I'm seeing (which, on further testing: seeing on nd4j w/o cudnn too)

agibsonccc:
Let's not worry about RNNs right now. Let's just get the first cnn pass done first...

saudet:
Ok, since it's seems to be working well, I'll finish this up today or tomorrow, at least for CNNs and merge this in

PR #1741, opened 2016-06-28
Add support for cuDNN to BatchNormalization and LocalResponseNormalization

saudet:
Also adjust a couple of things with CudnnConvolutionHelper and CudnnSubsamplingHelper. Before merging this in though, please let me know if it looks alright. And with that, DL4J will support more of cuDNN than TensorFlow. Yay! :)


agibsonccc:
LGTM

PR #1759, opened 2016-06-30
Fix: squid:S2583, Conditions should not unconditionally evaluate to "…

elkfrawy-df:
This pull request is focused on resolving occurrences of Sonar rule 
 squid:S2583 - “ Conditions should not unconditionally evaluate to "TRUE" or to "FALSE" ”. 
 You can find more information about the issue here: 
 https://dev.eclipse.org/sonar/rules/show/squid:S2583

 Please let me know if you have any questions.
Ayman Elkfrawy


deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/iterator/MultipleEpochsIterator.java
```diff
@@ -211,7 +211,7 @@ public void setPreProcessor(org.nd4j.linalg.dataset.api.DataSetPreProcessor preP
      */
     @Override
     public boolean hasNext() {
-        if (newEpoch || (newEpoch && epochs == numEpochs)) {
+        if (newEpoch || epochs == numEpochs) {
             log.info("Epoch " + epochs + ", number of batches completed " + lastBatch);
```
nyghtowl: Please remove this change because the condition is incorrect if we accept anytime epochs == numEpochs to say we have achieved an epoch. If anything it should just be newEpoch for the condition.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/layers/setup/ConvolutionLayerSetup.java
```diff
@@ -147,46 +147,44 @@ private void updateLayerInputs(int i, Layer inputLayer, Layer outputLayer){
                     }
                     break;
                 case "SubsamplingLayer":
-                    if(i < lastLayerNumber){
```
nyghtowl: Please keep this conditional because it ensures that subsampling is not the last layer.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/feedforward/rbm/RBM.java
```diff
@@ -316,11 +316,11 @@ public Layer transpose() {
      * @param v the visible layer
      * @return the approximated activations of the visible layer
      */
-    public INDArray propUp(INDArray v,boolean training) {
+    public INDArray propUp(INDArray v, boolean training) {
         INDArray W = getParam(PretrainParamInitializer.WEIGHT_KEY);
         if(training) {
-            if(conf.isUseDropConnect() && training) {
-                if (conf.getLayer().getDropOut() > 0 && training) {
```
nyghtowl: Please put this conditional back because we only use DropConnect when training is true

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
```diff
@@ -847,7 +847,7 @@ public INDArray params(boolean backwardOnly) {
         List<INDArray> params = new ArrayList<>();
         for (Layer layer: getLayers()){
             INDArray layerParams;
-            if( layer instanceof BasePretrainNetwork && backwardOnly)
+            if(layer instanceof BasePretrainNetwork)
                 layerParams = ((BasePretrainNetwork) layer).paramsBackprop();
```
nyghtowl: Same reason as noted about training. This is only for backwardOnly

deeplearning4j-scaleout/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/glove/GloveWeightLookupTable.java
```diff
@@ -69,21 +69,21 @@ public void resetWeights(boolean reset) {
             this.rng = Nd4j.getRandom();
 
         //note the +2 which is the unk vocab word and the bias
-        if(syn0 == null || syn0 != null && reset) {
+        if(syn0 == null || reset) {
             syn0 = Nd4j.rand(new int[]{vocab.numWords() + 1, vectorLength}, rng).subi(0.5).divi((double) vectorLength);
```
nyghtowl: @raver119 you should check this to make sure this conditional still performs as expected

deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/iterator/MultipleEpochsIterator.java
elkfrawy-df: @nyghtowl I made these changes because they are logically wrong, and I wanted to open such discussions. 
`if (newEpoch || (newEpoch && epochs == numEpochs))` this logically has a problem, if `newEpoch` is `true` then all the if will be `true` and the if body will be executed, but if `newEpoch` is `false`, then it will go to check `(newEpoch && epochs == numEpochs)` which will always evaluate to `false`, because `newEpoch` is false!

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/layers/setup/ConvolutionLayerSetup.java
elkfrawy-df: @nyghtowl if you go up in the file, you will find that all this code executed under condition `if(i < lastLayerNumber)`, and there is nothing that changes `i` nor `lastLayerNumber` before that. So it just redundant and confusing check.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
elkfrawy-df: @nyghtowl if you look above in the same method you will find `if(backwardOnly) return params();	` which means `if( layer instanceof BasePretrainNetwork && backwardOnly)` will be executed only if `backwardOnly` was `false` which will make the whole expression always `false`!

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/feedforward/rbm/RBM.java
elkfrawy-df: @nyghtowl the check of `training` is redundant and confusing because there is already `if (training)` before that, which means `training` always `true` inside the `if` body.

deeplearning4j-scaleout/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/glove/GloveWeightLookupTable.java
elkfrawy-df: @nyghtowl again `syn0 != null` check is redundant, because if it was `null`, then the first part of the expression would be `true` and it wouldn't evaluate the second part, so it would always be `true` and again that is redundant and less readable/confusing.

deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/iterator/MultipleEpochsIterator.java
AlexDBlack: Hm, old condition is definitely wrong. if(newEpoch) as the only condition makes more sense here.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/layers/setup/ConvolutionLayerSetup.java
AlexDBlack: Change looks fine to me.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/feedforward/rbm/RBM.java
AlexDBlack: Change is logically correct, but the 3 nested if statements there isn't great.
Can just be "if(training && conf.isUseDropConnect() && conf.getLayer().getDropOut())" instead of if(){ if(){ if(){ ...}}}

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
AlexDBlack: Yep. Change there is fine.

deeplearning4j-scaleout/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/glove/GloveWeightLookupTable.java
AlexDBlack: This is fine.

deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/iterator/MultipleEpochsIterator.java
nyghtowl: Right just newEpoch

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/conf/layers/setup/ConvolutionLayerSetup.java
nyghtowl: Good catch and looks good

elkfrawy-df:
@AlexDBlack @nyghtowl Thanks for your review, I will do the suggested modifications.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/layers/feedforward/rbm/RBM.java
elkfrawy-df: @AlexDBlack Actually there is a pull request that I will make that fixes all such unnecessary nested if.

deeplearning4j-core/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
nyghtowl: @AlexDBlack relooking at this realized why we have it because BasePretrainingNewtork's method is labeled paramsBackprop(). I'm going to change it to params() to work with this PR but give a shout out if there is a reason to reverse back.

nyghtowl:
@elkfrawy-df Thanks for updating based on feedback. One of the other contributions you made this is on master is conflicting with this branch. Merge master onto this branch, resolve conflicts and then we can merge this.

elkfrawy-df:
@nyghtowl Done.

Issue #1403, opened 2016-04-14
Documentation for nIn, nOut parameters 


wfrank2509:
Hi there, 

I would like to suggest an enhancement to add some additional information about the API. 
Especially for building a CNN it was not clear to me that the nOut parameters describes the depth of the volume. Maybe it could be useful to map the usage of the API to the lecture http://cs231n.github.io/convolutional-networks/  in order to explain the usage and wheter certain parameters are applicable? E.g. nOut would be the depth of the volume which means especially the number of kernels/filters applied to the input .... or something like that? Very generic names like nIn, nOut are very ambiguous and make it a lot harder to figure out the correct usage of the api.

Thanks for the great framework!
W.

eraly:
https://github.com/deeplearning4j/deeplearning4j/pull/1827
docs in dl4j.

also adding explanation to examples
https://github.com/deeplearning4j/dl4j-0.4-examples/pull/186

wfrank2509:
👍🏾

PR #1827, opened 2016-07-14
nin/nout details for API

eraly:
https://github.com/deeplearning4j/deeplearning4j/issues/1403


eraly:
Documentation to clarify what nIn/nOut is in the context of conv nets

PR #1851, opened 2016-07-20
Use JavaDoubleRDD instead of JavaRDD<Double> to make computing sum, mean more efficient

srowen:
A tiny change for your consideration. For computing a mean or sum over an RDD of doubles, it's not necessary to pull the whole RDD back to the driver, and also count it, and compute manually. It can be computed directly in the cluster.


srowen:
Whoa, thanks for the world record fast merge!

jpatanooga:
keep these suggestions coming, we know we have a lot of ways to get better in the area of spark. thanks for the PR.

PR #1931, opened 2016-08-02
Minor fixes around dl4j

raver119:
Minor fixes around dl4j:

- changes to spark w2v tests


saudet:
Use the testresources profile, don't add it as a dependency

PR #1936, opened 2016-08-04
Remove canova from instructions

vladimir-nazarenko:
Now canova is not required for build.
@saudet


saudet:
Thanks! Actually been renamed to datavec...

saudet:
Looks good, thanks!

Issue #1328, opened 2016-03-28
toString methods serialize collections


elusive-code:
Version: 0.4-rc3.8

Some classes has toString method implementation that includes collections, which might cause signinficant performance issues, when these collections are big.
It might be an especially nasty problem when third-party libraries involved. 
And it is quite common to use toString method for logging.
Not to mention it makes IDE crawl away crying.

In my case I used word2vec on several dictionaries (including morphological: couple hundreds thousands lexemes, several million words).

In particular, when I added hibernate, at some point
`org.deeplearning4j.models.embeddings.inmemory.InMemoryLookupTable#toString`
was invoked, and took like half an hour, because it writes full vocabulary and all other collections.
(Then it crashed with OutOfMemoryException).

Classes with suspicous looking toString methods:
```
org.deeplearning4j.models.embeddings.inmemory.InMemoryLookupTable (this one definitely has a problem)

org.deeplearning4j.models.word2vec.wordstore.inmemory.InMemoryLookupCache
org.deeplearning4j.text.documentiterator.LabelledDocument

org.deeplearning4j.models.sequencevectors.sequence.Sequence
org.deeplearning4j.models.word2vec.wordstore.VocabularyWord
org.deeplearning4j.models.word2vec.VocabWord

org.deeplearning4j.berkeley.PriorityQueue
org.deeplearning4j.eval.ConfusionMatrix
org.deeplearning4j.nn.conf.MultiLayerConfiguration
org.deeplearning4j.nn.conf.MultiLayerConfiguration.Builder
org.deeplearning4j.nn.layers.BaseLayer
org.deeplearning4j.nn.layers.feedforward.autoencoder.recursive.Tree
org.deeplearning4j.util.Index
```

vladimir-nazarenko:
Would it be appropriate to return a subset of the collection from toString? Like when we got less than  1001 elements, return all of them, otherwise return first 1000 and ellipsis.

raver119:
Nah, just exclude vocabulary itself. There's no sense printing it out.

raver119:
^^^
That refers to InMemoryLookupTable obviously.

There's no practical sense in printing out vocab and syn0/syn1 fields. They just may be way to big. However, there's not too much sense using toString() on IMLT as well.. :)

vladimir-nazarenko:
@raver119, well, if I were debugging something in this class, would have been glad to see all the contents, but if you say so, I have no objections

raver119:
[...]

PR #1956, opened 2016-08-06
Fix for #1328

vladimir-nazarenko:
Attempted to fix https://github.com/deeplearning4j/deeplearning4j/issues/1328
I did my best to guess whether omitting some fields from toString is appropriate, but still I'm not sure, so be careful.


raver119:
Thanks!

PR #1960, opened 2016-08-07
properly count iterations

akhodakivskiy:
Currently HistogramIterationListener prints every single iteration to the UI. This slows training quite a bit.


deeplearning4j-ui-parent/deeplearning4j-ui/src/main/java/org/deeplearning4j/ui/weights/HistogramIterationListener.java
```diff
@@ -98,7 +101,9 @@ public void invoke() {
 
     @Override
     public void iterationDone(Model model, int iteration) {
-        if(iteration % iterations == 0) {
+        if(curIteration % iterations == 0) {
+            curIteration += 1;
```
AlexDBlack: This should be outside of the if statement, not inside? Otherwise, it'll never be updated thereafter...

akhodakivskiy: Right you are right - fixed. Thanks!

PR #1966, opened 2016-08-07
Vectorize Evaluation implementations

akhodakivskiy:
Should work much faster

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/eval/Evaluation.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/eval/Evaluation.java
@@ -184,55 +184,55 @@ public class Evaluation implements Serializable {
 
         // For each row get the most probable label (column) from prediction and assign as guessMax
         // For each row get the column of the true label and assign as currMax
+
         int nCols = realOutcomes.columns();
-        for (int i = 0; i < realOutcomes.rows(); i++) {
-            INDArray currRow = realOutcomes.getRow(i);
-            INDArray guessRow = guesses.getRow(i);
+        int nRows = realOutcomes.rows();
 
+        if (nCols == 1) {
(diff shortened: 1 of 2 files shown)


akhodakivskiy:
That was quick! I though you'd scrutinize the piece that can ruin hours of coding and watts of computing if done wrong :P

agibsonccc:
FWIW it looks good. The vectorization made the code way more straightforward actually. A lot of the minor stuff like this we tend to neglect.

deeplearning4j-core/src/main/java/org/deeplearning4j/eval/Evaluation.java
```diff
@@ -184,55 +184,55 @@ public void eval(INDArray realOutcomes, INDArray guesses) {
...
+
+            truePositives.incrementCount(1, tp);
+            falsePositives.incrementCount(1, fp);
+            falseNegatives.incrementCount(1, fp);
+            trueNegatives.incrementCount(1, tp);
+
+            truePositives.incrementCount(0, tn);
+            falsePositives.incrementCount(0, fn);
+            falseNegatives.incrementCount(0, fn);
+            trueNegatives.incrementCount(0, tn);
```
akhodakivskiy: @agibsonccc I'd appreciate if you could take closer look at these 8 lines above. It's passing the test (which doesn't cover everything), but I'm not sure if it's doing the right thing.

agibsonccc: FWIW those tests aren't bad. The case here is right. This is basically the special binary class case. It's pretty straight forward to run the whole classification using the gt mask rather than the loop.  You are replacing he for loop that we were doing here with the series of equivalent arithmetic operations.

PR #2003, opened 2016-08-11
Tsne

Avricot:
The main changes are :

- BarnesHutTsne doesn't extend Tsne anymore

- WeightLookupTable expect a BarnesHutTsne and build a BarnesHutTsne by default instead of a Tsne

- theta is set to 0.5 by default

- numdim isn't passed to the plot/fit method anymore, I've put the old as deprecated


Avricot:
Ok, I've just rebased against the master no more conflicts

PR #2032, opened 2016-08-25
Custom Layer Support

AlexDBlack:
***Complete; can be merged any time***

This PR adds custom layer support to DL4J, without requiring users to modify DL4J core code.

Changes of note:
* Remove LayerFactory
* Add Layer (config) instantitate method: used for config -> implementation when creating layers
* Update MultiLayerNetwork, ComputationGraph and all unit tests for the above
* Handle JSON serialization/deserialization for custom layers
    - This is done using reflection: scan the classpath for concrete classes that extend the Layer (configuration) class, and registering subtypes with Jackson ObjectMappers where required.
* Tests for custom layer functionality; including custom output layer and Spark tests
* Introduced two new interfaces:
    - IOutputLayer extends Layer: has methods specialized to output layers (setLabels, etc). Allows users to implement output layers without being forced to extend BaseOutputLayer (that we use for OutputLayer and RnnOutputLayer)
    - RecurrentLayer: has methods specialized to RNN layers - tBPTT methods, rnnTimeStep, etc. Allows users to implement RNN layers without having to extend BaseRecurrentLayer
    - Also updated MultiLayerNetwork/ComputationGraph to use these interfaces where approapriate, instead of BaseOutputLayer and BaseRecurrentLayer

DL4J unit tests (local and spark) are passing; examples also seem fine.

Example using custom layer functionality: https://github.com/deeplearning4j/dl4j-examples/pull/224


marker68:
Thank you. An awesome work!

PR #2063, opened 2016-09-02
Various fixes/improvements

raver119:
***WIP; DO NOT MERGE***

Various bugfixes and improvements:
- static assets for UI, instead of externally linked
- fix for CNN listener on non-RGB input


deeplearning4j-nn/src/main/java/org/deeplearning4j/util/ModelSerializer.java
```diff
@@ -46,6 +47,8 @@ private ModelSerializer() {
     public static void writeModel(@NonNull Model model, @NonNull File file, boolean saveUpdater) throws IOException {
         try(BufferedOutputStream stream = new BufferedOutputStream(new FileOutputStream(file))){
             writeModel(model, stream, saveUpdater);
+            stream.flush();
+            stream.close();
```
reckart: It shouldn't be necessary to explicitly call close because you're using a try-with-resource here, right?

raver119: Actually yes, both classes implement autocloseable

PR #2101, opened 2016-09-12
Added nearestLabel(INDArray labelVector, int topN) implementation

juvaud:
now vec.extractLabels() has to be called after vec.fit()
Please have a look on the imports and the implementation might not be very fast. 
Tests are not implemented as the classification often delivers different results (due to only 3 epochs).
@raver119


raver119:
Cool, thanks, i'll build proper method based on this.

Will merge a bit later though

juvaud:
You're welcome and thanks for your help. I'm curious how the optimized version will look like.

raver119:
I've transferred your extractLabels() code into codebase + provided search without use of Tree.

PR #2129, opened 2016-09-24
Switch to shaded version of dropwizard (and deps)

AlexDBlack:
Local testing complete. Requires final testing with Spark + Spark Submit.


deeplearning4j-ui-parent/deeplearning4j-ui/pom.xml
```diff
@@ -50,38 +50,46 @@
...
-        <dependency>
-            <groupId>io.dropwizard</groupId>
-            <artifactId>dropwizard-forms</artifactId>
-            <version>${dropwizard.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.glassfish.jersey.media</groupId>
-            <artifactId>jersey-media-multipart</artifactId>
-            <version>2.16</version>
+            <groupId>org.nd4j</groupId>
```
huitseeker: Oh, explicitly shading. Rather than the shade plugin. OK. You might want to leave a line on the insufficiencies of the plugin in README so that sbdy down the line doesn't try to switch to that (you didn't comment much on that in https://github.com/deeplearning4j/deeplearning4j/pull/2122 either).
Where are the scripts that produce those BTW ?
Have you looked at [this related approach for producing them](https://github.com/pwendell/spark-shaded) ?

deeplearning4j-ui-parent/deeplearning4j-ui/pom.xml
```diff
@@ -50,38 +50,46 @@
             <version>${nd4j.version}</version>
             <scope>test</scope>
         </dependency>
+        <!--<dependency>-->
```
huitseeker: Are you sure you wanna keep the old deps around in such a  .. verbose form ?

AlexDBlack:
Stalled on account of logback logging dependencies in dropwizard. Closing for now.

PR #2143, opened 2016-09-28
Add additional UI/stats collection functionality

AlexDBlack:
***WIP DO NOT MERGE***


saudet:
BTW @raver119 have we switched to CUDA 8.0 by default in ND4J? If so, let's do the same in DL4J.

PR #2159, opened 2016-10-05
fix build and termination condition

akhodakivskiy:


treo:
from Gitter:
> @akhodakivskiy 
> bestEpoch is never reset between runs in Arbiter
> And the same instance is reused over and over

And the pom packaging option for dl4j-ui resulted in no jars being build.

PR #2166, opened 2016-10-08
Remove last org.nd4j.linalg.lossfunctions.LossCalculation usage

treo:
As far as I can tell this is the last place where `org.nd4j.linalg.lossfunctions.LossCalculation` was still used. 

After replacing it pretraining an AutoEncoder started to work :)


nyghtowl:
The code looks sound. The gradient check test for AutoEncoder is failing (https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTests.java#L748). Tracked it to LossKLD computeScore creating a scoreArr where values are too small to pass in a score that has value beyond NaN. Not immediately clear how to resolve.

treo:
Great, then that means that I wasn't missing something on LossKLD, as it ran straight into NaNs for me as well.

treo:
@nyghtowl can you rerun the gradient check with https://github.com/deeplearning4j/nd4j/pull/1340 applied? I'm not quite sure about the gradient there, so it might still fail.

nyghtowl:
@treo nd4j#1340 did fix the NaN score issue. All gradients pass except where tanh is used as activation on all layers including output and MSE is the output loss function. Merging this and will update test separately

PR #2180, opened 2016-10-15
Added new continuous labeled point static methods. This is for Regression

dmmiller612:
I have been using the methods in a local project of mine, and I figured that I would submit it to the core repo. This is specifically for labeled points with continuous features. The other methods run into issues when trying to use continuous features.


dmmiller612:
Was unsure if this would warrant a pom bump. Let me know, if you would like that to happen.

agibsonccc:
Thanks!

Issue #2187, opened 2016-10-18
LICENSE.txt should contain a copy of the AL 2.0


kottmann:
The LICENSE.txt file should contain a copy of the Apache License 2.0 and not the copyright header.

PR #2188, opened 2016-10-18
#2187 Replace copyright header with AL 2.0

kottmann:


kottmann:
Ups, this was the wrong file!

PR #2202, opened 2016-10-20
nn modular configuration.

crockpotveggies:
# WIP do not merge!

This branch is for work being performed on modules for generating layers in nn models, allowing for extensibility of layer generation.

This branch will evolve and change as authors see fit. Do not merge until it's ready.


crockpotveggies:
Updated. I have a couple considerations. The first is, how should we handle configuration? Should that be extended from a generic Config type? I think each module should handle its own configuration. Second, is it necessary to allow the user to implement `n + 2` parallel columns?

agibsonccc:
Look closer at the computation graph. Modules should just be vertices that
allow nested comp graphs.

crockpotveggies:
@agibsonccc are you referring to the second question re: `n + 2` columns? Just verifying.

crockpotveggies:
Merging into https://github.com/deeplearning4j/deeplearning4j/issues/2224.

Issue #2216, opened 2016-10-26
Add TripletLearningAlgorithm to DL4J


crockpotveggies:
Opening this issue to centralize discussion for adding a `TripletLearningAlgorithm` to Deeplearning4j. Triplet Embeddings have been shown to be extremely useful for calculating euclidean distance for classification and clustering, and personally solves a lot of business problems we have at Bernie AI.

I'd like to implement Triplet Embedding in DL4J and generalize it to `MultiLayerNetwork` and `ComputationGraph` so that it can also be shared with future users. The purpose of this issue is to consult with authors to understand the architecture of what I'm implementing, and ensure that it is planned correctly. As per Gitter chat, sounds like @raver119 and @AlexDBlack are the two helpful people to provide input.

![image](https://cloud.githubusercontent.com/assets/1445295/19708729/a1a4ce38-9ad5-11e6-98c3-562f077b730a.png)

Rough strategy for implementation:
1. Create a `TripletLearningAlgorithm` in `org.deeplearning4j.models.embeddings.learning`
2. Base some of the work using embeddings via @raver119 's `ElementsLearningAlgorithm`
3. Write a `TripletLoss` function in ND4J
4. Implement triplet generation in DataVec so datasets are automatically preprocessed (maybe this should be called `org.datavec.common.TripletConverter`?)
5. What else is needed if we want this to generalize to MLN and CG?

Related literature using Triplet Embeddings:
[FaceNet](https://arxiv.org/pdf/1503.03832v3.pdf)
[OpenFace](http://reports-archive.adm.cs.cmu.edu/anon/2016/CMU-CS-16-118.pdf)

Similar implementations include:
[Torch's TripletEmbedding](https://github.com/Atcold/torch-TripletEmbedding/blob/742599ce761b2d6cf128068b8f6fd327ad633a31/TripletEmbedding.lua) and [an example](https://github.com/Atcold/torch-TripletEmbedding/blob/742599ce761b2d6cf128068b8f6fd327ad633a31/xmp/fresh-embedding.lua)
[Caffe TripletLossLayer](https://github.com/deepxkn/triplet/blob/master/triplet/tripletloss_layer.py)

Your thoughts and ideas here will help me implement this, thanks!

AlexDBlack:
OK, so I've been reading this paper: https://arxiv.org/pdf/1412.6622v3.pdf

So, as I understand it:
- 3 inputs: x-, x, x+ where:
    - x and x+ are of the same class
[...]

PR #2224, opened 2016-10-27
Triplet Embedding functionality

crockpotveggies:
# WIP do not merge!

This PR addresses https://github.com/deeplearning4j/deeplearning4j/issues/2216 for adding Triplet Embedding functionality to DL4J.


AlexDBlack:
On the whole, looks pretty good. Some comments:
- Name: "SoftOutputLayer"	->	maybe we could call this LossLayer? I'm open to suggestions though.
- The lossFunction field - can be removed (same thing in BaseOutputLayer actually - feel free to remove that too while you are at it ).
- Use EmptyParamInitializer instead of DefaultParamInitializer
- SoftOutputLayer.instantiate: return a SoftOutputLayer/LossLayer (instance) not an output layer. Get rid of the params table, params view array stuff.
- Remove the customLossFunction(String) method on the builder.
- backpropGradient/getGradientsAndDelta: shouldn't have any gradient calculation or parameters in there... pretty sure you just want the gradient (which should be empty) and delta (triple.getSecond())
- Output method: just apply the activation function - don't call super.activate
- fit(DataSetIterator) should be a no-op
- definitely need some gradient checks... maybe just copy + paste lossFunctionGradientCheck() (see link below) and change output layer to another dense layer, and add the SoftOutputLayer/LossLayer as layer 2? That should be enough.

https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/LossFunctionGradientCheck.java#L45

crockpotveggies:
I've updated everything on your comments except for the gradient checks. Were you saying that I need to add a test spec for the gradient checks? Or are you thinking those should be included directly in the layer?

Regarding loss function, are we still going to need a `TripletLossFunction`?

I'll be adding more code later today for GraphVertex stuff.

AlexDBlack:
Re: gradient checks - yeah, I was suggesting adding a unit test using LossLayer, like the ones in LossfunctionGradientCheck.
Some basic forward/backward pass sanity checks (unit tests) wouldn't hurt either.

As for the TripletLossFunction - as per the issue discussion - I don't think we need it.
Instead of having a TripletLossFunction we'd have the "comparator" (which is LossLayer) and a two L2 vertices (and, technically  speaking a merge vertex) just below it.
The TripletLossFunction just combines all that into one unit.

Now, we *could* in principle do a merge on the unstacked network outputs (so hstack the 3 row vectors) then do a custom loss function that extracts out the relevant subsets - but to me, the l2 vertex + LossLayer is cleaner/easier to understand.


image


More comments:
- Don't remove the getLossFunction() method in BaseOutputLayer - but you can remove customLossFunction field. We'll get rid of that at 0.7.0 I think
- @JsonSubTypes.Type(value = LossLayer.class, name = "softoutput"), -> "loss" or "losslayer" perhaps
- not sure output/activate methods are correct... I'd expect to see a Nd4j.getExecutioner().execAndReturn(...) in there somewhere

Otherwise latest changes look good.

crockpotveggies:
Alright just want to make sure I'm interpreting you correctly, our graph structure is going to look like this:

triplet computation graph

I've shared the link since it sounds like we've changed those last layers, but I don't quite follow how the DenseLayer is going to then unstack and help with our L2 -> softmax.

AlexDBlack:
I was  thinking more along the lines of this:
image

LossLayer only at the end, as it calculates gradients with respect to labels (hence can't be in middle of the graph)

AlexDBlack:
image

crockpotveggies:
Added a basic L2 function, I'm not quite sure if I've done this correctly. I'm struggling to figure out how I can implement derivatives in ND4J. Can I get your help there?

AlexDBlack:
It's a bit rough, but I think this is basically what you want.

```java
public static void main(String[] args){

        Nd4j.getRandom().setSeed(12345);
        int miniBatch = 5;

        //+++ 2d case (MLPs) +++
        INDArray a = Nd4j.rand(miniBatch,10);
        INDArray b = Nd4j.rand(miniBatch,10);

        //+++ 3d case (RNNs) +++
//        int tsLength = 6;
//        INDArray a = Nd4j.rand(new int[]{miniBatch, 10, tsLength});
//        INDArray b = Nd4j.rand(new int[]{miniBatch, 10, tsLength});

        //+++ 4d case (CNNs) +++
//        int depth = 3;
//        int h = 10;
//        int w = 10;
//        INDArray a = Nd4j.rand(new int[]{miniBatch, depth, h, w});
//        INDArray b = Nd4j.rand(new int[]{miniBatch, depth, h, w});



//        System.out.println(x);
//        System.out.println(y);


        //Forward pass: l2 (euclidean) distance for each example (i.e., along all dimensions except 0)
        int[] dimensions = new int[a.rank()-1];
        for( int i=1; i<a.rank(); i++ ){
            dimensions[i-1] = i;
        }

        System.out.println("Reduce along dimensions: " + Arrays.toString(dimensions));
        INDArray out = Nd4j.getExecutioner().exec(new EuclideanDistance(a,b),dimensions);
        System.out.println(out);
        System.out.println("Output shape: " + Arrays.toString(out.shape()));       //minibatch, 1




        //Backprop:
        INDArray dLdlambda = Nd4j.rand(out.shape());      //dL/dlambda aka 'epsilon' - from layer above

        INDArray sNegHalf = out.rdiv(1.0);      //s^(-1/2) = 1.0 / s^(1/2) = 1.0 / out

        INDArray diff = a.sub(b);

        INDArray first = dLdlambda.mul(sNegHalf);   //Column vector for all cases

        INDArray dLda;
        INDArray dLdb;
        if (a.rank() == 2){
            //2d case (MLPs etc)
            dLda = diff.muliColumnVector(first);
            dLdb = dLda.neg();
        } else {
            //RNN and CNN case - Broadcast along dimension 0
            dLda = Nd4j.getExecutioner().execAndReturn(new BroadcastMulOp(diff,first,diff,0));
            dLdb = dLda.neg();
        }


//        System.out.println(dLda);
//        System.out.println(dLdb);
    }
```

crockpotveggies:
Merging https://github.com/deeplearning4j/deeplearning4j/pull/2202 into this PR.

crockpotveggies:
@AlexDBlack Finally a challenging bug. Backprop is failing from LossLayer, I'm not sure why. I can't follow the stack trace because it's calling `BaseOutputLayer` which is something I haven't explicitly implemented. Is there some sort of typecasting going on in CG?

```
 java.lang.NullPointerException
    at org.nd4j.linalg.api.ndarray.BaseNDArray.mmul(BaseNDArray.java:2587)
    at org.deeplearning4j.nn.layers.BaseLayer.preOutput(BaseLayer.java:354)
    at org.deeplearning4j.nn.layers.BaseOutputLayer.preOutput2d(BaseOutputLayer.java:397)
    at org.deeplearning4j.nn.layers.BaseOutputLayer.backpropGradient(BaseOutputLayer.java:148)
    at org.deeplearning4j.nn.graph.vertex.impl.LayerVertex.doBackward(LayerVertex.java:114)
    at org.deeplearning4j.nn.graph.ComputationGraph.calcBackpropGradients(ComputationGraph.java:1117)
    at org.deeplearning4j.nn.graph.ComputationGraph.computeGradientAndScore(ComputationGraph.java:892)
    at org.deeplearning4j.gradientcheck.GradientCheckUtil.checkGradients(GradientCheckUtil.java:202)
    at org.deeplearning4j.gradientcheck.GradientCheckTestsComputationGraph.testBasicIrisTripletStackingL2Loss(GradientCheckTestsComputationGraph.java:478)
```

I have no idea why BaseOutputLayer is being called, it doesn't make any sense because LossLayer doesn't implement it. 
```
public class LossLayer extends BaseLayer<org.deeplearning4j.nn.conf.layers.LossLayer> implements Serializable, IOutputLayer
```
The only similarity is `IOutputLayer` so there's something going on here I don't understand.

PR #2226, opened 2016-10-27
fix Korean & Japanese nlp modules

kepricon:
fix Korean & Japanese nlp modules

### DIFF
--- a/deeplearning4j-nlp-parent/deeplearning4j-nlp-japanese/pom.xml
+++ b/deeplearning4j-nlp-parent/deeplearning4j-nlp-japanese/pom.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>deeplearning4j-nlp-parent</artifactId>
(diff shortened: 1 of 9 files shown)


djKooks:
Looks good to me~

wmeddie:
One thing to be careful with Kuromoji 0.7.7 is that it's not available in Maven Central.  And Atilika's nexus instance does tend to go down from time to time.  That said 0.9.0 might not be ready for prime time and has a slightly different interface.  Just something to keep an eye on.

agibsonccc:
We could host our own bintray version if need be. Thanks for the warning.

marker68:
Confirmed the issues @wmeddie said:

```
[ERROR] Failed to execute goal on project deeplearing4j-nlp-japanese: Could not resolve dependencies for project org.deeplearning4j:deeplearing4j-nlp-japanese:pom:0.6.1-SNAPSHOT: Could not find artifact org.atilika.kuromoji:kuromoji:jar:0.7.7 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
```

kepricon:
WIP
kuromoji 0.7.7 is not available on maven central repository

we need to change deeplearing4j-nlp-japanese/pom.xml  

the below dependency might work.
<dependency>
  <groupId>com.atilika.kuromoji</groupId>
  <artifactId>kuromoji-ipadic</artifactId>
  <version>0.9.0</version>
</dependency>

but we need to change apis.(there is a difference between 0.7.7 apis and 0.9.0)
see : https://github.com/atilika/kuromoji

wmeddie:
I have a repository using the new apis. Feel free to use it as an example. 

https://github.com/wmeddie/dl4j-jp

agibsonccc:
Thanks. We will probably bundle then. @kepricon

PR #2233, opened 2016-10-28
fix Korean & Japanese Tokenizer

kepricon:
update kuromoji tokenizer from 0.7.7 to 0.9.0
apply modified apis due to update(kuromoji)


kepricon:
will embed kuromoji 0.9.0 to deeplearning4j-nlp-japanese module

PR #2263, opened 2016-11-02
Multi-token items in EmbeddingLayer

vtablan:
Changed embedding layer to also support multi-token examples (e.g. as used for CNN-based sentence classification). An example could be a document (a sequence of words) represented as a column vector of integer word IDs.

### DIFF
--- a/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTests.java
+++ b/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTests.java
@@ -2,14 +2,19 @@ package org.deeplearning4j.gradientcheck;
 
 import org.deeplearning4j.datasets.iterator.impl.IrisDataSetIterator;
 import org.deeplearning4j.nn.api.OptimizationAlgorithm;
+import org.deeplearning4j.nn.conf.ComputationGraphConfiguration;
 import org.deeplearning4j.nn.conf.MultiLayerConfiguration;
 import org.deeplearning4j.nn.conf.NeuralNetConfiguration;
 import org.deeplearning4j.nn.conf.Updater;
 import org.deeplearning4j.nn.conf.distribution.NormalDistribution;
 import org.deeplearning4j.nn.conf.distribution.UniformDistribution;
+import org.deeplearning4j.nn.conf.graph.PreprocessorVertex;
 import org.deeplearning4j.nn.conf.inputs.InputType;
 import org.deeplearning4j.nn.conf.layers.*;
+import org.deeplearning4j.nn.conf.preprocessor.ReshapePreProcessor;
 import org.deeplearning4j.nn.conf.preprocessor.RnnToCnnPreProcessor;
+import org.deeplearning4j.nn.graph.ComputationGraph;
+import org.deeplearning4j.nn.graph.vertex.GraphVertex;
 import org.deeplearning4j.nn.multilayer.MultiLayerNetwork;
 import org.deeplearning4j.nn.weights.WeightInit;
 import org.junit.Ignore;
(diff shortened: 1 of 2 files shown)


AlexDBlack:
Had a (very) quick look - what's the assumption here? fixed length for all inputs, right? (And then existing/current embedding layer implementation is just a special case of that, with 1 input per example).

For variable length, we'll need to do some more work to support per-input masking situations (which would be more work than just modifying the embedding later). I'd be fine doing that later though.

vtablan:
Yes, the idea is that documents get padded to the same length, so that the input is of the same size. That was just simpler for me to do (given that I've only started with DL4J). However, the embeddings go straight into a set of parallel convolutions (whose weights dimensions depend only on filter and embeddings sizes) followed by max-pooling over the entire document length. This means that the relevance of the document length is only localised to the data that gets passed from the convolutions to the max-pooling layer. Anyway, as you say, this can wait for a while.

Here is [an example that exercises this code](https://github.com/vtablan/dl4j-examples/blob/master/dl4j-examples/src/main/java/org/deeplearning4j/examples/nlp/cnntextclassification/CNNTextClassifier.java) which tries to replicate this [blog post](http://www.wildml.com/2015/12/implementing-a-cnn-for-text-classification-in-tensorflow/) which does the same thing with tensorflow.

BTW, when reviewing please pay attention to the backprop step - I don't fully understand what that does, so I had to make some  guesses. In linear algebra, if the dimensions work it's usually fine, but still... ;).

Thanks!

vtablan:
BTW, are any of those test failures related to my code? I don't think I have access to the CI server so I can't  see the actual output.

agibsonccc:
No dont worry about it.

AlexDBlack:
@vtablan thanks. So the best way to check this: implement an extra gradient check unit test...
https://github.com/deeplearning4j/deeplearning4j/blob/38c7fd80c4287b945ee1a9adf2298cdc957f609c/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTests.java#L226

Basically, take that existing embedding layer gradient check test, but change the input data size to test your new code. Not much in the way of documentation on gradient checks, but hopefully that should be straightforward to extend...

vtablan:
@AlexDBlack thanks Alex. I've added a new test along the lines of the one you pointed out. I had to change the config to use a ComputationGraph because I had to resize the input to the output layer. The new test succeeds on my machine (though there is one test in that class that fails for me  - testGradientCnnFfRnn).

BTW - why is ReshapePreprocessor deprecated, and what's its replacement? No one seemed to be able to answer that on gitter.

Thanks!

AlexDBlack:
Hey, sorry for not reviewing/replying earlier - been flat out with stuff.

To be honest: I'm not really happy with the modifications, now that I've actually step through it in the debugger.
You are feeding in 3d - time series - data into the embedding layer, and outputting a time series.
This is something we already support via RnnToFeedForwardPreprocessor -> EmbeddingLayer -> FeedForwardToRnnPreprocessor. That approach is more robust than what you are doing here too (more careful reshaping, etc). That's the same approach we use for Dense and CNN layers with RNNs.
What I had initially assumed you were doing is basically approximating a sparse input matrix (i.e., multiple in, 1 out) - not extending it to handle the time series case directly.

Maybe try doing it that way first, and if that doesn't do what you need, we'll revisit this?

vtablan:
Cool - I'll look at that approach and see if I can make it work. Thanks for the pointer!

(closing this pull request)

vtablan:
Hi,

I was able to re-engineer my code to use the RnnToFF preprocessor - thanks for the tip. I still needed to usethe bare Reshape processor at some point, because I needed to turn the time dimension into 'height', and there was no other preprocessor that did that.

You can see some code that illustrates this [in my examples fork](https://github.com/vtablan/dl4j-examples/blob/master/dl4j-examples/src/main/java/org/deeplearning4j/examples/nlp/cnntextclassification/CNNTextClassifier.java). 

Is there a plan to provide a replacement for Reshape processor?

PR #2333, opened 2016-11-18
Fix loss function on LossLayer builder.

crockpotveggies:
Currently the loss function parameter doesn't properly set the loss function. This PR fixes that.


AlexDBlack:
Confirmed issue; better fix implemented here: https://github.com/deeplearning4j/deeplearning4j/pull/2335

PR #2358, opened 2016-11-23
UI - Model scoring best fit line

crockpotveggies:
Sometimes noise created by high peak amplitudes in model scoring can obscure general intuition as to whether a model is performing well or not. This commit addresses this problem by adding a "best fit" summary line to help the user understand if overall score is *actually* decreasing.

screen shot 2016-11-22 at 6 58 11 pm


crockpotveggies:
Adds a best fit line on selected range. Otherwise, uses EMA.

screen shot 2016-11-22 at 9 44 37 pm

PR #2419, opened 2016-12-01
Fix for ParallelWrapper not working with MDS

fac2003:
ParallelelWrapper support for multi dataset iterators was unfinished and probably never tested. This fix correctly initializes the worker for MDS when the model is a computational graph, and loops over datasets in the queue instead of processing one and waiting ad infinitum.


raver119:
Bugs are real, fix isn't proper. New fix is incoming, as discussed in gitter.

Once again, thanks for highlighting this issue!

PR #2441, opened 2016-12-05
Add P-norm pooling to SubsamplingLayer.

crockpotveggies:
`PNORM` setting is a direct port of Torch's `SpatialLPPooling` functionality: https://github.com/torch/nn/blob/master/SpatialLPPooling.lua

Literature can be found here: http://yann.lecun.com/exdb/publis/pdf/bruna-icml-14.pdf

Suggest @AlexDBlack review the PR.


AlexDBlack:
Merged here: https://github.com/deeplearning4j/deeplearning4j/pull/2464

PR #2455, opened 2016-12-07
Create zh-simple-image-load-transform.md

yanbowang:


cn/zh-simple-image-load-transform.md
```diff
@@ -0,0 +1,128 @@
+---
+title: 用定制的数据加工管道将图像载入深度神经网络
+layout: default
```
gsw85: cn-default

PR #2473, opened 2016-12-11
Add DropoutLayer + change default activation for LossLayer

turambar:
A collection of commits related to LossLayer and the proposed DropoutLayer:

* Default activation for LossLayer is now `identity` so that calling `model.output()` gives the activation from the previous layer, not the loss.
* Added separate `DropoutLayer` (files: `nn/layers/DropoutLayer.java` and `nn/conf/layers/DroputLayer.java`).
* Added unit tests for `DropoutLayer`, similar to ActivationLayer unit tests.
* Minor change to `BaseLayer`: call to `applyDropoutIfNecessary` duplicates input so that the in-place dropout doesn't overwrite output of previous layer (in, e.g, calls to `model.feedForward()`.

### DIFF
--- a/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/DropoutLayerTest.java
+++ b/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/DropoutLayerTest.java
@@ -1,0 +1,236 @@
+package org.deeplearning4j.nn.layers;
+
+import org.deeplearning4j.datasets.iterator.impl.MnistDataSetIterator;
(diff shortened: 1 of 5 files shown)


turambar:
Open request to one or more of @AlexDBlack @nyghtowl @agibsonccc: please give this a once over to make sure everything looks OK before I merge.

turambar:
I think I overhead that we're doing a release this week? If so, I'd like to get this in. Passes its unit tests (with same level of testing thoroughness as ActivationLayer).

turambar:
@nyghtowl Cool. I'm going to merge this and then I'll fix the DropOutTest comments in a separate commit.

PR #2481, opened 2016-12-12
Fixes to ParallelWrapper and EarlyStoppingParallelTrainer

crockpotveggies:
Add ability to stop ParallelWrapper in mid-fit. Fix NPEs from empty zoo. Sanity checks. Fix tests.


AlexDBlack:
looking at the changes to ComputationGraph and MultiLayerNetwork - not sure how much I like adding 2 fields and 4 methods to both, just for combination of parallel wrapper + stats listener...

TBH I'd rather we keep the same design as in Spark: each worker would have a copy of the listener (probably cloned), and reports each worker's progress separately (a dropdown box will appear as in spark)
that also solves the missing update ratios issue - that's one of the key pieces of info for tuning, so it's important we have it IMO.

agibsonccc:
Agreed with alex here. The idea with parallelwrapper is to be a "local version of whatever we do with spark"

crockpotveggies:
Understood, also chatted it out over Gitter with Alex and I can make the appropriate changes.

agibsonccc:
@crockpotveggies I moved parallelwrapper (same package name etc, different maven module) Could you redo your conflicting changes here so I can merge this?

crockpotveggies:
Not a problem I'll get this done today.

PR #2484, opened 2016-12-13
Major refactor of Keras model import to support Functional API -> ComputationGraph

turambar:
Near rewrite of Keras model import module (which retains critical backwards compatibility). Witness:

- `KerasLayer` class for processing Keras layer configurations and building DL4J `Layer`s
- `KerasModel` class for processing Keras model configurations and building `ComputationGraph`s (for Functional API) and `MultiLayerNetwork`s (for Keras `Sequential`)
- `KerasModelImport` class for reading Keras model archive files (HDF5 archives for models and weights, text files for JSON configurations)

Backwards compatibility is maintained via the high-level, static model and configuration import functions in `Model` and `ModelConfiguration`, respectively. These work exactly as before but have been deprecated so we can remove them in some future release.

Main new feature is support for Keras the Functional API `Model`, which maps to DL4J `ComputationGraph`. We can also now import Keras loss functions (from training configuration) and add a corresponding loss layer. We also added a few additional layers.

Perhaps most important, new code is much more modular and readable and should be much easier to maintain, debug, and develop. Everything is pretty thoroughly commented and documented, including TODOs.


turambar:
I think I overhead that we're doing a release this week? If so, I'd like to get this in. Passes all existing unit tests and extends existing functionality. Should allow us to start moving much faster on expanding coverage of architectures ([Vgg](https://github.com/fchollet/deep-learning-models/blob/master/vgg16.py) and [Inception](https://github.com/fchollet/deep-learning-models/blob/master/inception_v3.py) being my main targets).

deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/DropoutLayerTest.java
```diff
@@ -0,0 +1,236 @@
...
+                .list()
+                .layer(0, new ConvolutionLayer.Builder(4, 4).stride(2, 2)
+                        .nIn(1).nOut(20).activation("relu")
+                        .weightInit(WeightInit.XAVIER).build())
+                .layer(1, new OutputLayer.Builder(LossFunctions.LossFunction.MCXENT)
+                        .weightInit(WeightInit.XAVIER)
+                        .activation("softmax")
+                        .dropOut(0.25)
+                        .nOut(10).build())
+                .backprop(true).pretrain(false)
```
agibsonccc: set a seed on the network for reproducible tests?

deeplearning4j-modelimport/pom.xml
```diff
@@ -125,5 +125,19 @@
             <artifactId>logback-classic</artifactId> <!-- Version set by deeplearning4j-parent dependency management -->
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.deeplearning4j</groupId>
+            <artifactId>deeplearning4j-core</artifactId>
+            <version>0.7.1-SNAPSHOT</version>
```
agibsonccc: Version is out of date. ALWAYS use ${project.version} Please do a git pull and update your versions if you haven't already.

deeplearning4j-modelimport/pom.xml
```diff
@@ -125,5 +125,19 @@
...
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.nd4j</groupId>
+            <artifactId>nd4j-native</artifactId>
+            <version>${nd4j.version}</version>
```
agibsonccc: This is usually project.version. For consistency, you can use nd4j.version though. Just keep this in mind for future projects.

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/InvalidKerasConfigurationException.java
```diff
@@ -0,0 +1,43 @@
...
+ * Indicates that user is attempting to import a Keras model configuration that
+ * is malformed or invalid in some other way.
+ *
+ * See http://deeplearning4j.org/model-import-keras for more information.
+ *
+ * @author dave@skymind.io
+ */
+public class InvalidKerasConfigurationException extends RuntimeException {
+
+    public InvalidKerasConfigurationException(String message) { super(appendDocumentationURL(message)); }
```
agibsonccc: Not sure how big of a fan I am of extending runtime exception. This is a special unchecked exception. Maybe we should  just do a normal checked exception here?

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/KerasSequentialModel.java
```diff
@@ -0,0 +1,172 @@
...
+        /* Handle truncated BPTT:
+         * - less than zero if no recurrent layerNamesOrdered found
+         * - greater than zero if found recurrent layer and truncation length was set
+         * - equal to zero if found recurrent layer but no truncation length set (e.g., the
+         *   model was built with Theano backend and used scan symbolic loop instead of
+         *   unrolling the RNN for a fixed number of steps.
+         *
+         * TODO: should we not throw an error for truncatedBPTT==0?
+         */
+        if (this.truncatedBPTT == 0)
```
agibsonccc: @AlexDBlack

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/UnsupportedKerasConfigurationException.java
```diff
@@ -0,0 +1,44 @@
...
+ * Indicates that user is attempting to import a Keras model configuration that
+ * is not currently supported.
+ *
+ * See http://deeplearning4j.org/model-import-keras for more information and
+ * file an issue at http://github.com/deeplearning4j/deeplearning4j/issues.
+ *
+ * @author dave@skymind.io
+ */
+public class UnsupportedKerasConfigurationException extends RuntimeException {
+
```
agibsonccc: Again same thing. I'm thinking we extend a checked exception here.

deeplearning4j-modelimport/src/test/java/org/deeplearning4j/nn/modelimport/keras/ModelTest.java
```diff
@@ -0,0 +1,66 @@
...
+ * TODO: Replace deprecated stuff and rename to something like KerasModelTest
+ * TODO: Move test resources to dl4j-test-resources
+ * TODO: Reorganize test resources
+ * TODO: Add more extensive tests including exotic Functional API architectures
+ *
+ * @author dave@skymind.io
+ */
+public class ModelTest {
+    private static Logger log = LoggerFactory.getLogger(ModelTest.class);
+
```
agibsonccc: Use @Slf4j instead. Lombok will generate the logger declaration.

deeplearning4j-modelimport/pom.xml
turambar: Oops, that's a laziness mistake. Will push a fixing commit shortly.

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/InvalidKerasConfigurationException.java
turambar: I will defer to you on this. I chose arbitrarily.

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/UnsupportedKerasConfigurationException.java
turambar: I will defer to you on this as well.

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/KerasSequentialModel.java
turambar: Yeah basically this raises the question of what the proper behavior should be when trying to import a Keras-defined RNN that does not use truncated backprop (not unusual with Theano backend). Should we
  - throw an error (which makes problem obvious to user, who can then go back and change their Keras configuration)?
  - choose some reasonable default BPTT value ourselves and maybe mention this in a warning?
  - NOT set truncated BPTT in the DL4J model configuration? What exactly would the behavior be for this (I haven't tried it in DL4J yet)?

deeplearning4j-modelimport/pom.xml
turambar: Leaving this as is for now since it seems consistent with other deeplearning4j modules (e.g., `deeplearning4j-nn`).

turambar:
@agibsonccc I made the trivial fixes: `${project.version}` in `pom.xml`, switched to `@Slf4j` for logging, changed Keras-specific exceptions to checked exceptions (and added all necessary throws and try/catches). I also changed `KerasModel` and `KerasSequentialModel` to Builder pattern constructors.

turambar:
Added support for YAML-format model config files. Simplified Builder stuff: now there's just one nested ModelBuilder class and you use `buildModel()` to build a ComputationGraph (for Functional API) and `buildSequential()` to build a MultiLayerNetwork (for Sequential model).

turambar:
More commits, including
- support for BatchNormalization layer
- fixes to loading of LSTM weights
- support for importing Keras model configurations from InputStreams
- stubbed out constructor for import Keras model+weights from InputStream (containing HDF5 archive bytes)

PR #2562, opened 2016-12-22
Keras backend - support for Theano sequential models

pkoperek:
This is a very basic version of the Server for dl4j for Keras integration. It has some shortcomings:
* different dimension ordering (Tensorflow vs Theano) isn't handled
* only sequential model is supported for now
* there is no feedback sent back to python
* not sure what should be a good automated test here (i just used a simple script from https://github.com/pkoperek/keras-dl4j to test it) - a sample model/dataset dumped into test dir and checking if training works (there are no exceptions)? 
* there should be a jar containing all the deps created at the end - not sure how to handle different backend (multiple versions of the jar???) 

All comments welcome!


agibsonccc:
>>here should be a jar containing all the deps created at the end - not sure how to handle different backend (multiple versions of the jar???) 

You can use maven-assembly for this

>> not sure what should be a good automated test here (i just used a simple script from https://github.com/pkoperek/keras-dl4j to test it) - a sample model/dataset dumped into test dir and checking if training works (there are no exceptions)?

Training without exception is acceptable.

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+
+  <groupId>org.deeplearning4j</groupId>
+  <artifactId>deeplearning4j-keras</artifactId>
+  <version>0.7.2-SNAPSHOT</version>
+  <packaging>jar</packaging>
+
+  <name>DL4J Keras support</name>
+  <description>DL4J as a backend for Keras</description>
+  <url>http://maven.apache.org</url>
+
```
agibsonccc: Strip the url please.

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+  <parent>
+    <artifactId>deeplearning4j-parent</artifactId>
+    <groupId>org.deeplearning4j</groupId>
+    <version>0.7.2-SNAPSHOT</version>
+  </parent>
+
+  <groupId>org.deeplearning4j</groupId>
+  <artifactId>deeplearning4j-keras</artifactId>
+  <version>0.7.2-SNAPSHOT</version>
+  <packaging>jar</packaging>
```
agibsonccc: Don't specify the version or groupId it's inherited from the parent already.

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.deeplearning4j</groupId>
+      <artifactId>deeplearning4j-modelimport</artifactId>
+      <version>${dl4j.version}</version>
+      <exclusions>
```
agibsonccc: Change this to project.version, there's zero reason to use dl4.version anywhere in the project.

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>org.deeplearning4j</groupId>
+      <artifactId>deeplearning4j-core</artifactId>
+      <version>${dl4j.version}</version>
+      <exclusions>
```
agibsonccc: Change this to project.version, there's zero reason to use dl4.version anywhere in the project.

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+      <groupId>net.sf.py4j</groupId>
+      <artifactId>py4j</artifactId>
+      <version>0.10.4</version>
+    </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
```
agibsonccc: Don't add the scope here, dependency management from the parent poms already declares the scope.

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+      <artifactId>nd4j-api</artifactId>
+      <version>${nd4j.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.nd4j</groupId>
+      <artifactId>nd4j-native</artifactId>
+      <version>${nd4j.version}</version>
+      <classifier>linux-x86_64</classifier>
+    </dependency>
```
agibsonccc: Don't specify the classifier here. Maven auto figures out what to use based on the os. This makes cross platform builds impossible otherwise. Also make sure to specify <scope>test</scope> (this is how we test all dl4j code) No code should EVER depend on a backend.

deeplearning4j-keras/src/main/java/org/deeplearning4j/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,134 @@
...
+    public DeepLearning4jEntryPoint() {
+
+    }
+
+    public void fit(
+            String modelFilePath,
+            String type,
+            String trainFeaturesFile,
+            String trainLabelsFile,
+            int batchSize,
```
agibsonccc: This is a lot. Maybe consider a POJO using lombok to generate the builder would be easier?

deeplearning4j-keras/src/main/java/org/deeplearning4j/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,134 @@
...
+                    ndIndexes[0] = NDArrayIndex.interval(begin, end);
+                    INDArray featuresBatch = features.get(ndIndexes);
+                    INDArray labelsBatch = labels.get(NDArrayIndex.interval(begin, end));
+                    multiLayerNetwork.fit(featuresBatch, labelsBatch);
+
+                    begin += batchSize;
+                }
+            }
+
+            logger.info("Learning model finished");
```
agibsonccc: Use lombok @Slf4j notation instead.

deeplearning4j-keras/src/main/java/org/deeplearning4j/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,134 @@
...
+        FloatPointer fp = new FloatPointer(dataBuffer);
+        dataSet.read(fp, new hdf5.DataType(hdf5.PredType.NATIVE_FLOAT()));
+        fp.get(dataBuffer);
+        return dataBuffer;
+    }
+
+    private long mulArray(long[] shape) {
+        long retVal = 1L;
+        for (int i = 0; i < shape.length; i++) {
+            retVal *= shape[i];
```
agibsonccc: Please use ArrayUtil.prodLong instead.

deeplearning4j-keras/src/main/java/org/deeplearning4j/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,134 @@
...
+            long nbEpoch,
+            String validationXFilePath,
+            String validationYFilePath,
+            String dimOrdering
+
+    ) throws IOException, InterruptedException, InvalidKerasConfigurationException, UnsupportedKerasConfigurationException {
+        try {
+            MultiLayerNetwork multiLayerNetwork;
+            if ("sequential".equals(type)) {
+                multiLayerNetwork = KerasModelImport.importKerasSequentialModelAndWeights(modelFilePath);
```
agibsonccc: Computation graph? /cc @turambar @eraly

pkoperek:
@agibsonccc: Updated with changes from comments - please have a look in a free moment

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+    <dependency>
+      <groupId>net.sf.py4j</groupId>
+      <artifactId>py4j</artifactId>
+      <version>0.10.4</version>
+    </dependency>
```
agibsonccc: Could you put a py4j.version at the root of the dl4j project?

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+      <groupId>org.nd4j</groupId>
+      <artifactId>nd4j-api</artifactId>
+      <version>${nd4j.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.nd4j</groupId>
+      <artifactId>nd4j-native</artifactId>
+      <version>${nd4j.version}</version>
+    </dependency>
```
agibsonccc: You should never have a dependency on a backend.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,112 @@
...
+import org.nd4j.linalg.util.ArrayUtil;
+
+import java.io.IOException;
+
+import static org.bytedeco.javacpp.hdf5.H5F_ACC_RDONLY;
+
+@Slf4j
+public class DeepLearning4jEntryPoint {
+
+    public void fit(EntryPointFitParameters entryPointFitParameters)
```
agibsonccc: javadoc

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,112 @@
...
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.linalg.indexing.INDArrayIndex;
+import org.nd4j.linalg.indexing.NDArrayIndex;
+import org.nd4j.linalg.util.ArrayUtil;
+
+import java.io.IOException;
+
+import static org.bytedeco.javacpp.hdf5.H5F_ACC_RDONLY;
+
```
agibsonccc: javadoc

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,112 @@
...
+            INDArray features = h5FileToNDArray(entryPointFitParameters.getTrainFeaturesFile());
+            INDArray labels = h5FileToNDArray(entryPointFitParameters.getTrainLabelsFile());
+
+            INDArrayIndex[] ndIndexes = new INDArrayIndex[features.shape().length];
+            for (int i = 0; i < ndIndexes.length; i++) {
+                ndIndexes[i] = NDArrayIndex.all();
+            }
+
+            for (int i = 0; i < entryPointFitParameters.getNbEpoch(); i++) {
+                log.info("Fitting: " + i);
```
agibsonccc: This is a lot of code. Could we maybe have a unit test for *just* this section? I would encapsulate this in a method. The views here are my main concern.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,112 @@
...
+            log.info("Learning model finished");
+        } catch (Throwable e) {
+            log.error("Error while handling request!", e);
+            throw e;
+        }
+    }
+
+    private INDArray h5FileToNDArray(String inputFilePath) {
+        hdf5.H5File h5File = new hdf5.H5File();
+        h5File.openFile(inputFilePath, H5F_ACC_RDONLY);
```
agibsonccc: try/with would be cleaner

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,112 @@
...
+        hdf5.DataSet dataSet = h5File.asCommonFG().openDataSet("data");
+        int[] shape = extractShape(dataSet);
+        long totalSize = ArrayUtil.prodLong(shape);
+        float[] dataBuffer = readFromDataSet(dataSet, (int) totalSize);
+
+        INDArray input = Nd4j.create(shape);
+        new RecursiveCopier(input, dataBuffer, shape).copy();
+
+        h5File.close();
+
```
agibsonccc: Again try /with

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,112 @@
...
+        int nbDims = space.getSimpleExtentNdims();
+        long[] shape = new long[nbDims];
+        space.getSimpleExtentDims(shape);
+        return toInt(shape);
+    }
+
+    private int[] toInt(long[] array) {
+        int[] retVal = new int[array.length];
+
+        for (int i = 0; i < array.length; i++) {
```
agibsonccc: Add this in to ArrayUtil if it's not already. (This is in nd4j where I put this kind of stuff.)

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/EntryPointFitParameters.java
```diff
@@ -0,0 +1,52 @@
+package org.deeplearning4j.keras;
+
+import lombok.Builder;
+
+@Builder
+public class EntryPointFitParameters {
```
agibsonccc: Javadoc

agibsonccc: Also use lombok rather than manual getters and setters.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/Server.java
```diff
@@ -0,0 +1,16 @@
+package org.deeplearning4j.keras;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import py4j.GatewayServer;
+
+public class Server {
```
agibsonccc: Javadoc.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/Server.java
```diff
@@ -0,0 +1,16 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(Server.class);
+
+    public static void main(String[] args) {
```
agibsonccc: Use @Slf4j in lombok.

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
```diff
@@ -0,0 +1,49 @@
...
+    public @Rule ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void shouldFitTheSampleSequentialModel() throws Exception {
+        // Given
+        File model = prepareResource("theano_mnist/model.h5");
+        File features = prepareResource("theano_mnist/x_train.h5");
+        File labels = prepareResource("theano_mnist/y_train.h5");
+
+        EntryPointFitParameters entryPointParameters = EntryPointFitParameters
```
agibsonccc: Comment on overall api. Is there any way you can just use a neuralnet configuration directly or better yet leverage the work @turambar and @eraly are doing with model import?

deeplearning4j-keras/pom.xml
pkoperek: :+1:

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/Server.java
pkoperek: :+1:

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
pkoperek: :+1:  

https://github.com/deeplearning4j/nd4j/pull/1554

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/Server.java
pkoperek: :+1:

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
pkoperek: :+1:

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
pkoperek: :+1:

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
```diff
@@ -0,0 +1,49 @@
...
+                .builder()
+                .modelFilePath(model.getAbsolutePath())
+                .trainFeaturesFile(features.getAbsolutePath())
+                .trainLabelsFile(labels.getAbsolutePath())
+                .batchSize(128)
+                .nbEpoch(2)
+                .type("sequential")
+                .build();
```
agibsonccc: Make "sequential" an enum

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
agibsonccc: Nvm I see how you're doing it now.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/EntryPointFitParameters.java
pkoperek: :+1:

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
pkoperek: :+1:

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
pkoperek: :+1:

pkoperek:
Two general questions about the backends:
1) You said that maven automatically selects the one with correct classifier based on the operating system: I was going through poms/plugins and couldn't figure out how this is done. Could you point me to some docs?
2) I moved the nd4j-native to test scope (without this tests fail as i don't have a backend on classpath). However, if i want to build a package which is shippable with python wrapper code and/or can be run separately - i think i will have to add the backend as a dependency (have a separate proflle for this? not sure if this would make much sense). 

@agibsonccc: What do you think about wrapping the server with a docker container? Using dl4j as a keras backend would require then:
1) install the wrapper package with pip
2) pull & start the docker container 
3) use the wrapper in keras script

I think there have to be some manual server installation steps - docker seems to offer least friction.

saudet:
The dependency is here: https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml#L18 but that doesn't work for Gradle, sbt, etc so don't rely on that, and use nd4j-native-platform as a separate dependency somewhere yes.

deeplearning4j-keras/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.nd4j</groupId>
+      <artifactId>nd4j-native-platform</artifactId>
+      <version>${nd4j.version}</version>
+      <scope>test</scope>
```
agibsonccc: I'm not sure if you tested this :D. It should be physically impossible for this to work. You can only use nd4j-native when building from source. This includes test. Please change it back.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -0,0 +1,45 @@
...
+     * Performs fitting of the model which is referenced in the parameters according to learning parameters specified.
+     *
+     * @param entryPointFitParameters Definition of the model and learning process
+     */
+    public void fit(EntryPointFitParameters entryPointFitParameters) throws Exception {
+
+        try {
+            MultiLayerNetwork multiLayerNetwork = neuralNetworkReader.readNeuralNetwork(entryPointFitParameters);
+
+            INDArray features = ndArrayHDF5Reader.readFromPath(entryPointFitParameters.getTrainFeaturesFile());
```
agibsonccc: Is this how keras data is stored? Bit concerned about data representation among other things here..What about reading in numpy arrays?

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/NDArrayHDF5Reader.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.bytedeco.javacpp.FloatPointer;
+import org.bytedeco.javacpp.hdf5;
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.linalg.util.ArrayUtil;
+
+import static org.bytedeco.javacpp.hdf5.H5F_ACC_RDONLY;
+
+public class NDArrayHDF5Reader {
```
agibsonccc: Javadoc.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/NeuralNetworkBatchLearner.java
```diff
@@ -0,0 +1,41 @@
+package org.deeplearning4j.keras;
+
+import lombok.extern.slf4j.Slf4j;
+import org.deeplearning4j.nn.multilayer.MultiLayerNetwork;
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.indexing.INDArrayIndex;
+import org.nd4j.linalg.indexing.NDArrayIndex;
+
+@Slf4j
+public class NeuralNetworkBatchLearner {
```
agibsonccc: Javadoc

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/NeuralNetworkBatchLearner.java
```diff
@@ -0,0 +1,41 @@
...
+        }
+        return ndIndexes;
+    }
+
+    public void fitInBatches(MultiLayerNetwork multiLayerNetwork, INDArray features, INDArray labels, int batchSize) {
+        final INDArrayIndex[] ndIndexes = createSlicingIndexes(features.shape().length);
+
+        int begin = 0;
+        int samplesCount = features.size(0);
+
```
agibsonccc: This feels wrong to me. Does it have to be an in memory object? We should be consistent and try to to use iterators if possible.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/NeuralNetworkReader.java
```diff
@@ -0,0 +1,27 @@
...
+import org.deeplearning4j.nn.modelimport.keras.KerasModelImport;
+import org.deeplearning4j.nn.modelimport.keras.UnsupportedKerasConfigurationException;
+import org.deeplearning4j.nn.multilayer.MultiLayerNetwork;
+
+import java.io.IOException;
+
+@Slf4j
+public class NeuralNetworkReader {
+
+    public MultiLayerNetwork readNeuralNetwork(EntryPointFitParameters entryPointFitParameters)
```
agibsonccc: Javadoc

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/RecursiveCopier.java
```diff
@@ -0,0 +1,33 @@
+package org.deeplearning4j.keras;
+
+import org.nd4j.linalg.api.ndarray.INDArray;
+
+public class RecursiveCopier {
```
agibsonccc: Javadoc

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/RecursiveCopier.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private final INDArray output;
+    private final float[] input;
+    private final int[] shape;
```
agibsonccc: Use databuffers here. (Nd4j.createBuffer()) the input should never assumed to be float.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/Server.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import py4j.GatewayServer;
+
+/**
+ * Main class for the DL4J-as-Keras-backend. Simply launches py4j GatewayServer with
+ * an entry point exposing API available on Python side.
+ */
+@Slf4j
```
agibsonccc: More information. Also add @author

deeplearning4j-keras/pom.xml
pkoperek: Maybe I did something weird with maven, but in IntelliJ runs the unit tests without a problem :) 
Doesn't nd4j-native-platform refer to nd4j-native? 

Excerpt from *-platform pom:
```
    <properties>
        <nd4j.backend>nd4j-native</nd4j.backend>
    </properties>

    <dependencies>
        <dependency>
            <groupId>${project.groupId}</groupId>
            <artifactId>${nd4j.backend}</artifactId>
            <version>${project.version}</version>
        </dependency>
```

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
pkoperek: No - on Python side I'm dumping the dataset myself (and I chose HDF5 - this is the same format Keras uses to export models).

I know that there are problems with this e.g. de/serialization slows down the process, there are obvious practical size limitations. The reason I did it this way is that on the Python side the data coming from the raw data set can be pre-processed (and I think it usually is :) ) before learning and we would have to somehow copy that logic to Java in order to be able to perform the fitting. 

The problem here is that we are not using Jython - because it cannot support numpy, and simply just export/import the Keras model. This requires exporting/importing the pre-processed data which was supposed to be fed into the model. 

For large distributed datasets it won't work completely - however I'm not sure if in this case someone wouldn't just invest time in rewriting his model in DL4J directly.

Regarding reading/writing numpy arrays: the format used here is pickle (https://docs.python.org/2/library/pickle.html). I was looking for some decent library to read this inside JVM and couldn't really find anything what would convince me that it is stable and well tested. On the other hand HDF5 is already used by Keras and seems to work pretty well.

I don't really have a strong opinion which lib/data format should be used - if you feel there is a better way to do it let me know.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/NeuralNetworkBatchLearner.java
pkoperek: I think that the whole HDF5 file is read into memory when I read it from disk to an NDArray
here:
```
INDArray input = Nd4j.create(shape);
new RecursiveCopier(input, dataBuffer, shape).copy();
```

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/Server.java
pkoperek: I always kind of thought that adding author just makes the file longer as you can check the real author just by looking on blame of the file ;) but ok - I'll add this :)

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -25,14 +23,15 @@ public void fit(EntryPointFitParameters entryPointFitParameters) throws Exceptio
         try {
             MultiLayerNetwork multiLayerNetwork = neuralNetworkReader.readNeuralNetwork(entryPointFitParameters);
 
-            INDArray features = ndArrayHDF5Reader.readFromPath(entryPointFitParameters.getTrainFeaturesFile());
-            INDArray labels = ndArrayHDF5Reader.readFromPath(entryPointFitParameters.getTrainLabelsFile());
-            int batchSize = entryPointFitParameters.getBatchSize();
+            DataSetIterator dataSetIterator = new HDF5MiniBatchDataSetIterator(
+                    entryPointFitParameters.getTrainFeaturesDirectory(),
+                    entryPointFitParameters.getTrainLabelsDirectory()
+            );
```
agibsonccc: Not quite what I expected but I can deal with this.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
```diff
@@ -25,14 +23,15 @@ public void fit(EntryPointFitParameters entryPointFitParameters) throws Exceptio
         try {
             MultiLayerNetwork multiLayerNetwork = neuralNetworkReader.readNeuralNetwork(entryPointFitParameters);
```
agibsonccc: What about ComputationGraph and sequential? This should only be a few lines of code difference anyways.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/HDF5MiniBatchDataSetIterator.java
```diff
@@ -0,0 +1,146 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean asyncSupported() {
+        /**
+         * Async support is turned off on purpose: otherwise there are indeterministic segfaults in JavaCPP
+         * when cleaning memory after HDF5 libs.
+         */
+        return false;
```
agibsonccc: @saudet anything we could do about this?

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/HDF5MiniBatchDataSetIterator.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public int numExamples() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void setPreProcessor(DataSetPreProcessor preProcessor) {
+        throw new UnsupportedOperationException();
```
agibsonccc: This doesn't strike me as reasonable. Just add a field for it and call set :D

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/HDF5MiniBatchDataSetIterator.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public DataSetPreProcessor getPreProcessor() {
+        throw new UnsupportedOperationException();
+    }
```
agibsonccc: Same it's 1 field.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/HDF5MiniBatchDataSetIterator.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public List<String> getLabels() {
+        return null;
```
agibsonccc: Throw an unsupported operation exception here.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/NDArrayHDF5Reader.java
```diff
@@ -2,34 +2,43 @@
...
+ * Reads and INDArray from a HDF5 dataset. The array is expected to be included as the "data" dataset inside the file.
+ * The shape of the output array is the same as the one stored in the HDF5 file.
+ *
+ * @author pkoperek@gmail.com
+ */
 public class NDArrayHDF5Reader {
 
-    public INDArray readFromPath(String inputFilePath) {
+    public INDArray readFromPath(Path inputFilePath) {
         try (hdf5.H5File h5File = new hdf5.H5File()) {
```
agibsonccc: javadoc

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/RecursiveCopier.java
```diff
@@ -21,7 +27,7 @@ public void copy() {
 
     private void copyRecursive(int shapeIdx, int[] indexes) {
         if (shape.length == shapeIdx) {
-            output.putScalar(indexes, input[inputIdx++]);
+            output.putScalar(indexes, input.getFloat(inputIdx++));
```
agibsonccc: Wondering if we could use memcpy or something better here?

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
```diff
@@ -1,31 +1,45 @@
...
+        final Path labels = prepareLabels("theano_mnist");
 
         EntryPointFitParameters entryPointParameters = EntryPointFitParameters
                 .builder()
                 .modelFilePath(model.getAbsolutePath())
-                .trainFeaturesFile(features.getAbsolutePath())
-                .trainLabelsFile(labels.getAbsolutePath())
+                .trainFeaturesDirectory(features.toAbsolutePath().toString())
+                .trainLabelsDirectory(labels.toAbsolutePath().toString())
                 .batchSize(128)
```
agibsonccc: Better. Why not also allow files though? The internal implementation of this would just be equivalent to calling director on the file's parent.

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
```diff
@@ -38,12 +52,46 @@ public void shouldFitTheSampleSequentialModel() throws Exception {
...
+    private File targetFile(Path tempFeaturesDirectory, String batchFileName) {
+        return Paths.get(
+                tempFeaturesDirectory.toFile().getAbsolutePath().toString(),
+                batchFileName
+        ).toFile();
+    }
+
+    private Set<String> listBatchFiles(String classpathDirectory) {
+        return new Reflections(classpathDirectory, new ResourcesScanner()).getResources(endsWith(".h5"));
+    }
```
agibsonccc: Not sure why you'd use reflections here. This shouldn't have to be on the classpath?

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
```diff
@@ -38,12 +52,46 @@ public void shouldFitTheSampleSequentialModel() throws Exception {
...
         File file = File.createTempFile("dl4j", "test");
         file.delete();
         file.deleteOnExit();
-        FileUtils.copyInputStreamToFile(this.getClass().getClassLoader().getResourceAsStream(resourceName), file);
+        copyClasspathResourceToFile(resourceName, file);
         return file;
     }
 
+    private void copyClasspathResourceToFile(String resourceName, File file) throws IOException {
+        FileUtils.copyInputStreamToFile(this.getClass().getClassLoader().getResourceAsStream(resourceName), file);
```
agibsonccc: Again, I don't think data needs to be on the classpath.

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
agibsonccc: Ahh, testing. That makes sense. That's still going to be slow though :D

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
agibsonccc: Ahh, testing. That makes sense. That's still going to be slow though :D

deeplearning4j-keras/src/test/java/org/deeplearning4j/keras/DeepLearning4jEntryPointTest.java
pkoperek: I think keeping only directories makes the server part more clear (simpler). The client side has to generate the batch files by using functions defined in the Python lib so either way we can 'enforce' the correct parameters.

agibsonccc: Fine with me :D.

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/RecursiveCopier.java
pkoperek: not sure if i'm not doing something colossally stupid, but maybe instead of using the *Copier I could just do: 
```
            INDArray input = Nd4j.create(shape);
            input.setData(dataBuffer);
```

deeplearning4j-keras/src/main/java/org/deeplearning4j/keras/DeepLearning4jEntryPoint.java
pkoperek: This is the Keras sequential model, no? :) 
Lets keep this PR simple - I'll add the ComputationGraph as a separate one (want to unblock myself to move on with the Python stuff)

PR #2566, opened 2016-12-23
<WIP> keras model import - computational graph

eraly:


deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/KerasModel.java
```diff
@@ -539,18 +548,41 @@ protected static Object checkAndGetTrainingField(Map<String,Object> map, String
             throws InvalidKerasConfigurationException {
         /* TODO: how might this break?
          * - mismatch between layer/parameter names?
```
eraly: Eventually rewrite this to go from layer to weights instead of the other way. For now, put off till after testing.

turambar:
@eraly I pushed a change to `KerasModelImport.readWeightsFromHdf5` that fixes the problem you ran into with user-provided layer names. Old version used an _ad hoc_ string splitting approach that was hardwired to Keras-generated parameter names (which always have format `[layer type]_[number]_[parameter name]`). We now use a RegEx that is substantially more robust. Works for both Keras-generated layer names and the layer names you were using. Needs more testing, but should be good to go. Once verified, we can remove the fix that you added to `KerasModel.copyWeightsToModel` (see commented TODO that I added).

turambar:
Fixed inference of DimOrder for input layers.

eraly:
180 degree fix for theano filters in.

PR #2634, opened 2017-01-06
Add method for shuffling DataSet objects in a RDD

AlexDBlack:


EdeMeijer:
Ohh shiny, even better than my current solution of just shuffling entire data sets around.

Issue #2616, opened 2017-01-04
Per-example weights


EdeMeijer:
I was discussing the option of a per-example weights feature on gitter with @AlexDBlack, but decided to continue it in an issue.

So, I would like to be able to use per-example weights in order to boost the importance of certain examples. One option is to use oversampling, but that has some drawbacks:

- Only effectively allowing integer weights
- Makes training slower (more artificial examples), especially if you need big weights

Just for fun I tried to implement example weights and make it work with comp graphs, which wasn't too hard. POC can be found [here (nd4j)](https://github.com/EdeMeijer/nd4j/commit/de2de89b4f1fa57d0e9e191cf851d37c174b9c33) and [here (dl4j)](https://github.com/EdeMeijer/deeplearning4j/commit/04f8c732ef106fa028fa54e67bde498168b160a1). However, after a little discussion Alex figured it might be easier to implement this feature by extending the masking infrastructure, and introduce an additional feature while we're at it: per-output-timestep masking and weighing. This could be done by adding a dimension to the output mask in case the user desires to specify weights for individual time steps.

So, what do you guys think?

AlexDBlack:
Right, my current thinking on the subject is basically that we can implement per-example (and: per time step) weights by extending the current masking infrastructure.
Currently masks are either 0 or 1 - but in principle could be any arbitrary weighting value.

Per *output* (not per example) weights could be implemented (along with per-output) masks as follows:
if mask rank == labels rank: it's per output masking
if mask rank == labels rank - 1: it's per example/time step masking (like what we have now: 2d mask array for per-time step masking of a 3d array)

Per output weighting is useful for some cases like missing values, some RL cases, and some architectures like this: http://arxiv.org/abs/1603.00806

One question of course is how other DL platforms handle this - it might be good to review that too.

fac2003:
We also need per example weights. It is useful when you have vastly imbalanced datasets (1 in 1 positive for every 1000 negative for instance). I am looking forward to the feature.

On the other hand, I personally prefer not to mix roles of methods and data attributes. Mixing masking and weighting is likely to be confusing. From a user point of view, I would prefer distinct methods (one for output weights, one for masks), with clear documentation each, even if you store the data in the INDArray in the mask fields internally. 

You could make it so that clients can only call one of the methods. When reading the code, it will be clear what the intent is from the name of the method without having to think about the rank of the parameter.

cacophany53:
[...]

PR #2659, opened 2017-01-09
Support per-example and per-timestep weights

EdeMeijer:
As discussed in https://github.com/deeplearning4j/deeplearning4j/issues/2616, this diff adds support for using arbitrary weights in output masks to enable per-example and per-timestep weights. It really only required a one liner change and implementing it this way guarantees backwards compatibility.

The tests are written using computation graph, but should work with multi layer networks as well since the actual change was in the base layer.


AlexDBlack:
So it's a good start - but I'm pretty sure there's a few more things we need here.
For example, the evaluation classes assume 0/1 mask values; I'd also want some gradient checks with this.

Perhaps I'll start from this and review those issues - and see if there are any other places that might not work as expected using this approach.

EdeMeijer:
Please do. I think we don't need full support for every conceivable scenario in the first version, but add support as they come. When documenting this feature, we could mention which cases are supported and which aren't. For example 4D probably won't work yet.

I guess evaluation is pretty important to get right immediately. Gradient checks would be nice too. I know about the concept, but haven't implemented my own so far, so I'd rather leave that to you.

AlexDBlack:
Sure, that's fine. I've worked entensively on both, so it shouldn't take me long.
Beyond that, I mainly just want to know where the issues are (and what we need to fix) before declaring to the world that we have this as a feature. :)

AlexDBlack:
@EdeMeijer fyi I do still intent to review and add to this, just haven't had the bandwidth to do it yet. I realize it's been a while.

EdeMeijer:
That's okay, I understand. I'm busy with some non deep learning stuff atm anyway.

mbkennel:
My request is for per-sample weights to work with the ROC performance measurement.

agibsonccc:
Where are we with this?

agibsonccc:
This was a lot more work than we expected to get merged. Closing this for now due to inactivity.

PR #2683, opened 2017-01-15
Add CenterLossOutputLayer for efficient training

crockpotveggies:
_Work in progress..._

Center loss has proven to be more efficient than triplet loss, and it enables classifier training which is also more speedy than triplets.

@AlexDBlack can you take a look at CenterLossParamInitializer and confirm it's on the right track? Also, should we just specify `numClasses` in layer conf? Let's keep discussion in Gitter :)


AlexDBlack:
Param initializer looks to be on the right path.

Some things I noticed
- Map order for init and GradientFromFlattened must match
- ``` conf.getLayer()).getKernelSize().length != 2``` CenterLossOutputLayer shouldn't have a kernel size config option
- centerLossView needs to be reshaped in both init and getGradientsFromFlattened (I'd suggest C order, shape [numClasses, nIn])
- backpropGradient is wrong currently

agibsonccc:
@crockpotveggies update here?

crockpotveggies:
Will tackle this on the weekend. Currently focused on Keras.

crockpotveggies:
Waiting on backprop math from @AlexDBlack on this and then will wrap it up

AlexDBlack:
Will be merged via this PR instead: https://github.com/deeplearning4j/deeplearning4j/pull/2808

PR #2709, opened 2017-01-19
Fix constructor in GraphNodes test.

crockpotveggies:
Self-explanatory. Cleanup from recent merges.


AlexDBlack:
Same fix applied here: https://github.com/deeplearning4j/deeplearning4j/pull/2711/files

PR #2781, opened 2017-01-31
Keras model import update: embedding layer, end-to-end unit tests

turambar:
* Embedding layer
* end-to-end unit tests


deeplearning4j-modelimport/src/test/java/org/deeplearning4j/nn/modelimport/keras/KerasModelEndToEndTest.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+        return predictions;
+    }
+
+    static public void compareINDArrays(String label, INDArray a, INDArray b, double eps) {
+        INDArray diff = a.sub(b);
+        double min = diff.minNumber().doubleValue();
+        double max = diff.maxNumber().doubleValue();
+        boolean eq = a.equalsWithEps(b, eps);
+//        log.info(label + ": " + a.equalsWithEps(b, eps) + ", " + min + ", " + max);
```
huitseeker: You may want to `log.debug` and leave this uncommented, no ?

deeplearning4j-modelimport/src/test/java/org/deeplearning4j/nn/modelimport/keras/KerasModelEndToEndTest.java
```diff
@@ -0,0 +1,156 @@
...
+import static org.junit.Assert.*;
+
+/**
+ * Unit tests for end-to-end Keras model import.
+ *
+ * TODO: find a more elegant fix for the per-layer activation comparisons
+ * since some layers (e.g., ActivationLayer) overwrite previous
+ * layer's activations when run in train=false mode.
+ *
+ * TODO: ndarray.equalsWithEps(ndarray) appears to be broken
```
huitseeker: In which way ? Is there an Nd4j issue ?

deeplearning4j-modelimport/src/test/java/org/deeplearning4j/nn/modelimport/keras/KerasModelEndToEndTest.java
turambar: yeah probably, these unit tests are WIP. ;)

deeplearning4j-modelimport/src/test/java/org/deeplearning4j/nn/modelimport/keras/KerasModelEndToEndTest.java
turambar: Not sure whether it's user error or a bug in ND4J. I just know that the stored Keras and computed DL4J activations are similar within 6-7 digits of precision, but `equalsWithEps` returns false. Investigating.

Issue #2819, opened 2017-02-03
NGramTokenizer tokenizes the string representation of a list


mjhugo:
I'm trying to use an NGramTokenizer on the DL4J Word2VecRawTextExample.  I took the example as is, modified the tokenizer factory from `DefaultTokenizerFactory` to `NGramTokenizerFactory` like this:

```java
TokenizerFactory t = new NGramTokenizerFactory(new DefaultTokenizerFactory(),1, 3);
```

and I changed the value sent into `vec.wordsNearest` from `day` to `[day]`

```java
log.info("Closest Words:");
Collection<String> lst = vec.wordsNearest("[day]", 10);
System.out.println("10 Words closest to '[day]': " + lst);
```

but I'm not getting anything out of that:

```
10 Words closest to '[day]': []
```

Turns out the NGramTokenizer is adding brackets (a `.toString()` representation of a list) to the tokens.

Using the following:

```java
Collection<String> lst = vec.wordsNearest(" [day]", 10);
```
results in the output:

```
10 Words closest to 'day': [ [day, ],  [night],  [was, every],  [every],  [another, day],  [every, day],  [today, s],  [but, today],  [day, after],  [one, day]]
```

@raver119 [suggested opening an issue](https://gitter.im/deeplearning4j/deeplearning4j?at=58949d854c04e9a44e676a85)

> that should be:
> a) configurable brackets
> b) no spaces if brackets are used

In looking at [the code](https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/text/tokenization/tokenizer/NGramTokenizer.java#L50), I think the intent was to separate the tokens with a space, e.g. `[another, day]` should be `another day`

raver119:
fix is merged

PR #2820, opened 2017-02-03
create ngram tokens as space delimited words

mjhugo:
This change makes the NGramTokenizer behave as I believe it was originally intended - with words delimited by spaces.  

This pull request does not address the other items in the issue, namely:

> a) configurable brackets
> b) no spaces if brackets are used

#2819

### DIFF
--- a/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/text/tokenization/tokenizer/NGramTokenizer.java
+++ b/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/text/tokenization/tokenizer/NGramTokenizer.java
@@ -47,7 +47,7 @@ public class NGramTokenizer implements Tokenizer {
             for (int i = minN; i < min; i++) {
                 for (int j = 0; j < nOriginalTokens - i + 1; j++) {
                     List<String> originalTokensSlice = this.originalTokens.subList(j, j+ i);
-                    this.tokens.add(StringUtils.join(" ", originalTokensSlice));
+                    this.tokens.add(StringUtils.join(originalTokensSlice, " "));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/text/tokenization/tokenizer/NGramTokenizer.java
```diff
@@ -47,7 +47,7 @@ public NGramTokenizer(Tokenizer tokenizer,Integer minN, Integer maxN) {
             for (int i = minN; i < min; i++) {
                 for (int j = 0; j < nOriginalTokens - i + 1; j++) {
                     List<String> originalTokensSlice = this.originalTokens.subList(j, j+ i);
-                    this.tokens.add(StringUtils.join(" ", originalTokensSlice));
+                    this.tokens.add(StringUtils.join(originalTokensSlice, " "));
```
raver119: I guess we should just get rid of  " ", instead of putting it to the end.

mjhugo: I think the method params just got mixed up

The method signature for .join is
```
    public static String join(Iterable<?> iterable, String separator) {
        return iterable == null?null:join(iterable.iterator(), separator);
    }
```

the collection comes first, the character to join with comes second.

given a collection 

```java
originalTokensSlice = ["foo", "bar", "baz"]
```

`StringUtils.join(" ", originalTokensSlice)` will return `" [foo, bar, baz]"`

where as 

`StringUtils.join(originalTokensSlice, " " )` will return `"foo bar baz"`

which is i think what we want, e.g. https://github.com/deeplearning4j/deeplearning4j/pull/2820/files#diff-9bca4ca3024bc576e4171421119da2ebR56

raver119: Ah, ok

PR #2823, opened 2017-02-05
AbstractDataSetIterator performance tweak

akhodakivskiy:
The change is too call `Nd4j.create()` once instead of multiple `Nd4j.create()` followed by `Nd4j.vstack()`. This tweak gives me up to 10x performance boost - depending on the batch size and the number of features/labels. It's an important improvement for in-memory datasets because these iterators are called during training all the time.


raver119:
Looks great, thanks!

PR #2836, opened 2017-02-09
Fix small issue with #2835

huitseeker:


huitseeker:
And jenkins is barfing on this one.

PR #2785, opened 2017-01-31
Unit test fixes

AlexDBlack:

PR #2837, opened 2017-02-09
Repalce #2785 with recent code changes

huitseeker:
Submitted to help out @AlexDBlack


huitseeker:
Actually, wait, I can fix this another way.

PR #2849, opened 2017-02-12
<WIP> Keras import support for global pooling

turambar:
Keras Model Import support for global pooling layers. This required some changes to `KerasLayer` and subclasses, `KerasModel` and `KerasSequentialModel` to add ability to infer preprocessors when adding layers. Verify unit tests pass before merging.

FYI: I discovered what I believe to be a general (non-model import specific) DL4J problem: `EmbeddingLayer` spits out an FF InputType but `Global*Pooling` layers throw Exceptions when they receive FF inputs. This means that DL4J won't support Fasttext unless the user manually sets a `FeedForwardToRnnInputPreProcessor` for the global pooling layer. This, I believe, is generally inconsistent with the DL4J ethos of handling input preprocessors under the hood (so most users don't have to think about input preprocessors most of the time).

Two possible fixes:

* `Global*Pooling` layer accepts FF input types and should do The Right Thing based on its own internal dimensionality. This would be a pretty easy fix, but would require the user to pass in the pooling dimensions (which is currently optional). I think this isn't unreasonable; Keras requires a choice about pooling dimension.

* `EmbeddingLayer` spits out an RNN InputType. This would require an internal change to embedding layers that would be inconsistent with how most DL4J layers work in general (i.e., they spit out the same InputType that they receive).

I can pursue either of the proposed fixes, but I'd like to hear feedback from the peanut gallery before I choose which one. @agibsonccc @AlexDBlack @eraly @crockpotveggies @raver119


AlexDBlack:
@turambar: EmbeddingLayer outputting an RNN InputType isn't viable here.
GlobalPooling layer directly using FF input types: not really viable here either. FF input type (in the case of RNNs + dense or RNN + embedding) means "time distributed"... global pooling layer shouldn't be concerned with details like that.

Thinking about it: 2 things I'd do here.
1. GlobalPoolingLayer.getPreProcessorForInputType defaults to returning a FeedForwardToRnnPreprocessor in case of FF input type. That seems reasonable for me (doesn't work for flattened cnn data, but we have convolutionalFlat for that anyway)
2. Use the pooling dimension (if set by the user) to infer the preprocessor to return in getPreprocessorForInputtype (dim 2: ff->rnn, dim3+4: ff->cnn)

PR #2932, opened 2017-02-27
Build script upgrades

huitseeker:
- adds optional testing
- optional shallow cloning
- makes directory deletion optional, deletes old artifacts in the maven cache instead
- optional detection of backend from the environment

### DIFF
--- a/build-dl4j-stack.sh
+++ b/build-dl4j-stack.sh
@@ -45,6 +45,9 @@ case $key in
     SCALAV="$2"
     shift # past argument
     ;;
+    -s|--shallow)
+    SHALLOW="YES"
+    ;;
     *)
             # unknown option
     ;;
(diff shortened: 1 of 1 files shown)


build-dl4j-stack.sh
```diff
@@ -70,26 +90,58 @@ if [ "$SCALAV" == "2.11" ]; then
...
+###########################
+# Script execution starts #
+###########################
+
+pushd ..
+
+# removes lingering snapshot artifacts from existing maven cache to ensure a
+# clean build
+JAVA_PROJECTS="nd4j datavec deeplearning4j"
+for dirName in $JAVA_PROJECTS; do
```
agibsonccc: Something that maybe out of scope here but worth acknowledging: Arbiter,rl4j, and other libs are needed as part of this as well.

crockpotveggies:
Is it worth adding extra repositories as a command line flag?

huitseeker:
I agree that other repos are a nice to have, but right now most of our users will be happy with this.

PR #2957, opened 2017-03-02
Add/restore normalizers to/from model files using new normalizer serializer

EdeMeijer:
Instead of using Java serialization, this diff uses the new NormalizerSerializer that serializes to a custom binary format. In order to support model files with normalizers added with the old functionality, the deprecated unserialization function is kept and used as a fallback approach. If an old format is detected, it will automatically be fixed by re-saving the normalizer.

---

Note: I manually tested the fallback deserialization and auto fixing. These are the steps I took to test it:
- Put back the old serialization function with a different name
- In the unit test `ModelSerializerTest.testWriteMlnModelInputStream` I used this function instead
- In the test, I restored the normalizer **twice**, and checked that I only saw the warning about automatically fixing the file **once**, which means that the second time the file was indeed updated.


EdeMeijer:
Oh yeah, I couldn't resist doing a little refactoring on the IO stuff because I noticed a lot of things could be simplified by using apache commons IO, which I added as a dependency.

PR #2976, opened 2017-03-04
1D convolutional and subsampling layers

turambar:
Adding 1D convolutional and subsampling layers. Subclassed off of respective existing (2D) ConvolutionLayer and SubsamplingLayer classes. Had to make some members and constructors in the super classes protected instead of private.


turambar:
OK, these appear to be working. I'm going to push an example to the examples repository where they do a nice job of solving a time series forecasting problem. I haven't added formal unit tests or gradient checks, but since we're currently just reusing the existing ConvolutionLayer and SubsamplingLayer classes with some input/output reshaping, they probably aren't necessary for the time being. I want to make sure they're included in the release, so I'm going to go ahead and merge.

turambar:
@eraly @AlexDBlack @agibsonccc You guys want to take a quick look at these and make sure you can't find anything objectionable?

turambar:
Here's a link to a PR with a example temporal convolutional network: https://github.com/deeplearning4j/dl4j-examples/pull/382

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution1DLayer.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        public Builder(int kernelSize) {
+            this();
+            this.kernelSize[0] = kernelSize;
+        }
+
+        public Builder() {
+            this.kernelSize = new int[] { 5, 1 };
+            this.stride = new int[] { 1, 1 };
```
agibsonccc: Where does the 5 come from? I feel like we should just throw an exception here.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/ConvolutionLayer.java
```diff
@@ -162,11 +162,11 @@ public double getLearningRateByParam(String paramName) {
...
     @AllArgsConstructor
     public static class Builder extends FeedForwardLayer.Builder<Builder> {
-        private ConvolutionMode convolutionMode = null;
-        private int[] kernelSize = new int[] {5,5};
-        private int[] stride = new int[] {1,1};
-        private int[] padding = new int[] {0, 0};
-        private AlgoMode cudnnAlgoMode = AlgoMode.PREFER_FASTEST;
+        protected ConvolutionMode convolutionMode = null;
+        protected int[] kernelSize = new int[] {5,5};
+        protected int[] stride = new int[] {1,1};
```
agibsonccc: Again, this kernel size feels arbitrary. I feel like this should just be soething people are required to define.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Subsampling1DLayer.java
```diff
@@ -0,0 +1,206 @@
...
+            this.convolutionMode = convolutionMode;
+            return this;
+        }
+
+        public Builder() {}
+
+        @SuppressWarnings("unchecked")
+        public Subsampling1DLayer build() {
+            if(poolingType==org.deeplearning4j.nn.conf.layers.PoolingType.PNORM && pnorm <= 0) throw new IllegalStateException("Incorrect Subsampling config: p-norm must be set when using PoolingType.PNORM");
+            ConvolutionUtils.validateCnnKernelStridePadding(kernelSize, stride, padding);
```
agibsonccc: Spacing.

turambar:
@agibsonccc I'll add tests and exceptions for not setting kernel size (rather than arbitrary defaults) over next day. I'll make similar changes in original ConvolutionLayer, too.

Meta-question for crowd: are we happy with semantics of a layer called Convolution1DLayer? If so, should we rename ConvolutionLayer to Convolution2DLayer (or add a wrapper by that name, for backwards compatibility)?

Alternatively, we could merge the 1D functionality into ConvolutionLayer. I think it would be straightforward to do the Right Thing, depending on kernel size. This would simplify decisions for users in some ways but would also increase the chances of them unintentionally doing something "wrong" and not realizing it because it still "works."

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution1DLayer.java
turambar: Fixed. Now initialized to `null`. When passed to constructor, an Exception is throw. Added a unit test to validate this.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/ConvolutionLayer.java
turambar: Fixed. Now initialized to `null`. When passed to constructor, an Exception is throw. Added a unit test to validate this.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Subsampling1DLayer.java
turambar: Fixed in `SubsamplingLayer` and `Subsampling1DLayer` classes.

turambar:
@agibsonccc actually, what sorts of tests are you looking for? Do you want tests to make sure that *building* a layer performs as expected (i.e., if you pass in an invalid (or no) `kernelSize`, you get an Exception), or tests to confirm that actual output of the layer has the expected shape, size, type, etc.?

AlexDBlack:
For names: I'm fine with Convolution1DLayer and the like.
As for tests: Gradient checks, at the very least. You can probably adapt some of the existing ones.

turambar:
@AlexDBlack yeah I was being lazy. ;) I'll fix the exceptions today.

Re: gradient checks I've started on those now (looked at the utilities last night so I know how to use them now).

Re: validating size, I think I'll just add some tests to make sure that the weights look right and getOutputType() returns the correct thing for various inputs. I'll add one or more tests for the wrong inputs to make sure the proper exception gets thrown.

PR #2997, opened 2017-03-07
Mnist error improvement

crockpotveggies:
Current download error doesn't give enough information about MNIST download failure. This improves that.


deeplearning4j-core/src/test/java/org/deeplearning4j/base/MnistFetcherTest.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @author Justin Long (crockpotveggies)
+ */
+public class MnistFetcherTest {
+
+    @Test
+    public void testIrisDataFetcher() throws Exception {
+        MnistFetcher mnistFetcher = new MnistFetcher();
+        mnistFetcher.downloadAndUntar();
+
```
agibsonccc: Add an assertion here for  the files being downloaded and extracted successfully?

crockpotveggies: It's now there.

crockpotveggies: It's now there

PR #3001, opened 2017-03-08
Add repository updates as an option in build script

huitseeker:


huitseeker:
The DL4J repo is not deleted, as was the behavior of the original repo. I'm happy to expand these modes — we could have the delete mode force dl4j to pull + `git clean -ffxd`. Delete is essentially the signal the current state of the repo does not matter.

crockpotveggies:
Originally the script didn't live inside the DL4J repo, so I didn't have to consider this. However, since it lives there now I suggest for `deeplearning4j` repo *only* we should map the following:

update mode -> `git pull`
delete mode -> HARD `git fetch`

huitseeker:
Done ! New semantics tested sucessfully with 
`$ ~/build-dl4j-stack.sh -c cuda -cc 61 -r delete  `

huitseeker:
ping @crockpotveggies

PR #3043, opened 2017-03-14
Spark dl4j ml

dmmiller612:
I was able to rip out domain specific code. As you will see below, the weirdness for trainingmaster can be changed, if the class/interface implements serializable. This is just a starting point, and looking to see what you guys want to do with this.

example usage:
```
        List<Row> dubs = Lists.newArrayList(
                RowFactory.create(1000.0, 1000.0, 1.0),
                RowFactory.create(90.0, 90.0, 0.0)
        );

        DataFrame df = sqlContext.createDataFrame(dubs, createStruct());
        Pipeline p = new Pipeline().setStages(new PipelineStage[]{getAssembler(new String[]{"x", "y"}, "features")});
        DataFrame part2 = p.fit(df).transform(df).select("features", "label");

        SparkDl4jNetwork sparkDl4jNetwork = new SparkDl4jNetwork()
                .setFeaturesCol("features")
                .setLabelCol("label")
                .setTrainingMaster(() -> new ParameterAveragingTrainingMaster.Builder(3)
                        .averagingFrequency(2)
                        .workerPrefetchNumBatches(2)
                        .batchSizePerWorker(2)
                        .build())
                .setMultiLayerConfiguration(getNNConfiguration());
        SparkDl4jModel sm = sparkDl4jNetwork.fit(part2);
        MultiLayerNetwork mln = sm.getMultiLayerNetwork();
        Assert.assertNotNull(mln);
        System.out.println(sm.output(Vectors.dense(0.0, 0.0)));
        sm.write().save("somewhere.model");
        SparkDl4jModel spdm = SparkDl4jModel.load("somewhere.model");
        System.out.println(spdm.predict(Vectors.dense(0.0, 0.0)));

```


dmmiller612:
This works well with both Java and scala (just like any other spark predictors). Let me know where you want me to put this, name this, and other related things.

agibsonccc:
@dmmiller612 could you create an actual separate submodule under spark? I see this being a separate module.

dmmiller612:
yep

dmmiller612:
If you want, I actually built a custom transformer for autoencoders that does the same thing as a spark pca transformer. I can add that to this sub-module in another pr.

dmmiller612:
Also, I added score listeners as well.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetwork.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    def setCollectionStats(collectStats: Boolean) : SparkDl4jNetwork = {
+        this._collectStats = collectStats
+        this
+    }
+}
+
+class SparkDl4jModel(override val uid: String, network: SparkDl4jMultiLayer)
+    extends PredictionModel[Vector, SparkDl4jModel] with Serializable with MLWritable {
```
agibsonccc: How would adding a matrix work? Can we support tensors?

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetwork.scala
```diff
@@ -0,0 +1,125 @@
...
+
+    override def copy(extra: ParamMap) : SparkDl4jModel = {
+        copyValues(new SparkDl4jModel(uid, network)).setParent(parent)
+    }
+
+    override def predict(features: Vector) : Double = {
+        val v = output(features)
```
agibsonccc: So again, how would we possibly do matrices here? What about tensors?

dmmiller612: I don't think matrices follow the ml contract. This overrides spark's predictor contract, which is used in their ml library. It processes each item vector individually.

agibsonccc: Thanks. That's why I commented rather than blocked the PR.

agibsonccc:
@huitseeker mind taking a look at this? I"m more or less ok with this aside from computation graphs. This could be a new spark package for us.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetwork.scala
dmmiller612: I don't know much about dl4j's computation graphs. I am looking at adding an autoencoder, word2vec, and paragraph vectors as a transformer (where you can apply the transformations to features through a ml pipeline). Maybe computation graphs could follow that pattern. It is slightly different than spark's predictor apis, and I think it can handle more data types.

dmmiller612:
One final note that I just thought of. This will probably eventually need the spark 1, spark 2 packaging. I can do this as well in a follow up pr, since dataframe is replaced with dataset in spark 2.

agibsonccc:
You might want to take a look at how we abstract spark 1/spark2: https://github.com/deeplearning4j/DataVec/tree/master/datavec-spark

dmmiller612:
Will do.

dmmiller612:
@agibsonccc Dug around the code a bit, and saw you guys have a computation graph spark implementation. I can definitely write a ml wrapper around that. I can add it to my list.

huitseeker:
Thanks for your contribution @dmmiller612 ! I'll take a look today!

dmmiller612:
@huitseeker @agibsonccc Made the change to support both spark 1 and 2. I think this is about as flexible as we can get with the change in dataset and vector from spark 1 and 2, and still be able to abide by the transformation contract.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/spark-1/org/deeplearning4j/spark/ml/impl/SparkDl4jNetwork.scala
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.spark.ml.param.ParamMap
+import org.apache.spark.ml.util._
+import org.apache.spark.mllib.linalg.Vector
+import org.apache.spark.mllib.regression.LabeledPoint
+import org.apache.spark.sql.{DataFrame, Row}
+import org.deeplearning4j.spark.impl.multilayer.SparkDl4jMultiLayer
+import org.deeplearning4j.spark.ml.utils.DatasetFacade
+
+
+final class SparkDl4jNetwork(override val uid: String) extends SparkDl4jNetworkWrapper[Vector, SparkDl4jNetwork, SparkDl4jModel](uid) {
```
huitseeker: For consistency in classes that have to be maintained for 2 Spark versions, elsewhere in the code base we use `MyClassAdapter` for what you name `MyClassWrapper`, and `MyClassAdaptee` for what you name `MyClass`. Would you mind shifting your names to this convention? We find it a useful reminder to keep both implementations in sync.

Also, why the final?

deeplearning4j-scaleout/spark/dl4j-spark-ml/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>    <!-- Version set by deeplearning4j-parent dependency management -->
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+
```
huitseeker: If you had time for setting up a scalastyle plugin stanza, it would be great. You can start with the Spark scalastyle-config, set up [here](https://github.com/apache/spark/blob/master/pom.xml#L2374) — but please configure it with `failOnViolation` set to `false`. Scalafmt is cool as well.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -0,0 +1,117 @@
...
+    protected var _multiLayerConfiguration : MultiLayerConfiguration = _
+    protected var _numLabels : Int = 2
+    protected var _freq : Int = 10
+    protected var _trainingMaster : ParamSerializer = _
+    protected var _listeners : util.Collection[IterationListener] = _
+    protected var _collectStats: Boolean = false
+
+    override def copy(extra: ParamMap) : E = defaultCopy(extra)
+
+    val mapVectorFunc : Row => LabeledPoint
```
huitseeker: Please make this a `def`, this way it can be overridden by a def, a lazy val or an object as well (and you'll avoid thorny initialization issues).

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -0,0 +1,117 @@
...
+
+    def setNumLabels(value: Int) : E = {
+        this._numLabels = value
+        this.asInstanceOf[E]
+    }
+
+    def setMultiLayerConfiguration(multiLayerConfiguration: MultiLayerConfiguration) : E = {
```
huitseeker: Why not simply return `SparkDl4jNetworkWrapper[T, E, M]` and avoid the cast (every time) ?

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -0,0 +1,117 @@
...
+        override protected def saveImpl(path: String): Unit = {
+            ModelSerializer.writeModel(network.getNetwork, path, true)
+        }
+    }
+
+    override def write : MLWriter = new SparkDl4jModelWriter(this)
+}
+
+trait SparkDl4jModelWrap extends MLReadable[SparkDl4jModel] {
+    override def read: MLReader[SparkDl4jModel] = new SparkDl4jReader
```
huitseeker: This could do with a few empty lines to ease reading :)

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
dmmiller612: The abstract class is basically a wrapper to get around the spark 1 and spark 2 issue. I didn't want somebody to use this as the type when instantiating a network. In fact, it may be best to limit the scope of this class to protected.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/spark-1/org/deeplearning4j/spark/ml/impl/SparkDl4jNetwork.scala
dmmiller612: I can change it, but it might be weird, since this is the class that a user would implement. I was trying to go for a similar naming convention on what dl4j has for the mllib network.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -0,0 +1,117 @@
...
+
+
+abstract class SparkDl4jNetworkWrapper[T, E <: SparkDl4jNetworkWrapper[T, E, M], M <: SparkDl4jModelWrapper[T, M]]
+(override val uid: String) extends Predictor[T, E, M]  {
+
+    protected var _multiLayerConfiguration : MultiLayerConfiguration = _
+    protected var _numLabels : Int = 2
+    protected var _freq : Int = 10
+    protected var _trainingMaster : ParamSerializer = _
+    protected var _listeners : util.Collection[IterationListener] = _
```
dmmiller612: I used util.Collection, since the dl4j library wants java's util library. I can use the implicit to make that happen though, if that seems cleaner.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.spark.mllib.linalg.Vector
+import org.apache.spark.mllib.regression.LabeledPoint
+import org.apache.spark.sql.Row
+import org.deeplearning4j.nn.conf.MultiLayerConfiguration
+import org.deeplearning4j.nn.multilayer.MultiLayerNetwork
+import org.deeplearning4j.optimize.api.IterationListener
+import org.deeplearning4j.spark.impl.multilayer.SparkDl4jMultiLayer
+import org.deeplearning4j.spark.ml.utils.{DatasetFacade, ParamSerializer}
+import org.deeplearning4j.util.ModelSerializer
+
```
dmmiller612: my dang ide took it out, because I have been having a build issue on one of my machines. Thanks for pointing this out.

dmmiller612:
Fyi, having issues running the tests (immediate failure in a separate module). This test is actually a carry over from what I have in my personal repo, and it works there.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
dmmiller612: done (marking for myself)

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -0,0 +1,117 @@
...
+    protected var _freq : Int = 10
+    protected var _trainingMaster : ParamSerializer = _
+    protected var _listeners : util.Collection[IterationListener] = _
+    protected var _collectStats: Boolean = false
+
+    override def copy(extra: ParamMap) : E = defaultCopy(extra)
+
+    val mapVectorFunc : Row => LabeledPoint
+
+    def setNumLabels(value: Int) : E = {
```
dmmiller612: Great idea, and that follows what Spark does with some ml algorithms. Done in latest commit.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -0,0 +1,117 @@
...
+        if (_listeners != null) {
+            sparkNet.setListeners(_listeners)
+        }
+        val lps = dataset.select(getFeaturesCol, getLabelCol).rdd
+            .map(mapVectorFunc)
+            .map(item => {
+                val features = item.features
+                val label = item.label
+                if (_numLabels > 1) {
+                    new DataSet(Nd4j.create(features.toArray), FeatureUtil.toOutcomeVector(label.asInstanceOf[Int], _numLabels))
```
dmmiller612: Done

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
dmmiller612: Done

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/utils/SparkDl4jUtil.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.spark.ml.param.{ParamPair, Params}
+import org.apache.spark.ml.util.{MLReadable, MLReader}
+import org.apache.spark.sql.types.{DataType, StructField, StructType}
+import org.deeplearning4j.spark.impl.paramavg.ParameterAveragingTrainingMaster
+import org.json4s.JsonDSL._
+import org.json4s.jackson.JsonMethods.{compact, parse, render}
+import org.json4s.{JObject, JValue, _}
+
+/**
+  * This is imported from Spark with a few minor changes, since the libraries are private. This is a known constraint in spark around params.
```
dmmiller612: Good idea. Done.

dmmiller612:
@huitseeker ha, this will be a fun one. If I use a java 8 lambda it works, if I switch it to be compatible with java 7, it is not serializable. I'll try and see what is going on.

dmmiller612:
@huitseeker Fixed. I think eventually it would be nice if trainingmaster was serializable to get around these issues. I didn't want to tackle that in this pr though.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/test/java/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkTest.java
```diff
@@ -0,0 +1,110 @@
...
+                .build();
+    }
+
+    public static VectorAssembler getAssembler(String[] input, String output){
+        return new VectorAssembler()
+                .setInputCols(input)
+                .setOutputCol(output);
+    }
+
+    static public class ParamHelper implements ParamSerializer {
```
huitseeker: This should be a singleton.

dmmiller612:
@huitseeker Cool. Looks like the site is down at the moment. I will check it out later. Thanks for the good notes! I will add the other classes soon.

huitseeker:
It works for me but just in case, the google cache is up to date:
https://webcache.googleusercontent.com/search?q=cache:https://www.nicolaferraro.me/2016/02/22/using-non-serializable-objects-in-apache-spark/

PR #3114, opened 2017-03-25
Paralelwrapperdecoupling

agibsonccc:
Decouples parallelwrapper from the trainer implementation.


agibsonccc:
Yeah major idea there was allowing us some flexibility without having to rewrite parallelwrapper multiple times.
We'll want this for spark too.

AlexDBlack:
@agibsonccc for the record: the TrainingMaster interface was intended for that purpose - no changes to Dl4jSparkMultiLayer or SparkComputationGraph, but swap the actual execution approach (averaging, async etc)

PR #3127, opened 2017-03-27
RegressionEvaluation argument handling improvement

vlovsky:


deeplearning4j-nn/src/main/java/org/deeplearning4j/eval/RegressionEvaluation.java
```diff
@@ -107,6 +107,11 @@ public void eval(INDArray labels, INDArray predictions) {
         //https://en.wikipedia.org/wiki/Pearson_product-moment_correlation_coefficient#For_a_sample
         //Doing online calculation of means, sum of squares, etc.
 
+        if (columnNames.size() != labels.size(1) || columnNames.size() != predictions.size(1)) {
+            throw new IllegalArgumentException(
```
agibsonccc: Maybe change this to -1? @AlexDBlack how are we handling time series? Are we assuming this is multi dimensional or are we going to add an evalTimeSeries similar to classification?

vlovsky:
labelsSumPerColumn.addi(labels.sum(0));  labelsSumPerColumn is created as a row vector. this would need to be changed first. Otherwise, there is no real benefit from -1

deeplearning4j-nn/src/main/java/org/deeplearning4j/eval/RegressionEvaluation.java
vlovsky: labelsSumPerColumn.addi(labels.sum(0)); labelsSumPerColumn is created as a row vector. this would need to be changed first. Otherwise, there is no real benefit from -1. Maybe just to not to revisit this in future. But otherwise either way works fine I think.

PR #3165, opened 2017-03-31
Add LSTM implementation (no peephole connections)

AlexDBlack:
***WIP DO NOT MERGE***

This PR adds LSTM class (same as GravesLSTM, minus the peephole connections). This is also the variant that CuDNN (and Keras?) implement.

Needs additional unit tests, but seems to be in good shape already - is passing all gradient checks.


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/LSTM.java
```diff
@@ -0,0 +1,84 @@
...
+    private LSTM(Builder builder) {
+        super(builder);
+        this.forgetGateBiasInit = builder.forgetGateBiasInit;
+        this.gateActivationFn = builder.gateActivationFn;
+    }
+
+    @Override
+    public Layer instantiate(NeuralNetConfiguration conf, Collection<IterationListener> iterationListeners,
+                    int layerIndex, INDArray layerParamsView, boolean initializeParams) {
+        LayerValidation.assertNInNOutSet("GravesLSTM", getLayerName(), layerIndex, getNIn(), getNOut());
```
agibsonccc: Shouldn't this be LSTM and not GravesLSTM?

AlexDBlack: Yep, good catch, I'll fix that now. And re: backward compatibility: separate layers (i.e., GravesLSTM and LSTM will co-exist) so there shouldn't be any issues there.

Open to suggestions on the naming - though it seems like the no-peephole model (ie., this one) is in a sense more standard these days. Hence my thinking with just calling it "LSTM".

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/AbstractLSTM.java
```diff
@@ -0,0 +1,148 @@
...
+            return (T)gateActivationFunction(Activation.fromString(gateActivationFn));
+        }
+
+        /**
+         * Activation function for the LSTM gates.
+         * Note: This should be bounded to range 0-1: sigmoid or hard sigmoid, for example
+         *
+         * @param gateActivationFn Activation function for the LSTM gates
+         */
+        public T gateActivationFunction(Activation gateActivationFn) {
```
agibsonccc: Maybe add a validation warning here? It shouldn't be hard to detect which activation functions have this property.

huitseeker:
rebase here https://github.com/huitseeker/deeplearning4j/tree/pr/3165

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/3221

agibsonccc:
I'm going to start closing PRs that don't get finished. Could we merge this if we're ready?

saudet:
Seems to work well from the tests that I have been running with it!

AlexDBlack:
I'll re-run tests after updating to latest master, and merge soon there-after. It should be good to go otherwise.

Issue #2863, opened 2017-02-14
deeplearning4j-keras module creates shaded jar


AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-keras/pom.xml#L127

Currently, a build/install results in a 96MB shaded jar - this is bad practice, and we don't want to be releasing this onto maven central.

![image](https://cloud.githubusercontent.com/assets/2360237/22924998/476384ae-f2fb-11e6-98b6-1513539c22fe.png)

huitseeker:
@crockpotveggies

crockpotveggies:
Yup this would be an artifact from its original author. I can remove it in my current branch.

PR #3169, opened 2017-03-31
Fix https://github.com/deeplearning4j/deeplearning4j/issues/2863#issuecomment-287263842

huitseeker:
One issue that almost fell by the wayside


crockpotveggies:
Since we're ditching our current approach with Keras API, perhaps instead of removing shading we should just remove the keras submodule altogether until we complete the new approach?

agibsonccc:
I don't agree with this, I'm working on the keras integration now and will be reusing some of this. I'm just going to get rid of creating the uber jar for now.

PR #3180, opened 2017-04-02
added epochs to spark ml

dmmiller612:
I noticed a difference in the model performance when using multilayer configuration's iterations, than when calling the sparkNet.fit function over a series of the same number of epochs. I don't know if that is expected, but I added this epoch parameter, just in case.


agibsonccc:
Yeah we typically loop at the java level so this makes sense to add.

dmmiller612:
Cool, yeah I should have thought of it earlier.

PR #3195, opened 2017-04-05
Implement layer normalization for LSTMs (https://arxiv.org/pdf/1607.6450.pdf)

fac2003:

 LSTM configurations offer the setUseLayerNormalization method. When set to true, activations are normalized by mean and standard deviation. This improves the speed of training (faster convergence).

Also adjust Builders to propagate their variables when given another builder and have a no arg constructor, for consistency with the builder pattern (and because I could not put useLayerNormalization in BaseRecurrentLayer.Builder otherwise).
Added two tests in GravesBidirectionalLSTMTest to run with useLayerNormalization true and false.

## How was this patch tested?
 I have not been able to run any tests in master on my machine, but have modified the tests so that both useLayerNormalization =true and false are run. 

I tested the changes manually with CPU and GPU backends and verified improved convergence. 

Please review
https://github.com/deeplearning4j/deeplearning4j/blob/master/CONTRIBUTING.md before opening a pull request.


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
```diff
@@ -249,6 +251,15 @@ static public FwdPassReturn activateHelper(final Layer layer, final NeuralNetCon
                 currHiddenUnitActivations.muliColumnVector(timeStepMaskColumn);
                 currentMemoryCellState.muliColumnVector(timeStepMaskColumn);
             }
+           // boolean layerNormalization = true;
+            if (useLayerNormalization) {
+                Number stdev = currHiddenUnitActivations.stdNumber();
+                currHiddenUnitActivations = currHiddenUnitActivations.sub(Nd4j.mean(currHiddenUnitActivations));
```
agibsonccc: Use subi (in place) rather than blah = blah.sub This saves a whole allocation.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
```diff
@@ -249,6 +251,15 @@ static public FwdPassReturn activateHelper(final Layer layer, final NeuralNetCon
...
+                if (stdev.floatValue() == stdev.floatValue()) {
+                    // don't divide by NaN. i.e., all zero activations.
+                    currHiddenUnitActivations = currHiddenUnitActivations.div(stdev);
+                }
```
agibsonccc: Maybe a small eps would be better here? Calculating the std deviation twice is expensive.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
```diff
@@ -249,6 +251,15 @@ static public FwdPassReturn activateHelper(final Layer layer, final NeuralNetCon
                 currHiddenUnitActivations.muliColumnVector(timeStepMaskColumn);
                 currentMemoryCellState.muliColumnVector(timeStepMaskColumn);
             }
+           // boolean layerNormalization = true;
+            if (useLayerNormalization) {
+                Number stdev = currHiddenUnitActivations.stdNumber();
```
agibsonccc: I don't have much context on this part but the standard deviation number is for the whole activation. Shouldn't we be doing the std of each column? Typically zeor mean unit variance is column wise not for the whole array.

agibsonccc: Also - shouldn't we be accounting for masks here as well?

fac2003: See equations 3 and 4 in the referenced preprint (https://arxiv.org/pdf/1607.06450.pdf). From what I understand, the mean and standard deviations are for all activations, not column-wise. They are scalars in the equations. I think that is why the normalization can be applied to a mini-batch of size 1 (as claimed in the preprint as an advantage over batch normalization).

Regarding masks, it is possible handling masks could improve the results, or not. I have tried this on a problem where we use masking, and found adequate normalization of the activations without explicitly handling the masks. There is also no information in the preprint about masking, so I think the implementation is accurate in this respect.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
fac2003: Sounds good. I'll integrate this.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
fac2003: The test is testing for equality, not small differences. A float is equal to itself iff it is not a NaN (in Java, see 6 in http://stackoverflow.com/questions/1456566/how-do-you-test-to-see-if-a-double-is-equal-to-nan). 

I will refactor the floatValue() call to a variable to reduce the redundant computation.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/BaseRecurrentLayer.java
```diff
@@ -14,6 +14,16 @@ protected BaseRecurrentLayer(Builder builder) {
         super(builder);
     }
 
+    protected boolean useLayerNormalization;
```
AlexDBlack: Really minor style nitpick: we've been using the following order convention: fields -> constructors -> methods. Move it to the top of the class?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/BaseRecurrentLayer.java
```diff
@@ -42,10 +52,33 @@ public void setNIn(InputType inputType, boolean override) {
...
-
+    //class Builder<T extends FeedForwardLayer.Builder<T>> extends Layer.Builder<T>
 
     @AllArgsConstructor
-    public static abstract class Builder<T extends Builder<T>> extends FeedForwardLayer.Builder<T> {
+    public static abstract class Builder<T extends BaseRecurrentLayer.Builder<T>> extends FeedForwardLayer.Builder<T> {
+        protected boolean useLayerNormalization = false;
+
+
+        public Builder(Builder<T> builder) {
```
AlexDBlack: I don't see why we need these sorts of constructors... We've never needed them before, and there's nothing special about the useLayerNormalization arg (it's just another  boolean)... unless I'm missing something?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
AlexDBlack: Don't need masks here thankfully (it's applied per time step) - that makes the implementation simpler.
@agibsonccc is totally correct about it being activations per time step. That means: currHiddenUnitActivations.std(1), .mean(1) and muliColumnVector/diviColumnVector.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
AlexDBlack: Use if(!Double.isNaN(stdev.doubleValue()) instead
It's clearer/cleaner, and we might as well use double precision if available (very small chance of overflow with floats).

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/BaseRecurrentLayer.java
fac2003: You have such constructors in the code base, just not consistently: see BaseRecurrentLayer, BaseOutputLayer. The goal is to construct a builder from another and transfer the configuration to the new instance. Whether this is needed in practice is unclear, but since (1) I saw examples where it was done correctly (see BaseOutputLayer) and (2) I had to because the super class does not have a no arg constructor, I also implemented the functionality. 
Removing the constructor with builder won't compile unless I remove it from the superclass as well (BaseRecurrentLayer).

fac2003:
Re:
You've only implemented the forward pass in LSTM helpers - not the backward pass.

I understood the helper was used for both forward and backward. This is written in the javadoc at the top:

 /** Returns FwdPassReturn object with activations/INDArrays. Allows activateHelper to be used for forward pass, backward pass
     * and rnnTimeStep whilst being reasonably efficient for all */
    static public FwdPassReturn activateHelper(...)

Furthermore, the code just below the insertion shows how the normalized values are used in the backprop (unless I am missing something):

if (forBackprop) {
                toReturn.fwdPassOutputAsArrays[time] = currHiddenUnitActivations;
..

If the normalization is not used for backprop, where is the code that performs backprop for LSTMs?

fac2003:
I think the new commits address the comments. I tested the suggestion to normalize by column (records of the minibatch), but don't see much improvement in practice, at least on the use case I was evaluating:

cursor_and_layer_normalization_comparison_-_google_sheets

The code now makes it possible to test normalization on records or on a whole minibatch.

The normalization code is in its own class (see LayerNormalization and implementations), and should be reusable for other types of layers. I don't know these well enough to know where to use the normalization though.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LayerNormalization.java
```diff
@@ -0,0 +1,10 @@
+package org.deeplearning4j.nn.layers.recurrent;
+
+import org.nd4j.linalg.api.ndarray.INDArray;
+
+/**
+ * Created by fac2003 on 4/6/17.
+ */
+public interface LayerNormalization {
+    void normalize(INDArray activations);
```
agibsonccc: Javadoc?

agibsonccc: Could also use an @see annotation on the implementation.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LayerNormalizationPerRecord.java
```diff
@@ -0,0 +1,23 @@
...
+ * on one use case though, your experience may differ).
+ * Created by fac2003 on 4/6/17.
+ */
+public class LayerNormalizationPerRecord implements LayerNormalization {
+    @Override
+    public void normalize(INDArray activations) {
+        INDArray stdevVector = activations.std(1);
+        activations.subiColumnVector(Nd4j.mean(activations, 1));
+        if (!Double.isNaN(stdevVector.sum().getDouble(0))) {
+            // don't divide by NaN. i.e., all zero activations.
```
agibsonccc: So this scares me a bit. If we add an epsilon instead we can do some optimizations later to make it inline.Reduction algorithms (sums,..) are expensive.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LayerNormalizationWholeMinibatch.java
```diff
@@ -0,0 +1,21 @@
...
+ * Created by fac2003 on 4/6/17.
+ */
+public class LayerNormalizationWholeMinibatch implements LayerNormalization {
+    @Override
+    public void normalize(INDArray activations) {
+        // normalize across entires mini-batch:
+        Number stdev = activations.stdNumber();
+        activations.subi(Nd4j.mean(activations));
+        final double stdNumber = stdev.doubleValue();
+        if (!Double.isNaN(stdNumber)) {
```
agibsonccc: So this scares me a bit. If we add an epsilon instead we can do some optimizations later to make it inline.Reduction algorithms (sums,..) are expensive.

deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/recurrent/GravesBidirectionalLSTMTest.java
```diff
@@ -169,21 +177,22 @@ public void testGravesBidirectionalLSTMForwardPassHelper() throws Exception {
...
 
+        LayerNormalization layerNormalization=useLayerNormalization?new LayerNormalizationWholeMinibatch() : new NoNormalization();
         final INDArray fwdPassFalse = LSTMHelpers.activateHelper(lstm, lstm.conf(), new ActivationSigmoid(),
                         lstm.input(),
                         lstm.getParam(GravesBidirectionalLSTMParamInitializer.RECURRENT_WEIGHT_KEY_FORWARDS),
                         lstm.getParam(GravesBidirectionalLSTMParamInitializer.INPUT_WEIGHT_KEY_FORWARDS),
                         lstm.getParam(GravesBidirectionalLSTMParamInitializer.BIAS_KEY_FORWARDS), false, null, null,
                         false, true, GravesBidirectionalLSTMParamInitializer.INPUT_WEIGHT_KEY_FORWARDS,
-                        null).fwdPassOutput;
+                        null,layerNormalization).fwdPassOutput;
```
agibsonccc: Could we use getters here?

deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/recurrent/GravesBidirectionalLSTMTest.java
```diff
@@ -169,21 +177,22 @@ public void testGravesBidirectionalLSTMForwardPassHelper() throws Exception {
...
 
         final INDArray[] fwdPassTrue = LSTMHelpers.activateHelper(lstm, lstm.conf(), new ActivationSigmoid(),
                         lstm.input(),
                         lstm.getParam(GravesBidirectionalLSTMParamInitializer.RECURRENT_WEIGHT_KEY_FORWARDS),
                         lstm.getParam(GravesBidirectionalLSTMParamInitializer.INPUT_WEIGHT_KEY_FORWARDS),
                         lstm.getParam(GravesBidirectionalLSTMParamInitializer.BIAS_KEY_FORWARDS), false, null, null,
                         true, true, GravesBidirectionalLSTMParamInitializer.INPUT_WEIGHT_KEY_FORWARDS,
-                        null).fwdPassOutputAsArrays;
+                        null,layerNormalization).fwdPassOutputAsArrays;
```
agibsonccc: Same - getters?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LayerNormalizationPerRecord.java
fac2003: The per record implementation is not recommended for general use. I added it only to show you what the suggestion to normalize by column would get you: not very much (compare yellow and red curves in the figure).

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LayerNormalizationWholeMinibatch.java
fac2003: I prefer to focus on correctness and worry about optimization later if shown to matter by profiling in various scenarios. 
In this case, also, if you look at the figure, the efficiency of training is about doubled (need half the number of epochs to get to the same score), so I am not too concerned by slightly inefficient code.
If you want to optimize, I would suggest to add a hasAnyNaN method to INDArray. It should return true if any element (i.e., first one encountered) is NaN. This can be parallelized any way you like. I can create an issue if this is something you would find useful (I was surprised not to find this method).

deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/recurrent/GravesBidirectionalLSTMTest.java
fac2003: Not sure what you mean. A getter on what? It's in a test with no access to the configuration instance, called for each type of layer.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LayerNormalization.java
fac2003: And I thought I was following the project unwritten guidelines for documentation :-)
Fine, I will add a doc for the interface. The @see is unnecessary when you use a modern IDE (i.e.,  click on the arrow down on an interface to get a list of implementations in IDEA).

AlexDBlack:
@fac2003 We have activateHelper for the forward pass, and backpropGradientHelper for backprop in LSTMHelpers.
The point that I was making: let X be the pre-normalization input activations, and Y be post normalization activations
You are correctly using Y during the forward pass, but then returning dL/dY instead of dL/dX during backprop. You can't do this, and it won't pass gradient checks (and I'm extremely hesitant to merge any changes like this that aren't gradient checked). You'll need an extra step during backprop to calculate and return dL/dX from dL/dY.

Re: normalizePerRecord - that's actually what is suggested in the paper as the standard/default approach - see the discussion below equation 3 (end of page 2).

I'm not really happy with the LayerNormalization interface... first, there's no way to do the backprop currently (as discussed above), and second: if we need per-layer customization (for example, 'whole minibatch' layer normalization must take into account masking in RNNs, or the fact that it's done per time step for RNNs in general) then we either (a) can't do it, or (b) need a whole lot of layer-specific code and special cases in the implementations.
A better approach perhaps: add two methods to the layer interface... applyLayerNormalization and backpropLayerNormalization. Standard implementation would go in BaseLayer, overridden if/when necessary for other layers like RNNs (for example, in BaseRecurrentLayer).

fac2003:
Re: backprop and gradients. Could you point me to existing gradient tests for LSTMs that I could run? If these tests fail then clearly the work is not done and backprop needs to be addressed as well. I don't think I know enough right now to implement gradient tests (and if I did I would not trust that they check what you think should happen). So if such tests exist in the code base, great, I will see if I can make them pass with layer normalization on.

agibsonccc:
This is our gradient checks: https://github.com/deeplearning4j/deeplearning4j/blob/17d6e46b4c2ca23d9231afdaaa5b3019b6199ca2/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNNGradientCheckTest.java 

Please rebase to get rid of conflicts thanks!

fac2003:
i confirmed that the gradient tests do not pass (relevant class is GradientCheckTestsMasking).
I took at look at the backpropGradientHelper method, but I have no idea what it is doing. There's no javadoc to explain what the parameters are and I don't know what most of the Nd4j methods do, so I don't think I can help with this. I am closing the PR.

Issue #3190, opened 2017-04-04
Keras model import is unstable 


thammegowda:
#### Keras model importer crashes due to issues in `libhdf5.so`

We have used  Keras model importer tools to import Inception-V3 model to DL4J.
It works *most of the times*, however, occasionally JVM crashes due to issue in the native code.

Out of 100 times to test the model-import and to classify a test image,  DL4J crashed 6 times for the following reason:
```
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fddefb988e4, pid=15318, tid=0x00007fddf410d700
#
# JRE version: Java(TM) SE Runtime Environment (8.0_101-b13) (build 1.8.0_101-b13)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.101-b13 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libhdf5.so.100+0x1138e4]  H5FL_reg_free+0x84
#
# Core dump written. Default location: /home/tg/work/projects/apache/tika/core or core.15318
#
# An error report file with more information is saved as:
# /home/tg/work/projects/apache/tika/hs_err_pid15318.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
```

By setting a system property `-Dorg.bytedeco.javacpp.nopointergc=true`, the crash can be avoided, however it results in memory leak and demands excessive memory.

Details to reproduce the bug are here https://github.com/apache/tika/pull/165#issuecomment-291394161 and https://github.com/apache/tika/pull/165#issuecomment-291402383 

#### Version Information
* Deeplearning4j version : 0.8.0
* platform information : 
```
$ uname -a
Linux hackb0x 4.4.0-21-generic #37-Ubuntu SMP Mon Apr 18 18:33:37 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```
```
JRE version: Java(TM) SE Runtime Environment (8.0_101-b13) (build 1.8.0_101-b13)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.101-b13 mixed mode linux-amd64 compressed oops)
```
* CUDA version : Not used
* NVIDIA driver version: Not used

saudet:
This is most likely caused by an object from HDF5 being garbage collected prematurely.

[...]

PR #3197, opened 2017-04-06
Fix potential memory corruption from occuring in HDF5

saudet:

Make sure memory used by HDF5 doesn't get prematurely deallocated. See #3190 

## How was this patch tested?

Unit tests pass.


chrismattmann:
AWESOME 👍

PR #3201, opened 2017-04-06
Handles Transforms

dmmiller612:
For the spark multilayer network, I had to use the actual multilayer network to make it serializable for transforms. Spark 2.0 makes heavy use of transforms, and in fact, has the predict method as protected (which hopefully changes in the future). For users to use spark's evaluation metrics and batch predictions, the network will have to be serializable.


deeplearning4j-scaleout/spark/dl4j-spark-ml/src/test/java/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkTest.java
```diff
@@ -44,13 +46,17 @@ public void testNetwork() {
...
-        SparkDl4jNetwork sparkDl4jNetwork = new SparkDl4jNetwork(mc, 2, ps, 1, il)
+        SparkDl4jNetwork sparkDl4jNetwork = new SparkDl4jNetwork(mc, 2, ps, 1, il, true)
                 .setFeaturesCol("features")
                 .setLabelCol("label");
 
         SparkDl4jModel sm = sparkDl4jNetwork.fit(part2.get());
         MultiLayerNetwork mln = sm.getMultiLayerNetwork();
         Assert.assertNotNull(mln);
+        DatasetFacade transformed = DatasetFacade.dataRows(sm.transform(part2.get()));
+        List<?> rows = transformed.get().collectAsList();
```
dmmiller612: test showing that both transformformations and stats still work.

huitseeker:
Thanks for your contribution! I'll take a look at it as soon as possible!

raver119:
predict is the same output() underneath, which is public.

dmmiller612:
in spark ml methods? Like gradient boosted trees?

raver119:
Ah, sorry, i've thought you're speaking about dl4j predict method

dmmiller612:
ah nope. Dl4j has that as public. I am trying to make it easier for spark 2.0 users to migrate to a dl4j networks with little changes in spark pipelines.

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
```diff
@@ -79,15 +115,28 @@ abstract class SparkDl4jModelWrapper[T, E <: SparkDl4jModelWrapper[T, E]](overri
...
 
     protected def outputTensor(vector: Vector) : INDArray = getMultiLayerNetwork.output(MLLibUtil.toVector(vector))
 
     protected def outputFlattenedTensor(vector: Vector) : Vector = Vectors.dense(flattenTensor(outputTensor(vector)))
 
     protected[SparkDl4jModelWrapper] class SparkDl4jModelWriter(instance: SparkDl4jModelWrapper[T,E]) extends MLWriter {
         override protected def saveImpl(path: String): Unit = {
-            ModelSerializer.writeModel(network.getNetwork, path, true)
+            ModelSerializer.writeModel(network, path, true)
         }
```
agibsonccc: Hmm - this only handles local. I'm wondering if maybe we could have this be a bit more agnostic and write to hdfs where hdfs *could* be local?

dmmiller612: I think that is a good point. I mostly use EMR clusters at work, and s3 would be a really nice asset as well. I am assuming this won't work with s3, but I haven't tried yet.

dmmiller612:
@agibsonccc I got a question for you. I am looking at making some rnn helpers within this class. Do you know if someone is using multiple time series to train a single lstm model, if they need to run `.rnnClearPreviousState()` after each `fit`? or just for the prediction?

deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/SparkDl4jNetworkWrapper.scala
huitseeker: Yes ! This works well, it's actually the approach taken by Spark : to use the HDFS API to access everything, with a `FileSystem` configuration that differs based on the case you are looking at. It often amounts to an import of configuration followed by `val fs = HdfsUtils.getFileSystemForPath(path, hadoopConf)`.

It would indeed be very appropriate to follow suit here.
You can look here for inspiration:
https://github.com/apache/spark/blob/master/streaming/src/main/scala/org/apache/spark/streaming/util/HdfsUtils.scala

dmmiller612: Does DL4J have something already that writes to hdfs?

dmmiller612:
https://github.com/deeplearning4j/deeplearning4j/pull/3253

PR #3253, opened 2017-04-14
Updated the transforms

dmmiller612:
The rebase to my forked repo caused a substantial amount of merge conflicts on things I hadn't touched, so I just put this in a clean new branch.

@huitseeker I am still unclear if there is an hdfs solution for dl4j. If there isn't, I will create a separate issue, and work on it later. I will need it for my project, so I will eventually get to it. I just won't have time right now.

### DIFF
--- a/deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/AutoEncoderWrapper.scala
+++ b/deeplearning4j-scaleout/spark/dl4j-spark-ml/src/main/scala/org/deeplearning4j/spark/ml/impl/AutoEncoderWrapper.scala
@@ -14,6 +14,8 @@ import org.deeplearning4j.util.ModelSerializer
 import org.nd4j.linalg.dataset.DataSet
 import org.nd4j.linalg.factory.Nd4j
 
+import scala.collection.JavaConverters._
+
 
 trait AutoEncoderParams extends Params {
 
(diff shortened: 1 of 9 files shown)


dmmiller612:
my fault, it is supposed to have a teardown method.

dmmiller612:
fixed now.

dmmiller612:
@huitseeker Btw, I figured out a way good way to save to hdfs, and it is in a spark ml way. I'll add it soon (hopefully tomorrow).

huitseeker:
@dmmiller612 Yeah, you're reusing the [filesystem getter and associated utils](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/Utils.scala#L1843) from `SparkUtils` right ? I thought I mentioned that way in a comment to you, but I realize I now did in another PR from somebody else. Sorry. Nice that you figured it out.

dmmiller612:
@huitseeker Actually, Most 2.0 methods don't seem to do that directly. That actually save the parameters to a dataset/dataframe and create the model on load. We can easily do that, and it significantly lowers the footprint of the saved model as well (saving the parameters along with the multilayerconfiguration). Obviously under the hood it does exactly what you linked to, but it is a layer of abstraction.

dmmiller612:
@huitseeker I moved the code over to here, if you want to check it out.

dmmiller612:
sure. I have it available for my own project. I'll add it soon.

dmmiller612:
@huitseeker tests are added.

PR #3258, opened 2017-04-15
Parallelize RecordReaderDataSetIterator

crockpotveggies:

This PR parallelizes the RRDSI next(num) operation. DataSets returned are guaranteed fair order.

## How was this patch tested?

*In progress*. Record readers will need to be thread-safe.


crockpotveggies:
Hmm, I guess I could use `Thread.currentThread()` to determine which thread in the thread pool is being used. And then I can assign a pipeline to that specific thread if it hasn't been seen before.

crockpotveggies:
@AlexDBlack hopped on the phone with raver and he walked me through a solution that solves your concern. Basically, we can parallelize RRDSI w/ fair order *after* calling RR.next(). Internally on IRR, we can modify it so it works like AsyncInterator and performs prefetch. Changes incoming.

huitseeker:
For people like me who are looking, @crockpotveggies means https://github.com/deeplearning4j/DataVec/pull/233/ is a prerequisite.

crockpotveggies:
Going to close this in favor of recent changes by @raver119. If you change your mind we can reopen it.

Issue #3270, opened 2017-04-19
save datasets in parallel manner


kepricon:
# What I'm going to do
write **DataSetSaver**(this name can be changed) which can save datasets in parallel mode

# purpose
there are some situations we should **make pre-saved datasets**. For examples, benchmarks(https://deeplearning4j.org/benchmark#etl--asynchronous-etl) or training on large image datasets(like ImageNet). sometimes it takes lots of time. this can be helpful for the situations.

raver119:
Don't forget to make it configurable: number of workers should be configurable at very least.

kepricon:
thanks for your comments. I've applied your opinion in the PR.

huitseeker:
All merged !

PR #3272, opened 2017-04-19
[WIP] parallel DataSetSaver

kepricon:

#3270 

## How was this patch tested?

included unit tests, I've also put Ignore to test class though.

### DIFF
--- a/deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/savers/DataSetSaver.java
+++ b/deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/savers/DataSetSaver.java
@@ -1,0 +1,72 @@
+package org.deeplearning4j.datasets.savers;
+
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.FilenameUtils;
+import org.nd4j.linalg.dataset.api.DataSet;
+import org.nd4j.linalg.dataset.api.iterator.DataSetIterator;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import java.io.File;
+
+/**
+ * Created by kepricon on 17. 4. 19.
+ */
+@Slf4j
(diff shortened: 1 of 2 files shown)


crockpotveggies:
One thought here is since this is ETL related, it might be better served to live in DataVec. @AlexDBlack what are your thoughts on that?

crockpotveggies:
Another reason we should move this to either DataVec or ND4J: `ExistingMiniBatchDataSetIterator`, which is the class that uses this on the other side of the pipeline, exists in ND4J.

kepricon:
that makes sense. Thanks

i'll close this and re-open in either ND4J or DataVec

kepricon:
as discussed with @raver119, I decided to locate this stuff in dl4j repo.

cc @crockpotveggies

deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/savers/DataSetSaver.java
```diff
@@ -0,0 +1,111 @@
...
+    public static void setNumThreads(int numThreads) {
+        DataSetSaver.numThreads = numThreads;
+    }
+
+    public static void setQueueCapacity(int qCapacity) {
+        DataSetSaver.QUEUE_CAPACITY = qCapacity;
+    }
+
+    private static void initExecutors() {
+        taskExecutor = new ThreadPoolExecutor(numThreads, numThreads,
```
raver119: In this case it might be easier to use something like Executors.newFixedPoolExecutor() with custom Thread factory. Like this: https://github.com/deeplearning4j/deeplearning4j/blob/4bafd6c7b108d1c1d03d97bf6b3ce4f51c4da304/deeplearning4j-core/src/main/java/org/deeplearning4j/datasets/datavec/RecordReaderDataSetIterator.java#L208-L208

This way allows to avoid shutdown, and guarantees threads to be daemonic, so not blocking app exit.

P.s. I'm NOT requesting change here. It's jfyi.

kepricon: thanks

PR #3276, opened 2017-04-19
new AsyncDataSetIterator mechanics

raver119:
This PR introduces new AsyncDataSetIterator mechanics, as was discussed with @AlexDBlack ages ago.

Absolutely same functionality provided, just less cryptic code. Similar pattern is used for last few months in NLP. Now it's also available for neural nets.


deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
```diff
@@ -60,14 +40,14 @@ public AsyncDataSetIterator(DataSetIterator iterator, int queueSize, BlockingQue
...
     }
 
-    /**
-     * Create an AsyncDataSetIterator with a queue size of 1 (i.e., only load a
-     * single additional DataSet)
-     *
-     * @param iterator
-     * @param queueSize
-     * @param queue BlockingQueue instance that will be used as backing queue. MagicQueue probably?
-     */
```
agibsonccc: Any reason we got rid of the javadoc?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
```diff
@@ -113,291 +91,282 @@ public AsyncDataSetIterator(DataSetIterator iterator, int queueSize, BlockingQue
...
-            runnable.runCompletedSemaphore.tryAcquire(5, TimeUnit.SECONDS);
-        } catch (InterruptedException e) {
-        }
-
-        //Clear the queue, reset the base iterator, set up a new thread
-        blockingQueue.clear();
-        baseIterator.reset();
-        runnable = new IteratorRunnable(baseIterator.hasNext(), this.workspace);
-        thread = runnable;
+    public void reset() {
```
agibsonccc: Doesn't this need to be synchronized still?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
```diff
@@ -60,14 +40,14 @@ public AsyncDataSetIterator(DataSetIterator iterator, int queueSize, BlockingQue
...
-     * single additional DataSet)
-     *
-     * @param iterator
-     * @param queueSize
-     * @param queue BlockingQueue instance that will be used as backing queue. MagicQueue probably?
-     */
+    public AsyncDataSetIterator(DataSetIterator baseIterator, int queueSize) {
+        this(baseIterator, queueSize, new LinkedBlockingQueue<DataSet>(queueSize));
+    }
+
```
agibsonccc: Any reason we can't pass in our own workspace?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
raver119: No, it doesn't. ADSI isn't (and wasn't) supposed to be thread-safe, training is single-threaded by design.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
raver119: Yep, it was too late for it :)
Will add today.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
raver119: Sure, this reasons is learnable size. We want workspace to learn memory requirements of a given DataSetIterator, and match that requirements. It's simple learning at this moment, but i'll improve it today, as discussed with @AlexDBlack earlier.

raver119: I mean - each DataSet has its own size, it's easy - i've added method for that - DataSet.getMemoryFootprint(), there's a logic flaw here: DataSet final size != amount of memory used to produce this DataSet. And this should be improved.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
```diff
@@ -113,291 +93,282 @@ public AsyncDataSetIterator(DataSetIterator iterator, int queueSize, BlockingQue
...
-            //exception thrown while getting data from base iterator
-            throw runnable.exception;
-        } catch (InterruptedException e) {
-            throw new RuntimeException(e); //Shouldn't happen under normal circumstances
+            if (nextElement == terminator)
+                return false;
+
+            return true;
+        } catch (Exception e) {
+            log.error("Premature end of loop!");
```
AlexDBlack: Exception should be logged also for debugging purposes?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
```diff
@@ -113,291 +93,282 @@ public AsyncDataSetIterator(DataSetIterator iterator, int queueSize, BlockingQue
...
+
+
+        if (thread != null)
+            thread.interrupt();
+        try {
+            // Shutdown() should be a synchronous operation since the iterator is reset after shutdown() is
+            // called in AsyncLabelAwareIterator.reset().
+            if (thread != null)
+                thread.join();
+        } catch (InterruptedException e) {
```
AlexDBlack: This usually shouldn't happen, but should we be silently swallowing the exception here if it does?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
raver119: There shouldn't be any besides of InterruptedException, really. It's only buffer.take() happens here.
And real exceptions are propagated from prefetch thread.
However, let's just add it.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
raver119: Ok, let's just throw it out, not a big deal :)

raver119:
New functionality added to this PR:
- New implementation for pre-saved datasets iterator, to address multi-gpu environments
- ETL time reports for ParallelWrapper with PerformanceListener
- Builder pattern for RecordReaderDataSetIterator
- Workspaces adopted for model.score(DataSet)

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/parallel/ParallelExistingMiniBatchDataSetIterator.java
```diff
@@ -0,0 +1,397 @@
...
+    public boolean asyncSupported() {
+        return true;
+    }
+
+    @Override
+    public void reset() {
+        shouldWork.set(false);
+        thread.interrupt();
+        thread.shutdown();
+        buffer.clear();
```
AlexDBlack: This may cause a RC if the thread doesn't shut down quickly enough while prefetching - it could in principle still add to the buffer?

raver119: no, thread.shutdown() call is blocking. so buffer.clear() is guaranteed to happen only after current thread is dead.

PR #3286, opened 2017-04-20
Only rebuild needed modules between different scala/spark version builds

### LINKED ISSUE #2873: Optimize scala versions build script
Currently, buildmultiplescalaversions.sh does everything 3 times (spark 1 + scala 2.10 & 2.11, and spark 2 + scala 2.11). But most of the project doesn't have a scala/spark dependency. Consequently, it should be possible to build most of the project modules only once, instead of 3 times.

huitseeker:
Reduces build time of `./buildmultiplescalaversions.sh clean install -DskipTests=true`
 from 293s to 199s
Fixes #2873

### DIFF
--- a/buildmultiplescalaversions.sh
+++ b/buildmultiplescalaversions.sh
@@ -13,7 +13,7 @@ set -eu
 ./change-scala-versions.sh 2.11 # should be idempotent, this is the default
 mvn "$@"
 ./change-spark-versions.sh 2
-mvn -Dspark.major.version=2 -Dmaven.clean.skip=true -pl $(whatchanged| tr '\n' ',') -amd "$@"
+mvn -Dmaven.clean.skip=true -pl $(whatchanged| tr '\n' ',') -amd "$@"
 ./change-scala-versions.sh 2.10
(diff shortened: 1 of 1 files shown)


buildmultiplescalaversions.sh
```diff
@@ -1,10 +1,20 @@
...
+	cd $BASEDIR
+    done
+}
+
 set -eu
 ./change-scala-versions.sh 2.11 # should be idempotent, this is the default
 mvn "$@"
 ./change-spark-versions.sh 2
-mvn -Dspark.major.version=2 "$@"
+mvn -Dspark.major.version=2 -Dmaven.clean.skip=true -pl $(whatchanged| tr '\n' ',') -amd "$@"
```
AlexDBlack: Don't need/want ```-Dspark.major.version=2``` here - it doesn't matter with the change-spark-versions.sh earlier, and it wouldn't be flattened anyway (i.e., value would still read from pom, instead of being substituted/hard-coded in).

huitseeker:
@crockpotveggies it's the `-amd` option : it rebuilds these dependent modules.

PR #3311, opened 2017-04-26
ParameterAveragingTrainingMaster should use treeAggregate

sethah:

This patch adds a new parameter to the `ParameterAveragingTrainingMaster` called `aggregationDepth` which controls the level of tree aggregation used during parameter averaging. This patch changes `processResults` to use `RDD.treeAggregate` instead of `RDD.aggregate`. 

I could not find prior discussion on this either through the issues or through gitter. I had problems doing transfer learning on VGG16 using Spark, as my driver was running out of memory. Tree aggregation should help alleviate this. Though I'm still new to the project, I can't see a reason not to use tree aggregate here.

This patch also modifies a couple of public constructors, but from taking a look at some past patches, this does not seem to be an issue? I think it could be avoided by adding another alternate constructor.

## How was this patch tested?

I am more than happy to add unit tests. For now, I did not since things like parameter bound checks do not seem to be tested in the current code, and a unit test for this does not seem to me to fit in any of the existing test modules. So for now, I've added the parameter to a couple of existing usages in the tests. Please let me know if I should go ahead and add unit tests to check the bounds on the aggregation depth parameter.

### DIFF
--- a/deeplearning4j-scaleout/spark/dl4j-spark/src/main/java/org/deeplearning4j/spark/impl/paramavg/ParameterAveragingTrainingMaster.java
+++ b/deeplearning4j-scaleout/spark/dl4j-spark/src/main/java/org/deeplearning4j/spark/impl/paramavg/ParameterAveragingTrainingMaster.java
@@ -1,6 +1,7 @@
 package org.deeplearning4j.spark.impl.paramavg;
 
 import lombok.extern.slf4j.Slf4j;
+import static com.google.common.base.Preconditions.checkArgument;
 import org.apache.spark.api.java.JavaRDDLike;
 import org.deeplearning4j.api.storage.*;
 import org.deeplearning4j.nn.conf.ComputationGraphConfiguration;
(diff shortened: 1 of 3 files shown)


deeplearning4j-scaleout/spark/dl4j-spark/src/main/java/org/deeplearning4j/spark/impl/paramavg/ParameterAveragingTrainingMaster.java
```diff
@@ -160,30 +162,40 @@ public ParameterAveragingTrainingMaster(boolean saveUpdater, Integer numWorkers,
...
-                    RepartitionStrategy repartitionStrategy, StorageLevel storageLevel, boolean collectTrainingStats) {
+                    int batchSizePerWorker, int averagingFrequency, int aggregationDepth, int prefetchNumBatches,
+                    Repartition repartition, RepartitionStrategy repartitionStrategy, StorageLevel storageLevel,
+                    boolean collectTrainingStats) {
         if (numWorkers <= 0)
             throw new IllegalArgumentException("Invalid number of workers: " + numWorkers + " (must be >= 1)");
         if (rddDataSetNumExamples <= 0)
             throw new IllegalArgumentException(
                             "Invalid rdd data set size: " + rddDataSetNumExamples + " (must be >= 1)");
+        if (averagingFrequency <= 0)
```
huitseeker: Great initiative, but we use `[Guava's preconditions](https://github.com/google/guava/wiki/PreconditionsExplained)` to do checks of this sort.

sethah: Just to make sure I understand, you are suggesting:

````java
checkArgument(averagingFrequency <= 0, "Invalid input: averaging frequency must be >= 1");
checkArgument(aggregationDepth <= 0, "Invalid input: tree aggregation depth must be >= 1");
````

If that's correct, shall I replace the other usages of `if (...) throw ...`?

sethah:
@huitseeker I went ahead and changed it to use `checkArgument` for the parameters that had checks previously. Let me know if that's what you had in mind. Thanks for the review!

sethah:
Thanks @huitseeker and @agibsonccc!

huitseeker:
@sethah Just what I meant. Perfect ! Thank *you* !

PR #3330, opened 2017-04-29
Add restoring Normalizer from a model file InputStream

wmeddie:

Adds a method that can restore a normalizer that is included in a model file.

The new method is restoreNormalizerFromInputStream instead of an overload because the File version is named in an objective-c style restoreNormalizerFromFile.

## How was this patch tested?

I've added tests for the happy and sad case and verified that the restored normalizer restores values correctly.

### DIFF
--- a/deeplearning4j-core/src/test/java/org/deeplearning4j/util/ModelSerializerTest.java
+++ b/deeplearning4j-core/src/test/java/org/deeplearning4j/util/ModelSerializerTest.java
@@ -1,5 +1,6 @@
 package org.deeplearning4j.util;
 
+import org.datavec.api.transform.transform.doubletransform.MinMaxNormalizer;
 import org.deeplearning4j.datasets.iterator.impl.IrisDataSetIterator;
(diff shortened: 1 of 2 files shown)


deeplearning4j-core/src/test/java/org/deeplearning4j/util/ModelSerializerTest.java
```diff
@@ -153,4 +159,80 @@ public void testWriteCGModelInputStream() throws Exception {
...
+        ModelSerializer.writeModel(cg, tempFile, true);
+
+        ModelSerializer.addNormalizerToModel(tempFile, norm);
+        FileInputStream fis = new FileInputStream(tempFile);
+
+
+        NormalizerStandardize restored = ModelSerializer.restoreNormalizerFromInputStream(fis);
+
+        assertNotEquals(null, restored);
+
```
agibsonccc: Technically works but also could do assumeNotNull

wmeddie: Noted.

Issue #3143, opened 2017-03-29
Support for cuDNN 6.0


crockpotveggies:
Release candidate is available for cuDNN 6.0. Downloads are available via http://blog.yannisassael.com/2017/02/cudnn-v6-0-rc/.

At surface level does not appear to be any breaking changes with existing code. @saudet would I need to recompile javacpp bindings or can I simply drop in the library files and give it a whirl?

Personally I'm wondering if this fixes issues like:
- https://github.com/deeplearning4j/deeplearning4j/issues/3091
- https://github.com/deeplearning4j/deeplearning4j/issues/3064

saudet:
You could try to create a link to `libcudnn.so.5` and such and see what that gives, but I doubt that will work..

agibsonccc:
@saudet fixed this

saudet:
@crockpotveggies Let us know if this fixes the issues you've been having with cuDNN 5!

crockpotveggies:
Unfortunate news, it looks like there's been API changes in cuDNN 6 that are affecting us. Our examples run fine since they're smaller convolutions, however they fail when trying to run larger models like Inception or AlexNet. I've reproduced this across multiple projects.

Do we have an NVIDIA dev who can give us some insight on what exactly was dropped?

crockpotveggies:
@agibsonccc please reopen

crockpotveggies:
So when using a convolutional model such as AlexNet or Inception that uses SAME padding and also configures padding >0 in convolutions, cuDNN 6 will return a CUDNN_STATUS_NOT_SUPPORTED.

It appears that when input images are > 1 channel, this error will be thrown. I'm not quite sure why, but that currently seems to be the only problem with support for cuDNN 6.

crockpotveggies:
I've narrowed this to a potential "bug" or "feature" in cuDNN (depending on how you look at it).

It *looks* like `cudnnGetConvolutionForwardAlgorithm` eventually calls `cudnnGetConvolutionForwardWorkspaceSize` when selecting an algorithm for convolution. According to the user guide, a CUDNN_STATUS_NOT_SUPPORTED is thrown on this function when:

> The combination of the tensor descriptors, filter descriptor and convolution descriptor is not supported for the specified algorithm.

After browsing the User Guide, this can be the only explanation for why we're getting this specific error on `cudnnGetConvolutionForwardAlgorithm`, even though the docs itself say this function does not throw an UNSUPPORTED error.

[...]

PR #3339, opened 2017-05-01
Add CudnnLSTMHelper

saudet:

Support for accelerating LSTM with cuDNN on GPU.

## How was this patch tested?

After a 's/GravesLSTM/LSTM/g', the output of BasicRNNExample and GravesLSTMCharModellingExample is similar and run about twice as fast on my GPU.

I still need to add a unit test, and the presets for cuDNN had to be updated, so that also needs to be released before we can merge this.


huitseeker:
Nice ! Plus, small variation on the existing. Mind adding a small example alongside (& derived from) https://github.com/deeplearning4j/dl4j-examples/pull/384 when it's in a better place (& merged) ?

saudet:
@huitseeker It will just load automatically when using LSTM layers. There's usually nothing special to do that would merit an example. Although there are a few cuDNN specific settings I'm wondering what to do about: https://github.com/deeplearning4j/deeplearning4j/issues/3143#issuecomment-291962994
Do you think that adding settings like this to the generic layer configuration be acceptable?

In any case, I've released updated presets for CUDA, and also added a unit test that does gradient checks, but unfortunately, it fails. I hope it's only something wrong with my code...

deeplearning4j-cuda/pom.xml
```diff
@@ -74,8 +74,8 @@
         </dependency>
         <dependency>
             <groupId>org.bytedeco.javacpp-presets</groupId>
-            <artifactId>cuda-platform</artifactId>
-            <version>${cuda.version}-${javacpp-presets.version}</version>
+            <artifactId>cuda</artifactId>
+            <version>${cuda.version}-6.0-1.3</version>
```
AlexDBlack: Hard-coded version

deeplearning4j-cuda/src/main/java/org/deeplearning4j/nn/layers/recurrent/CudnnLSTMHelper.java
```diff
@@ -0,0 +1,648 @@
...
+import static org.bytedeco.javacpp.cuda.*;
+import static org.bytedeco.javacpp.cudnn.*;
+
+/**
+ * cuDNN-based helper for the recurrent LSTM layer (no peephole connections).
+ *
+ * @author saudet
+ */
+public class CudnnLSTMHelper implements LSTMHelper {
+    protected static final Logger log = LoggerFactory.getLogger(CudnnLSTMHelper.class);
```
AlexDBlack: Not a big deal, but could use ```@Slf4j``` lombok annotation instead.

deeplearning4j-cuda/src/main/java/org/deeplearning4j/nn/layers/recurrent/CudnnLSTMHelper.java
```diff
@@ -0,0 +1,648 @@
...
+    // These constants might eventually become variable parameters...
+    static final int numLayers = 1;
+    static final float dropout = 0;
+    static final boolean bidirectional = false;
+    static final int RNNMode = CUDNN_LSTM;
+    static final int numLinearLayers = 8; // CUDNN_LSTM
+
+    static int dataType = Nd4j.dataType() == DataBuffer.Type.DOUBLE ? CUDNN_DATA_DOUBLE
+                        : Nd4j.dataType() == DataBuffer.Type.FLOAT ? CUDNN_DATA_FLOAT : CUDNN_DATA_HALF;
+    static int dataTypeSize = Nd4j.dataType() == DataBuffer.Type.DOUBLE ? 8
```
AlexDBlack: I'm a little worried about these static fields... I can imagine in tests: for example, CudnnLSTMHelper is initialized in a gradient check (double precision) then used again later in a different unit test (with float precision)...

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
```diff
@@ -168,6 +170,15 @@ static public FwdPassReturn activateHelper(final Layer layer, final NeuralNetCon
             prevOutputActivations = Nd4j.zeros(new int[] {miniBatchSize, hiddenLayerSize});
         }
 
+        if (helper != null && Nd4j.dataType() != DataBuffer.Type.HALF) {
```
AlexDBlack: ActivationSigmoid and ActivationTanH: these should be checked at this point too I think - before we get to CudnnLSTMHelper, so we can fall back to our implementation if we have non sigmoid or non tanh activations.
Or: we add a method like activationsSupported(IActivation gates, IActivation afn) on the lstm helper interface.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
```diff
@@ -357,6 +369,16 @@ static public FwdPassReturn activateHelper(final Layer layer, final NeuralNetCon
             rwGradientsGG = rwGradientsOut.get(NDArrayIndex.all(), NDArrayIndex.point(4 * hiddenLayerSize + 2));
         }
 
+        if (helper != null && Nd4j.dataType() != DataBuffer.Type.HALF) {
```
AlexDBlack: As per previous comment checks.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
saudet: We can make the method return null to fall back. So let's do that and log a warning? There are a few other cases like that, so I'll replace all IllegalArgumentException that way. Sounds good?

saudet:
Yes, gradient checks are failing. I've released updated binaries for CUDA and cuDNN so it works transparently without having to build anything. Well, we do need to install the DLL for cuDNN 6.0 in the PATH, but that's it.

Will create a CudnnBaseHelper to take care of code duplication!

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTMHelpers.java
AlexDBlack: Sure, I'd be happy with that - my criteria is basically (a) works seamlessly when we can't use the helper, and (b) the user gets a clear log message (but only once per network, or at most once per layer). Which the null+log approach can satisfy.

AlexDBlack:
@saudet Thanks.
Let me know if you want me to take a look at gradient checks.

FWIW I do recall that the CuDNN CNN implementation was a bit less precise for gradient checks (possibly due to the different algorithms it's using internally). i.e., things would pass at something like the 1e-7 level not the 1e-9 or so level we'd see with our implementation. It's possible something similar is happening here.

saudet:
I took care of most the small issues with #3388, now the only thing remaining is the gradient checks. It's not like with CNNs, the gradients appear approximated in some way. The results of the of tests look like this:
```
o.d.g.GradientCheckUtil - Param 0 (0_W) FAILED: grad= -0.7128224274768038, numericalGrad= -0.7316349865504889, relError= 0.013023962417302255, scorePlus=9.982779084123962, scoreMinus= 9.982793716823693
o.d.g.GradientCheckUtil - Param 1 (0_W) FAILED: grad= -0.6807088992040051, numericalGrad= -0.7105617168434718, relError= 0.021457232902881887, scorePlus=9.982779294842064, scoreMinus= 9.982793506076401
o.d.g.GradientCheckUtil - Param 2 (0_W) passed: grad= -1.0583938543793163, numericalGrad= -1.065967362468001, relError= 0.0035650754818073193
o.d.g.GradientCheckUtil - Param 3 (0_W) FAILED: grad= 0.045644051461737514, numericalGrad= 0.026105233263962187, relError= 0.272320738422313, scorePlus=9.982786661163848, scoreMinus= 9.982786139059183
o.d.g.GradientCheckUtil - Param 4 (0_W) passed: grad= -0.680173054811838, numericalGrad= -0.6879813294169422, relError= 0.005707158998365321
o.d.g.GradientCheckUtil - Param 5 (0_W) FAILED: grad= 0.010573463322012897, numericalGrad= -0.0032696132024057074, relError= 1.0, scorePlus=9.982786367281644, scoreMinus= 9.982786432673908
o.d.g.GradientCheckUtil - Param 6 (0_W) FAILED: grad= -0.5756392534240149, numericalGrad= -0.6028694153670244, relError= 0.023105610220876278, scorePlus=9.982780371464614, scoreMinus= 9.982792428852921
o.d.g.GradientCheckUtil - Param 7 (0_W) FAILED: grad= -0.13536425116385312, numericalGrad= -0.1504382405315141, relError= 0.05274268001738818, scorePlus=9.982784895962146, scoreMinus= 9.982787904726957
o.d.g.GradientCheckUtil - Param 8 (0_W) passed: grad= -0.015615674335483911, numericalGrad= -0.015663808117238887, relError= 0.0015388292254428146
o.d.g.GradientCheckUtil - Param 9 (0_W) passed: grad= -0.017155488096700998, numericalGrad= -0.017450240008543005, relError= 0.00851743130344199
o.d.g.GradientCheckUtil - Param 10 (0_W) passed: grad= -0.014910195547197075, numericalGrad= -0.014845088358583778, relError= 0.002188088301205819
o.d.g.GradientCheckUtil - Param 11 (0_W) passed: grad= -0.016159574851945777, numericalGrad= -0.016093181454124306, relError= 0.002058534073535171
o.d.g.GradientCheckUtil - Param 12 (0_W) FAILED: grad= -0.004558664821885198, numericalGrad= -0.004695243216445988, relError= 0.01475899630675598, scorePlus=9.982786352975461, scoreMinus= 9.982786446880326
o.d.g.GradientCheckUtil - Param 13 (0_W) FAILED: grad= -0.003354030675957785, numericalGrad= -0.0037738506541984402, relError= 0.05889828390723977, scorePlus=9.982786362189445, scoreMinus= 9.982786437666459
o.d.g.GradientCheckUtil - Param 14 (0_W) FAILED: grad= -0.008883353470499228, numericalGrad= -0.009263826239447326, relError= 0.020965944848144024, scorePlus=9.982786307290294, scoreMinus= 9.982786492566818
o.d.g.GradientCheckUtil - Param 15 (0_W) FAILED: grad= -0.002332607523515075, numericalGrad= -0.0030059236877377766, relError= 0.12612386021148447, scorePlus=9.98278636986846, scoreMinus= 9.982786429986934
o.d.g.GradientCheckUtil - Param 16 (0_W) FAILED: grad= -1.3167488657553854E-4, numericalGrad= -1.347648215244135E-4, relError= 0.011597126310151163, scorePlus=9.982786398579995, scoreMinus= 9.982786401275291
o.d.g.GradientCheckUtil - Param 17 (0_W) FAILED: grad= -1.2689822586094997E-4, numericalGrad= -1.2979848307281827E-4, relError= 0.011298380972296043, scorePlus=9.98278639862966, scoreMinus= 9.98278640122563
o.d.g.GradientCheckUtil - Param 18 (0_W) FAILED: grad= -3.612568349806358E-4, numericalGrad= -3.6967886529737365E-4, relError= 0.01152225881638364, scorePlus=9.982786396230878, scoreMinus= 9.982786403624456
o.d.g.GradientCheckUtil - Param 19 (0_W) passed: grad= -5.3938561600882315E-6, numericalGrad= -5.75086644971634E-6, relError= 0.03203402203245764; absolute error = 3.570102896281085E-7 < minAbsoluteError = 1.0E-6
...
```

One thing I'm wondering about, should we be passing something else than those `null` to `cudnnRNNBackwardData()`?
https://github.com/deeplearning4j/deeplearning4j/blob/1f848a9cb8314937c77e53424cbc1a0a93c5bc0e/deeplearning4j-cuda/src/main/java/org/deeplearning4j/nn/layers/recurrent/CudnnLSTMHelper.java#L242

agibsonccc:
Any progress on this? Anywhere I could help test? Would love to get this in to the release.

saudet:
@agibsonccc It still /works/, the gradients that it computes are just not very accurate it seems...

agibsonccc:
We verified this with double data type right?

saudet:
Yes, gradient test use double.

PR #3356, opened 2017-05-04
Parallel ETL backbone

raver119:
This PR introduces symmetric parallel ETL suitable for ParallelWrapper on high-end systems (4+ GPUs):
- Joint parallel iterator that basically does union over multiple existing DataSetIterators
- File-based parallel iterator for pre-saved datasets
- RecordReader-based parallel iterator to be added though

Additionally:
- ParallelWrapper now doesn't have legacyAveraging option, transparent CPU fallback is used instead. https://github.com/deeplearning4j/nd4j/issues/1812
- Minor changes within Async iterators, to match variable input fixes/tweaks


deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/FileSplitDataSetIterator.java
```diff
@@ -0,0 +1,127 @@
...
+
+
+    @Override
+    public DataSet next(int num) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public int totalExamples() {
+        return numFiles;
```
AlexDBlack: For totalExamples(), inputColumns(), totalOutcomes(), batch(), numExamples() etc - I'd personally prefer these to throw exceptions than silently return incorrect values.
Or even better: we clean up the DataSetIterator api, as we don't really ever actually use these methods :)

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/callbacks/DataSetDeserializer.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @author raver119@gmail.com
+ */
+@Slf4j
+public class DataSetDeserializer implements FileCallback {
+    @Override
+    public <T> T call(File file) {
+        DataSet dataSet = new DataSet();
+        dataSet.load(file);
+        return (T) dataSet;
```
AlexDBlack: Not sure I get the purpose of the generic type here?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/callbacks/DefaultCallback.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.nd4j.linalg.api.concurrency.AffinityManager;
+import org.nd4j.linalg.dataset.api.DataSet;
+import org.nd4j.linalg.dataset.api.MultiDataSet;
+import org.nd4j.linalg.factory.Nd4j;
+
+/**
+ * @author raver119@gmail.com
+ */
+public class DefaultCallback implements DataSetCallback {
```
AlexDBlack: I'm being a bit pendantic here: but "DefaultCallback" as a name doesn't really tell people what it does or if/why it should be used. Maybe something like "EnforceAfinityDataSetCallback" might be more informative?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/parallel/FileSplitParallelDataSetIterator.java
```diff
@@ -0,0 +1,107 @@
...
+            int cDev = cnt % numDevices;
+            asyncIterators.add(new AsyncDataSetIterator(new FileSplitDataSetIterator(part, callback), bufferPerThread, true, cDev));
+            cnt++;
+        }
+
+    }
+
+    @Override
+    public boolean hasNextFor(int consumer) {
+        if (consumer >= numProducers)
```
AlexDBlack: Again, being pedantic, but if we're checking, it might as well be ```consumer < 0 || consumer >= numProducers)```

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/parallel/MultiBoolean.java
```diff
@@ -0,0 +1,93 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.nd4j.linalg.exception.ND4JIllegalStateException;
+
+/**
+ * This is utility class, that allows easy handling of multiple joint boolean states
+ *
+ * @author raver119@gmail.com
+ */
+@Slf4j
+public class MultiBoolean {
```
AlexDBlack: Non-thread safe/atomic here? And what's the maximum - 32 values? (This should be in the javadoc)
It's a little inefficient in terms of space, but extending AtomicIntegerArray would be an alternative.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/callbacks/DataSetDeserializer.java
raver119: MultiDataSets... :)

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/callbacks/DefaultCallback.java
raver119: This one will be removed, it's for me only for perf testing, i wonder if it affects performance for any of existing RecordReaders.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/parallel/FileSplitParallelDataSetIterator.java
raver119: True. Will fix.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/FileSplitDataSetIterator.java
raver119: +1 to DataSetIterator change. It should be just Iterator copy, like MultiDataSetIterator

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/parallel/MultiBoolean.java
raver119: Probably yes. And probably i should just make it package-private. The only use for this thing is supposed to be single-threaded & for consumers state tracking. So, limited basically to 8 independent states.

Issue #3370, opened 2017-05-06
Unit tests currently failing on master


AlexDBlack:
deeplearning4j-core unit tests currently failing on master are shown below.

![image](https://cloud.githubusercontent.com/assets/2360237/25770888/0facb21e-3285-11e7-8ac4-a91a3a329143.png)

See also https://github.com/deeplearning4j/deeplearning4j/issues/3365

AlexDBlack:
Regression tests are now passing here: https://github.com/deeplearning4j/deeplearning4j/pull/3373

AlexDBlack:
All tests listed above are now passing after merging this PR: https://github.com/deeplearning4j/deeplearning4j/pull/3373

Issue #3005, opened 2017-03-09
Review updater design and add support for custom updaters


AlexDBlack:
This would allow users to experiment with things like this: https://github.com/deeplearning4j/nd4j/issues/1692

Some steps required to add this:
- custom updater support in the layer configuration classes
- handle JSON (in NeuralNetConfiguration) for custom updaters
- Modify LayerUpdater updater initialization https://github.com/deeplearning4j/deeplearning4j/blob/1f8af820c29cc5567a2c5eaa290f094c4d1492a7/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/updater/LayerUpdater.java#L254

Edit: Another thing I would like to do is  break the configuration for updaters out into classes.
So, instead of something like ```updater(Updater.ADAM).adamMeanDecay(...)...``` we do either ```.updater(new Adam.builder().meanDecay(...).build())``` or ```.updater(new Adam(mean,var))```
This type of design also gives us block updater support for custom updaters (i.e., updater combining for my recent changes).

My concern here is that with the current design as we add more and more updaters, we end up with a situation where the hyperparameters for all other updaters are:
- Present on all of the builder patterns
- Always present in the JSON representation of the network
even when they are being used.
For example, adding the Eve updater (https://arxiv.org/pdf/1611.01505v1.pdf) will add 4 (or 6) new hyperparameters that will always be present (methods/json) even when using any other updater.

crockpotveggies:
I agree with moving the params to the updater class itself. This is something that often affects me when trying to make the updater configurable in a project.

AlexDBlack:
Implemented here: https://github.com/deeplearning4j/deeplearning4j/pull/3373
Refactoring complete and custom updaters are now supported.

PR #3373, opened 2017-05-06
Updater refactoring

AlexDBlack:
***WIP DO NOT MERGE***
Functionality and new tests are complete. Needs javadoc for new/deprecated methods and some last test fixes (most of which are failing on master too).

New API is pretty self-explanatory. For example,
```
.updater(Updater.ADAM).adamMeanDecay(0.9).adamVarDecay(0.999)
```
Becomes
```
.updater(Adam.builder().beta1(0.9).beta2(0.999).build())
```

More examples: https://github.com/deeplearning4j/dl4j-examples/pull/451

Only thing I don't like: there's 2 ways to set LR...
- Use the ```.learningRate(x)``` method -> takes precidence over second option below, if used
- Use the updaters or builders ```.updater(new Nesterov(lr, momentum))```

https://github.com/deeplearning4j/deeplearning4j/issues/3005
https://github.com/deeplearning4j/nd4j/pull/1819

- Cleanup of configuration - updater-specific configuration is moved:
    - From layer/neuralnetconfiguration builders to (newly added) Nesterovs/Adam/whatever config classes
    - Consequently, we also won't (for example) have Adam-specific code in our JSON when using say Nesterov updater
- Configuration cleanup is in preparation for Eve (and AdaMax?) updaters, to avoid polluting API with many more updater-specific configuration options (as discussed in https://github.com/deeplearning4j/deeplearning4j/issues/3005 )
- Custom updaters are now supported + tested
- Fixes regression tests failing on master (adds backward compatibility for JSON for distributions) - https://github.com/deeplearning4j/deeplearning4j/issues/3370
    - Also adds 0.8.0 regression tests
- Adds AdaMax updater support (thanks to @crockpotveggies )
- Optimization: bias gradient calculation no longer requires a working array + assign op


crockpotveggies:
I agree, there should be a single way to set learning rate. Considering we have the API for setting per-layer LR, perhaps it should be removed from the class? And throw an exception if not present?

Issue #3383, opened 2017-05-09
Feature Request: Listener that evaluates the network during training


ajohnsonz:
#### Issue Description

In the same way that the ScoreIterationListener can periodically print out the cost of the training set during training, it would be helpful to print out the accuracy of predictions on the test set. For example, combine the output of ScoreIterationListener with the output of new Evaluation(n).eval(), every X iterations.

#### Version Information

Not available in 0.8.0

raver119:
Implemented & available at current master. Once snapshots will be rebuilt - it will be available there as well.

PR #3384, opened 2017-05-09
EvaluativeListener

raver119:
This PR adds new special listener that does evaluation during training with specified frequency.

Feature request: https://github.com/deeplearning4j/deeplearning4j/issues/3383


raver119:
Reset tests added as requested

raver119:
TrainingListener change implemented.

However i'm not sure about B change. I don't see other options besides providing callback there, which will get access to Model and Listener itself....

raver119:
Such call back might be actually nice way to go. Just to save models after eval... Many people do that :)

raver119:
Callbacks support added :)

Issue #3385, opened 2017-05-10
Cache for preOutput()


raver119:
In CNN architecture significant amount of time is spent within preOutput() method, which gets called twice: once in FF pass,  and later in BP pass.

We could cache output of this method in host memory during FF pass. We already have concept of "external"  workspace, that holds activations/gradients during iteration, we could introduce concept similar to that: stash. I.e. k/v storage limited to HOST-only memory, so once we store some INDArray there - it gets copied to host memory. Once we need to get that INDArray back - we copy it back from storage.

raver119:
@AlexDBlack

AlexDBlack:
Following up on our gitter conversation:
The whole "do forward pass twice" design was decided some time ago to minimise memory requirements.

In some cases like dense layers, the additional memory requirement is equal to the activation size. However in the case of CNNs and sub sampling layers, also need the im2col arrays, which can be much much larger than the activation arrays by factor of four (for 2x2 conv layers stride 1) to nine (3x3 stride 1) or even more some cases.

If we implement this, I think we should also give users the option to prioritise memory or speed.

Anyway I'll go through this and quantify the exact memory requirements, and exactly what arrays we need to keep in memory each layer type.

AlexDBlack:
DenseLayer and OutputLayer:
- Performance: we save doing ```input.mmul(W).addiRowVector(b);``` that's ```[minibatch,nIn]x[nIn,nOut]``` for the mmul
- Memory: increased by ```[minibatch,nOut]```

Convolution layer:
- Memory:
    - Need to save im2col array: ```[miniBatch, outH, outW, inDepth, kH, kW]```. For same mode, stride 1, (worst case) this equates to ```(kH*kW)*inputSize```. In general for same mode, I think it becomes ```(kH*kW)*inputSize/(sH*sW)``` so for the common setting (k=3,s=1) and (k=3,s=2) we have 9x and 2.25x input size respectively.
    - Need to store pre-out: increases memory by ```[minibatch, outDepth, outH, outW]```
- Performance:
   - We save a costly im2col op
   - We save an mmul op: ```[minibatch*outH*outW, (inDepth*kH*kW)] x [(inDepth*kH*kW), outDepth]``` and an addiRowVector

Subsampling layer:
- looks quite similar to convolution layer case. We save an im2col and a max/avg op, with the same im2col memory cost.


My main concern at this point are the size of those im2col arrays.

raver119:
[...]

PR #3464, opened 2017-05-27
Gradients sharing

raver119:
***WIP; DO NOT MERGE;***

This PR adds option to run ParallelWrapper without averaging, via gradients updates sharing between workers. Following implementations were added:
- direct sharing. Updates are shared "as is", without encoding/decoding. Probably should be default mode for PW
- threshold encoding & topN threshold encoding. only portion of updates are shared. These implementations will be used for distributed training.


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/Model.java
```diff
@@ -53,6 +53,13 @@
      */
     void setListeners(IterationListener... listeners);
 
+    /**
+     * This method ADDS additional IterationListener to existing listeners
+     *
+     * @param listener
+     */
+    void addListener(IterationListener listener);
```
AlexDBlack: Good idea. Minor improvement: we might as well make this ```addListeners(IterationListener...)``` and ```addListeners(Collection<?
 extends IterationListener>)```
That would save users having to chain addListener calls for multiple listeners.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
```diff
@@ -318,6 +319,21 @@ public void setLabels(INDArray... labels) {
...
 
     /**
+     * This method allows you to specificy GradientsAccumulator instance to be used with this model
+     *
+     * PLEASE NOTE: Do not use this method unless you understand how to use GradientsAccumulator & updates sharing.
+     * PLEASE NOTE: Do not use this method on standalone model
+     *
+     * @param accumulator
+     */
+    public void setGradientsAccumulator(GradientsAccumulator accumulator) {
```
AlexDBlack: Random thought with these sorts of methods: should we introduce our own annotations (probably defined in ND4J and used everythere) for these sorts of things?
Similar to what Spark does: https://spark.apache.org/docs/1.5.2/api/java/org/apache/spark/annotation/package-summary.html

It'd let people know, and (if easy/possible) it should automatically add warnings to (html/rendered) javadoc etc.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionLayer.java
```diff
@@ -363,6 +366,15 @@ public INDArray activate(boolean training) {
         applyDropOutIfNecessary(training);
 
         INDArray z = preOutput(training);
+
+        if (training) {
+            try (MemoryWorkspace wsB = Nd4j.getWorkspaceManager()
+                    .getWorkspaceForCurrentThread(ComputationGraph.workspaceExternal).notifyScopeBorrowed()) {
+                //preOutput = z.unsafeDuplication();
+                //Nd4j.getExecutioner().commit();
```
AlexDBlack: What's the idea here? seems like a no-op currently?

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/StochasticGradientDescent.java
```diff
@@ -37,6 +38,7 @@
  * No line search
  * @author Adam Gibson
  */
+@Slf4j
```
AlexDBlack: Not used?

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/BasicGradientsAccumulator.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * This class provides accumulation for gradients for both input (i.e. updates coming from network) and output (comint from one ore more models training at the same time)
+ *
+ * @author raver119@gmail.com
```
AlexDBlack: Extra JavaDoc might be good here: it's basic, but in what sense? Compare to what? How does it work?

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/BasicGradientsAccumulator.java
```diff
@@ -0,0 +1,259 @@
...
+            barrier.await();
+            if (firstOne.get() == Thread.currentThread().getId()) {
+                // one thread just nullifies this array
+                updates.assign(0.0);
+                hasSomething.set(false);
+
+                firstOne.set(-1L);
+            }
+
+            updatesLock.readLock().unlock();
```
AlexDBlack: What happens if anything in this try block throws an exception before .unlock()?
Same thing for other uses of locks/barriers in this class.
It's pretty unlikely, but not impossible.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/MessageHandler.java
```diff
@@ -0,0 +1,29 @@
...
+     */
+    void initialize(GradientsAccumulator accumulator);
+
+    /**
+     * This method does broadcast of given update message across network
+     *
+     * @param updates
+     * @return TRUE if something was sent, FALSE otherwise
+     */
+    boolean broadcastUpdates(INDArray updates);
```
AlexDBlack: If implementations may be using network comm - interface should have error handling? (like IOException etc?)

deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/ParallelWrapper.java
```diff
@@ -62,14 +86,17 @@
...
+    protected Object[] trainerContextArgs;
+
+    @Getter @Setter protected GradientsAccumulator gradientsAccumulator;
 
     private MagicQueue mq;
 
     // log uncaught exceptions
     Thread.UncaughtExceptionHandler handler = new Thread.UncaughtExceptionHandler() {
         public void uncaughtException(Thread th, Throwable ex) {
             log.error("Uncaught exception: " + ex);
```
AlexDBlack: It doesn't really matter and I'm being pedantic here, but generally with logging we should use ```log.error("Uncaught exception", ex);```
https://www.slf4j.org/faq.html#paramException

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/Model.java
raver119: Mmmm, yes, makes sense

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
raver119: All up for that. API gets more and more complex, so any hints for end-user expectations are more then welcome

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionLayer.java
raver119: That's preOutput cache draft, i'm playing with it at this moment.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/BasicGradientsAccumulator.java
raver119: model dies due to jvm crash.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/MessageHandler.java
raver119: Agreed.

deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/ParallelWrapper.java
raver119: Not my code, sorry. But i'm going to revisit this thing during spark integration. I don't like UncaughtException at all.

raver119:
So, due to https://github.com/deeplearning4j/deeplearning4j/issues/3385 i'm rewriting encoder/decoder algorithm and adding new kernels, to make it suit CUDA better. Current implementation is ok for CPU though.

Issue #3442, opened 2017-05-23
Add an 'early termination' (Multi)DataSetIterator


AlexDBlack:
The idea: I have an iterator, with say 10k minibatches, but I want to do something every 1k minibatches, that doesn't fit the listener pattern well.

Basically, it'd do something like:
```
public boolean hasNext() {
    return underlying.hasNext() && minibatchCount < terminationPoint;
}
```
every other method would essentially pass through to the underlying iterator.

Priority: low ("nice to have")

eraly:
Merged after review. Closing.

PR #3502, opened 2017-06-09
Simple early termination iterator

eraly:
https://github.com/deeplearning4j/deeplearning4j/issues/3442

## How was this patch tested?
Unit tests

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationDataSetIterator.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationDataSetIterator.java
@@ -1,0 +1,112 @@
+package org.deeplearning4j.datasets.iterator;
+
+import org.nd4j.linalg.dataset.DataSet;
+import org.nd4j.linalg.dataset.api.DataSetPreProcessor;
+import org.nd4j.linalg.dataset.api.iterator.DataSetIterator;
+
+import java.util.List;
+
+/**
+ * Builds an iterator that terminates once the number of minibatches returned with .next() is equal to a specified number
+ * Note that this essentially restricts the data to this specified number of minibatches.
+ */
+public class EarlyTerminationDataSetIterator implements DataSetIterator {
+
+    private DataSetIterator underlyingIterator;
+    private int terminationPoint;
+    private int minibatchCount = 0;
+
+    /**
+     * Constructor takes the iterator to wrap and the number of minibatches after which the call to hasNext()
+     * will return false
+     * @param underlyingIterator, iterator to wrap
+     * @param terminationPoint, minibatches after which hasNext() will return false
+     */
+    public EarlyTerminationDataSetIterator(DataSetIterator underlyingIterator, int terminationPoint) {
+        this.underlyingIterator = underlyingIterator;
+        this.terminationPoint = terminationPoint;
+    }
+
+    @Override
+    public DataSet next(int num) {
+        minibatchCount += num;
+        return underlyingIterator.next(num);
+    }
+
+    @Override
+    public int totalExamples() {
+        return underlyingIterator.totalExamples();
+    }
+
+    @Override
+    public int inputColumns() {
+        return underlyingIterator.inputColumns();
+    }
+
+    @Override
+    public int totalOutcomes() {
+        return underlyingIterator.totalOutcomes();
+    }
+
+    @Override
+    public boolean resetSupported() {
+        return underlyingIterator.resetSupported();
+    }
+
+    @Override
+    public boolean asyncSupported() {
+        return underlyingIterator.asyncSupported();
+    }
+
+    @Override
+    public void reset() {
+        minibatchCount = 0;
+        underlyingIterator.reset();
+    }
+
+    @Override
+    public int batch() {
+        return underlyingIterator.batch();
+    }
+
+    @Override
+    public int cursor() {
+        return underlyingIterator.cursor();
+    }
+
+    @Override
+    public int numExamples() {
+        return underlyingIterator.numExamples();
+    }
+
+    @Override
+    public void setPreProcessor(DataSetPreProcessor preProcessor) {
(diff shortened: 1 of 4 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationMultiDataSetIterator.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param terminationPoint, minibatches after which hasNext() will return false
+     */
+    public EarlyTerminationMultiDataSetIterator(MultiDataSetIterator underlyingIterator, int terminationPoint) {
+        if (terminationPoint <= 0 ) throw new IllegalArgumentException("Termination point (the number of calls to .next() or .next(num)) has to be > 0");
+        this.underlyingIterator = underlyingIterator;
+        this.terminationPoint = terminationPoint;
+    }
+
+    @Override
+    public MultiDataSet next(int num) {
```
huitseeker: if minibatchCount > (terminationPoint - num), before `next(num)` this call with succeed, returning a total # of elements strictly greater than terminationPoint

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationDataSetIterator.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public EarlyTerminationDataSetIterator(DataSetIterator underlyingIterator, int terminationPoint) {
+        if (terminationPoint <= 0 ) throw new IllegalArgumentException("Termination point (the number of calls to .next() or .next(num)) has to be > 0");
+        this.underlyingIterator = underlyingIterator;
+        this.terminationPoint = terminationPoint;
+    }
+
+    @Override
+    public DataSet next(int num) {
+        if (minibatchCount < terminationPoint) {
```
huitseeker: if minibatchCount > (terminationPoint - num), before next(num) this call with succeed, returning a total # of elements strictly greater than terminationPoint

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationDataSetIterator.java
```diff
@@ -0,0 +1,124 @@
...
+        return underlyingIterator.getLabels();
+    }
+
+    @Override
+    public boolean hasNext() {
+        return underlyingIterator.hasNext() && minibatchCount < terminationPoint;
+    }
+
+    @Override
+    public DataSet next() {
```
huitseeker: This call (and the associated one with `next(num)` does not seem to be thread-safe. Should we sync on the miniBatchCount, or make it Atomic ? Or is it better to deal with this by adding javadoc on how this class is not thread-safe ?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationMultiDataSetIterator.java
```diff
@@ -0,0 +1,87 @@
...
+        underlyingIterator.reset();
+    }
+
+    @Override
+    public boolean hasNext() {
+        return underlyingIterator.hasNext() && minibatchCount < terminationPoint;
+    }
+
+    @Override
+    public MultiDataSet next() {
```
huitseeker: This call (and the associated one with `next(num)` does not seem to be thread-safe. Should we sync on the miniBatchCount, or make it Atomic ? Or is it better to deal with this by adding javadoc on how this class is not thread-safe ?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationDataSetIterator.java
raver119: All our iterator implementations are not thread safe, and they are never used in concurrent way.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationDataSetIterator.java
eraly: Termination point is the number of minibatches.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationMultiDataSetIterator.java
eraly: The termination point is not the number of elements. It is the number of minibatches. I have made that clear in the javadoc. 
https://github.com/deeplearning4j/deeplearning4j/pull/3502/files#diff-b986a13d8f51bf0dae2466fc7faa4d52R11

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationMultiDataSetIterator.java
huitseeker: Answered by raver

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/EarlyTerminationDataSetIterator.java
huitseeker: Ah, ok, sory for the overlook. In that case, how about changing the error message to mention the "budgeted # of minibatches"  explicitly ?

eraly: Done.

eraly:
Thanks for the review @huitseeker Merging.

PR #3514, opened 2017-06-12
Added ShiftVertex

bnsh:

This patch will add ShiftVertex, something sort of complementary to ScaleVertex. It will _add_ a constant, rather than multiplying a constant.

## How was this patch tested?

I tested the derivatives manually, and have a test program that tested all the ActivationFunctions (except RReLU, because that's random, and my manual code didn't account for non-deterministic functions.) The gist below is that program.
https://gist.github.com/bnsh/a0458d28e79aa3f159cec43ebeb8b177

## Quick checklist

### DIFF
--- a/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/conf/graph/ShiftVertexTest.java
+++ b/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/conf/graph/ShiftVertexTest.java
@@ -1,0 +1,248 @@
+package org.deeplearning4j.nn.conf.graph;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.apache.commons.math3.util.Pair;
+import org.deeplearning4j.nn.api.OptimizationAlgorithm;
+import org.deeplearning4j.nn.conf.ComputationGraphConfiguration;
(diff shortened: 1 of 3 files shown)


bnsh:
Unfortunately, I don't have much experience in writing real unit tests in Java. Can you look at my gist, and see if they would be good tests? If so, I will try to adapt that to the tests that I see in the repository.

agibsonccc:
My advice here would be to check for the basics: like 0 numParams,testing the various feedforward methods for shape sizes and the like..

Copy this: https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/feedforward/dense/DenseTest.java

Put your test in deeplearning4j-core.

Thanks for your time!

bnsh:
OK. I just added a JUnit Test. It tests that

 * numParams(true) returns 0
 * numParams(false) returns 0
 * a simple test which makes a graph and verifies that what comes out of the graph is _indeed_ shifted by a known value.
 * a comprehensive test which makes a graph with an activation function in front and back, and checks that the derivatives and the score are both what I manually calculate.

agibsonccc:
Thanks! Merged.

PR #3518, opened 2017-06-13
Distributed Spark

raver119:
***WIP; DO NOT MERGE;***

This PR adds distributed training over the network functionality, without parameters averaging.

Basic idea highlights: 
- Spark is used only to feed data, training/communication happens in our codebase
- Training is handled via ParallelWrapper instance
- Weight updates are travelling over the network using Aeron
- Epochs are handled without re-sending original model via broadcasts
- Spark Master always has up-to-date weights via "silent worker" approach
- Hypothetically different encoding schemes might be added later, current implementation is threshold-based encoding
- Updaters states will be delivered at last epoch via hidden hooks to eval or model saving


deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/CudaGradientsAccumulator.java
```diff
@@ -107,16 +159,26 @@ public void applyUpdate(StepFunction function, INDArray params, INDArray updates
...
-        function.step(params, updates);
+        if (externalSource != null)
+            while (!externalSource.isEmpty()){
+                INDArray compressed = externalSource.poll();
+
+                INDArray decoded = Nd4j.getExecutioner().thresholdDecode(compressed, updates);
+                cnt++;
+            }
+
+        // TODO: average updates probably?
```
huitseeker: OK, I'm not sure I understand everything here, so bear with me : is there or isn't there an averaging in the step function ?

At the end of the day, IIUC your accumulation is large / unlimited, and based on that alone it sound like it might diverge. If all m executors (thread, what have you) starts from the same vector wold, then after processing a large batch, the new vector on each executor will be close to the optimal solution w*.

If the variable is updated by vnew = vold + Σ_{i=1}^m ∆(S_i) ,
then we have
wnew − w* =  wold - w* + + Σ_{i=1}^m ∆(S_i) ≈  wold - w* + + Σ_{i=1}^m ∆(w* - wold) = (m-1) (w* - wold) 

Which diverges if m >=3.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/pom.xml
```diff
@@ -38,4 +53,17 @@
...
     </profiles>
 
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
+                    <target>1.8</target>
```
raver119: Aeron assumes Java 8

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/pw/SharedTrainingWrapper.java
```diff
@@ -0,0 +1,237 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Observer;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * This class maintains ParallelWrapper instance in Spark environment, and provides primitives for inter-executor
+ * communication during training over partitions.
```
raver119: That's already happened, right there. SilentTrainerDriver provides connectivity with Aeron, and gradients are passed into GradientsAccumulator

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/CudaGradientsAccumulator.java
raver119: I'm totally, literally - totally - agree with you. Averaging is needed for performance reasons as well. I'm half way there.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/functions/SharedFlatMapDataSet.java
```diff
@@ -0,0 +1,54 @@
...
+    public Iterable<R> call(Iterator<DataSet> dataSetIterator) throws Exception {
+        /*
+            That's the place where we do our stuff. Here's the plan:
+            1) we pass given iterator to VirtualDataSetIterator, which acts as holder for them
+            2) Virtual iterator will provide load balancing between available devices
+            3) we'll lock out here
+         */
+
+        // iterator should be silently attached to VirtualDataSetIterator, and used appropriately
+        SharedTrainingWrapper.getInstance().attachDS(dataSetIterator);
```
raver119: In this particular case it's datasets, so they are wrapped into VirtualDataSetIterator, and those original Iterator&lt;DataSet&gt; are consumed one by one, releasing corresponding executors block

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/functions/SharedFlatMapDataSet.java
```diff
@@ -0,0 +1,54 @@
...
+    public SharedFlatMapDataSetAdapter(TrainingWorker<R> worker) {
+        // we're not going to have anything but Shared classes here ever
+        this.worker = (SharedTrainingWorker) worker;
+    }
+
+    @Override
+    public Iterable<R> call(Iterator<DataSet> dataSetIterator) throws Exception {
+        /*
+            That's the place where we do our stuff. Here's the plan:
+            1) we pass given iterator to VirtualDataSetIterator, which acts as holder for them
```
raver119: Yep, once dust settles i'll make final javadoc/comments pass sure. I try to do that every time :)

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/functions/SharedFlatMapMultiPDS.java
```diff
@@ -0,0 +1,69 @@
...
+
+    protected final SharedTrainingWorker worker;
+    protected final PortableDataStreamMDSCallback callback;
+
+    public SharedFlatMapMultiPDSAdapter(TrainingWorker<R> worker) {
+        this(worker, null);
+    }
+
+    public SharedFlatMapMultiPDSAdapter(TrainingWorker<R> worker, PortableDataStreamMDSCallback callback) {
+        // we're not going to have anything but Shared classes here ever
```
raver119: Yep, won't hurt

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/pw/SharedTrainingWrapper.java
raver119: That part was implemented earlier, for standalone ParallelWrapper with GradientsAccumulator

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/CudaGradientsAccumulator.java
raver119: But, i'm not quite sure why it should diverge, if all models will be applying the same updates derived from original wold state.

deeplearning4j-scaleout/spark/dl4j-spark/src/main/java/org/deeplearning4j/spark/impl/paramavg/BaseTrainingMaster.java
```diff
@@ -0,0 +1,247 @@
...
+
+    protected static ObjectMapper getNewMapper(JsonFactory jsonFactory) {
+        ObjectMapper om = new ObjectMapper(jsonFactory);
+        om.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        om.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+        om.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
+        om.enable(SerializationFeature.INDENT_OUTPUT);
+        om.setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE);
+        om.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+        return om;
```
AlexDBlack: I realize this is just copied over from the original impl - but these JSON mapper get/config methods - I'm thinking they are probably reusable elsewhere... maybe move them to a util class so we can use them elsewhere as needed?

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
```diff
@@ -0,0 +1,1044 @@
...
+    protected int rddDataSetNumExamples;
+    protected long debugLongerIterations = 0L;
+
+    // TODO: this option should be abstracted, if we decide to generalize this trainingmaster
+    protected double threshold;
+
+    protected Repartition repartition;
+    protected RepartitionStrategy repartitionStrategy;
+
+    protected ParameterAveragingTrainingMasterStats.ParameterAveragingTrainingMasterStatsHelper stats;
```
AlexDBlack: Maybe we pull out the helper into its own class?

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
```diff
@@ -0,0 +1,1044 @@
...
+        network.getNetwork().init();
+
+        // this instance will be SilentWorker - it'll accept and apply messages, but won't contribute to training. And we init it only once
+        if (isFirstRun.compareAndSet(false, true)) {
+            trainingDriver = new SilentTrainingDriver(network.getNetwork().params(), network.getNetwork().getOptimizer().getStepFunction());
+            VoidParameterServer.getInstance().init(voidConfiguration, transport, trainingDriver);
+        }
+    }
+
+    protected void prepareNetworkAndStuff(SparkDl4jMultiLayer network) {
```
AlexDBlack: Code duplication - very minimal difference between this and the previous method... Maybe use a method with 2 args, a SparkDl4jMultiLayer and SparkComputationGraph, with one always null?

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
```diff
@@ -0,0 +1,1044 @@
...
+        log.info("Setting controller address to {}:{}", voidConfiguration.getControllerAddress(), voidConfiguration.getUnicastPort());
+        voidConfiguration.setShardAddresses(voidConfiguration.getControllerAddress());
+        voidConfiguration.setNumberOfShards(1);
+
+        Transport transport = voidConfiguration.getTransportType() == TransportType.ROUTED ? new RoutedTransport() : voidConfiguration.getTransportType() == TransportType.BROADCAST ? new MulticastTransport() : this.transport;
+
+        if (transport == null)
+            throw new DL4JInvalidConfigException("No Transport implementation was defined for this training session!");
+
+        // TODO: Alex, any better ideas here? Plus, if we're not on SGD updater, we might want to eventually dump updater to Master?
```
AlexDBlack: > we might want to eventually dump updater to Master?

As in: no updater on each worker, just post raw gradients and update in master?

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
```diff
@@ -0,0 +1,1044 @@
...
+
+import java.io.IOException;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author raver119@gmail.com
+ */
+@Slf4j
+public class SharedTrainingMaster extends BaseTrainingMaster<SharedTrainingResult, SharedTrainingWorker> implements TrainingMaster<SharedTrainingResult, SharedTrainingWorker> {
```
AlexDBlack: General comment here: I'm not 100% clear or the splitting/repartitioning - though it looks much the same as the parameter averaging training master?
If so: Because of the different approach to training here, we may want to rethink whether that design is optimal or not. Though with the whole "export DataSets and pass around strings" approach, it probably doesn't make too much difference how we do the splitting, now that I think about it...

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingWorker.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public void removeHook(TrainingHook trainingHook) {
+
+    }
+
+    @Override
+    public void addHook(TrainingHook trainingHook) {
+
```
AlexDBlack: TrainingHook doesn't *really* fit the design here, though do we want to throw an UnsupportedOperationException? I'm just worried about people trying to use it and not knowing it's a no-op here.

deeplearning4j-scaleout/spark/dl4j-spark/src/main/java/org/deeplearning4j/spark/impl/paramavg/ParameterAveragingTrainingWorker.java
```diff
@@ -40,7 +40,7 @@
  *
  * @author Alex Black
  */
-public class ParameterAveragingTrainingWorker implements TrainingWorker<ParameterAveragingTrainingResult> {
+public class ParameterAveragingTrainingWorker extends BaseTrainingWorker<ParameterAveragingTrainingResult> implements TrainingWorker<ParameterAveragingTrainingResult> {
```
AlexDBlack: I think the ```implements``` after the ```extends``` is redundant here maybe?

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/BasicGradientsAccumulator.java
```diff
@@ -264,4 +264,20 @@ public void reset() {
...
     }
+
+    /**
+     *
+     * @param worker
+     * @return
+     */
+    @Override
+    public int getFreeSpace(int worker) {
+        return Integer.MAX_VALUE;
```
raver119: Nah, bad design idea, happened before that fancy queue idea discussed with you.  Method will be removed.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/FancyBlockingQueue.java
```diff
@@ -0,0 +1,288 @@
...
+/**
+ * This BlockingQueue implementation is suited only for symmetric gradients updates, and should NOT be used anywhere else.
+ *
+ * Basic idea: all worker threads requesting via poll()/take() method will be advancing only once all consumers get the same element from Queue.
+ * So, multiple consumers are guaranteed to be consuming the same elements in the same order served by this queue.
+ *
+ * @author raver119@gmail.com
+ */
+@Slf4j
+public class FancyBlockingQueue<E> implements BlockingQueue<E>, Registerable {
```
raver119: I have strong feeling that we really want that. I can't prove that mathematically yet, but async training/sync updates give us some sort of guarantees that gradients are fresh enough, and we're not facing stale gradients.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/FancyBlockingQueue.java
```diff
@@ -0,0 +1,288 @@
...
+        this.currentConsumers.set(consumers);
+
+        lock.writeLock().unlock();
+    }
+
+    @Override
+    public void put(E e) throws InterruptedException {
+        lock.readLock().lock();
+        backingQueue.put(e);
+        lock.readLock().unlock();
```
raver119: That's not for queue. That's to ensure that all local consumers will have exactly the same amount of external gradients applied, since queue advances for everyone at the same time. CUDA-style parallelism.

And read locks are cheap :)

deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/trainer/DefaultTrainer.java
```diff
@@ -56,7 +56,7 @@
     @Builder.Default protected volatile boolean useMDS = false;
     protected final String uuid = UUID.randomUUID().toString();
     @Builder.Default protected boolean onRootModel = false;
-    @Builder.Default protected AtomicLong lastEtlTime = new AtomicLong(0);
+    @Builder.Default protected volatile AtomicLong lastEtlTime = new AtomicLong(0);
```
raver119: Ye, that was me looking for some race condition in standalone PW, and just tried to be sure that's not the case.

deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/trainer/DefaultTrainer.java
```diff
@@ -165,6 +165,8 @@ protected void setupIfNeccessary() {
             shouldUpdate = new AtomicBoolean(false);
         if (isStopped == null)
             isStopped = new AtomicBoolean(false);
+        if (lastEtlTime == null)
```
raver119: Unfortunately no. That volatile change in comment above was exactly about this field being null.  Lombok issue in other words.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/networking/SilentTrainingDriver.java
```diff
@@ -0,0 +1,207 @@
...
+                //log.info("Resending message, skipping {}", message.getOriginatorId());
+                transport.sendMessageToAllClients(message, message.getOriginatorId(), transport.getOwnOriginatorId());
+            } // else log.info("No known Clients so far");
+        } else
+            throw new DL4JInvalidConfigException("Neither GradientsAccumulator or StepFunction is defined!");
+    }
+
+    @Override
+    public void pickTraining(SilentUpdatesMessage message) {
+        // this message won't be ever called
```
raver119: Nobody can call it, and get that exception.

TrainingDriver methods are invoked only from messages. So no chance to use wrong thing here. But ok, will add :)

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/networking/SilentTrainingDriver.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    @Override
+    public void addCompletionHook(long originatorId, long frameId, long messageId) {
+        // no-op
+    }
+
+    @Override
+    public String targetMessageClass() {
+        return SilentUpdatesMessage.class.getSimpleName();
```
raver119: That's for provider. Simple name works fine, since provide does mapping. However, not used too.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/pw/SharedTrainingWrapper.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    public SharedTrainingResult run(SharedTrainingWorker worker) {
+        /*
+            first call instantiates pw, messenger etc, and gets in charge here.
+         */
+        if (isFirst.compareAndSet(false, true)) {
+            // getting model from worker, and instantiating PW
+            try {
+                Thread.sleep(5000);
```
raver119: Nah, you're right, that shouldn't get into master. I'll change mechanics from "First one runs PW" to "Last one runs PW"

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/pw/SharedTrainingWrapper.java
```diff
@@ -0,0 +1,340 @@
...
+            } else {
+                // if wrapper is null, we're fitting standalone model then
+                if (iteratorDS != null) {
+                    if (model instanceof ComputationGraph) {
+                        ((ComputationGraph) model).fit(iteratorDS);
+                    } else if (model instanceof MultiLayerNetwork) {
+                        ((MultiLayerNetwork) model).fit(iteratorDS);
+                    }
+                } else if (iteratorMDS != null) {
+                    ((ComputationGraph) model).fit(iteratorMDS);
```
raver119: And maybe we'll finally get single DataSet interface/impl? :)

deeplearning4j-scaleout/spark/dl4j-spark/src/main/java/org/deeplearning4j/spark/impl/paramavg/BaseTrainingMaster.java
raver119: That's exactly what i did here - both Averaging and Sharing use single impl here :)

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
raver119: Right, extended/real tests will show what fits best here. I've kept repartitioning alive after your concerns about partitions equality.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
raver119: Nah that's what we've already discussed and what i'm doing right now: passing back/averaging  updaters states.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingWorker.java
raver119: Okay.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
```diff
@@ -76,8 +76,8 @@ public BaseLayer(NeuralNetConfiguration conf) {
     }
 
     public BaseLayer(NeuralNetConfiguration conf, INDArray input) {
-        this(conf);
         this.input = input;
+        this.conf = conf;
```
raver119: Ugh. Bad merge from master. Should be fixed.

raver119: Thanks! Missed it somehow during conflicts resolution... I've backported preOutput cache to master, and then merge conflict appeared.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
raver119: Yep, good idea.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
raver119: Sure.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/FancyBlockingQueue.java
raver119: Plus there's one more thing. One of possible further improvements there, will be local averaging of updates before sharing them over network. To make that possible - updates should be symmetrical.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/iterators/MultiPdsIterator.java
```diff
@@ -0,0 +1,44 @@
...
+import org.nd4j.linalg.dataset.DataSet;
+import org.nd4j.linalg.dataset.api.MultiDataSet;
+
+import java.util.Iterator;
+import java.util.function.Consumer;
+
+/**
+ * @author raver119@gmail.com
+ */
+public class MultiPdsIterator implements Iterator<MultiDataSet> {
```
raver119: There's small, but significant difference. Your implementation assumes that PDS might only hold MultiDataSets inside, in serialized form. And i want to get an option to pass data from PDS to DataVec, for further vectorization.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
raver119: And yes, as i see - we want as close data amount per executor (not worker!) as possible.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/CudaGradientsAccumulator.java
```diff
@@ -93,7 +112,83 @@ protected CudaGradientsAccumulator(int parties, @NonNull MessageHandler handler,
...
+    }
+
+    @Override
+    public void registerConsumers(int numConsumers) {
+        // we don't want double spending here
+        if (registered.get()) {
+            if (isDebug)
+                log.info("Master thread locks at RC");
+
+            while (registered.get()) {
```
raver119: Good point. Fixed.

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/iterators/VirtualIterator.java
```diff
@@ -0,0 +1,63 @@
...
+        if (!state.get()) {
+            this.setChanged();
+            notifyObservers();
+        }
+        return u;
+    }
+
+    @Override
+    public E next() {
+        return iterator.next();
```
raver119: Mmmm, what for? We're strictly following pattern here: next() is always following hasNext(), without any exclusions.

raver119: P.s. and if someone breaks it (who?), he'll get NoSuchElementException from underlying iterator.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/FancyBlockingQueue.java
huitseeker: Why not sync updates from a configurable m out of n requirement (m < n, a.k.a "the first m to answer") ? That's the aproach of https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/45187.pdf

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/NeuralNetwork.java
```diff
@@ -0,0 +1,84 @@
...
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.dataset.api.DataSet;
+import org.nd4j.linalg.dataset.api.MultiDataSet;
+import org.nd4j.linalg.dataset.api.iterator.DataSetIterator;
+import org.nd4j.linalg.dataset.api.iterator.MultiDataSetIterator;
+
+/**
+ * @author raver119
+ */
+public interface NeuralNetwork {
```
AlexDBlack: I like this 👍

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/MultiLayerConfiguration.java
```diff
@@ -561,6 +561,7 @@ public MultiLayerConfiguration build() {
             conf.inferenceWorkspaceMode = inferenceWorkspaceMode;
             conf.cacheMode = cacheMode;
 
+            log.info("Setting seed: {}", conf.getConf(0).getSeed());
```
AlexDBlack: Probably don't want this in final merge to master.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
```diff
@@ -2751,6 +2762,26 @@ public Evaluation evaluate(DataSetIterator iterator, List<String> labelsList) {
...
     }
 
+    @Override
+    public INDArray updaterState() {
+        return getUpdater() != null ? getUpdater().getStateViewArray() : null;
+    }
+
+    @Override
+    public void fit(MultiDataSet dataSet) {
+        throw new DL4JInvalidInputException("MultiLayerNetwork can't handle MultiDataSet. Please consider use of ComputationGraph");
```
AlexDBlack: Not a big deal, but we *could* if we want check for single feature array + single label array, and convert to DataSet. And wrap the MultiDataSetIteartor in a DataSetIterator
(I don't mind doing this post merge of this; we can open an issue for that unless you want to tackle it now)

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/FancyBlockingQueue.java
AlexDBlack: This is definitely something to come back to and profile - but I'd be ok with the current implementation for now.
In theory, it shouldn't be too bad, as this only impacts the computations within a single node (like multi-GPU). The only exception would be really imbalanced time series.

We'd need to think carefully about the implications of something like what @huitseeker is proposing with respect to what we're implementing here. But again, let's revisit this after profiling.

deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/trainer/DefaultTrainer.java
```diff
@@ -187,12 +189,19 @@ public void shutdown() {
...
-        if (replicatedModel instanceof MultiLayerNetwork) {
-            ((MultiLayerNetwork) replicatedModel).setLastEtlTime(lastEtlTime.get());
-            ((MultiLayerNetwork) replicatedModel).fit(dataSet);
-        } else if (replicatedModel instanceof ComputationGraph) {
-            ((ComputationGraph) replicatedModel).setLastEtlTime(lastEtlTime.get());
-            ((ComputationGraph) replicatedModel).fit(dataSet);
+        try {
+            if (replicatedModel instanceof MultiLayerNetwork) {
+                if (lastEtlTime == null)
+                    lastEtlTime = new AtomicLong(0);
```
AlexDBlack: Different handling for lastEtlTime == null for ComputationGraph and MultiLayerNetwork?

deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/trainer/DefaultTrainer.java
```diff
@@ -187,12 +189,19 @@ public void shutdown() {
...
+
+                ((MultiLayerNetwork) replicatedModel).setLastEtlTime(lastEtlTime.get());
+                ((MultiLayerNetwork) replicatedModel).fit(dataSet);
+            } else if (replicatedModel instanceof ComputationGraph) {
+                ((ComputationGraph) replicatedModel).setLastEtlTime(lastEtlTime.get());
+                ((ComputationGraph) replicatedModel).fit(dataSet);
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
AlexDBlack: Is it safe to just print this exception and move on?

deeplearning4j-scaleout/spark/dl4j-spark-parameterserver/src/main/java/org/deeplearning4j/spark/parameterserver/training/SharedTrainingMaster.java
AlexDBlack: Yup, fair enough. Another thing to profile/test once merged.

PR #3521, opened 2017-06-13
Build Optimizations

huitseeker:
- Silence Eclipse on import of projects containing lint-maven-plugin
- tag jars with git-commit-id plugin

Builds!

### DIFF
--- a/change-spark-versions.sh
+++ b/change-spark-versions.sh
@@ -23,7 +23,7 @@ set -e
 
 VALID_VERSIONS=( 1 2 )
 SPARK_2_VERSION="2\.1\.0"
-SPARK_1_VERSION="1\.6\.2"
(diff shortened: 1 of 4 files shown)


raver119:
Btw, can we have automatically inject ND4j/DL4j versions somewhere? It would be nice to add that information to Nd4jEnvironmentInformation for debugging purposes...

Plus automated check for versions equality...

huitseeker:
Yes, but that is unrelated to `git-commit-id`: it's something that has always been allowed by maven-resources-plugin, which can generate a `.properties` file at source preparation, and you can query these resources from a POJO afterwards. Could you open an issue ?

huitseeker:
@saudet yep! added as part of https://github.com/deeplearning4j/DataVec/pull/306

PR #3533, opened 2017-06-16
New IterationListener that display the remaining time and the end of …

pierrelevy:
This listener displays into INFO logs the remaining time in minutes and the date of the end of the process. 

Here is a preview of the output with the Score listener : 
`22:57:56.673 [main] INFO  o.d.o.l.ScoreIterationListener - Score at iteration 5317 is 0.7158182582504605

22:57:56.673 [main] INFO  o.d.o.l.TimeIterationListener - Remaining time : 457mn - End expected : Sat Jun 17 06:35:11 CEST 2017
`

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/listeners/TimeIterationListener.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/listeners/TimeIterationListener.java
@@ -25,6 +25,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Date;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * Time Iteration Listener.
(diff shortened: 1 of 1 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/listeners/TimeIterationListener.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public void invoke()
+    {
+        invoked = true;
+    }
+
+    @Override
+    public void iterationDone( Model model, int iteration)
+    {
+        int currentIteration = iteration + 1;
```
raver119: This thing. In ParallelWrapper environment, this Listener instance will be shared between multiple workers, each of them will have own iteration number. So we just need to use AtomicLong counter and increment it, so this Listener will be also perfectly compatible with PW

pierrelevy: No problem

raver119:
Thank you very much!

Issue #3469, opened 2017-05-28
Invalid handling of Conv Layer in SubsetVertex


woru:
SubsetVertex doesn't work for CNN input type.

There's an error in 
https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/graph/SubsetVertex.java line 104.

It should be
```java
return InputType.convolutional(conv.getWidth(), conv.getHeight(), from - to + 1);
```
but right now it's:

```java
 return InputType.convolutional(from - to + 1, conv.getWidth(), conv.getHeight());
```

AlexDBlack:
Yes, you are (almost) correct:
https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/graph/SubsetVertex.java#L37
>In the case of convolutional (4d) activations, this is done along depth.

The relevant method signature:
> public static InputType convolutional(int height, int width, int depth) {

The corrected version should be:
```
return InputType.convolutional(conv.getHeight(), conv.getWidth(), from - to + 1);
```

Feel free to open a pull request, or I'll push that fix up later.

AlexDBlack:
Was fixed here: https://github.com/deeplearning4j/deeplearning4j/pull/3555

PR #3555, opened 2017-06-21
Fixed parameters order

woru:

Just a fix.

## How was this patch tested?

I run my nn with this change and it worked. I'll add a test case when I have time.


woru:
For issue #3469

PR #3783, opened 2017-08-04
Knn distances

emsixteeen:

Adding a distances field to Knn Results

## How was this patch tested?

Verified existing tests pass

### DIFF
--- a/deeplearning4j-nearestneighbors-parent/deeplearning4j-nearestneighbor-server/src/main/java/org/deeplearning4j/nearestneighbor/server/NearestNeighbor.java
+++ b/deeplearning4j-nearestneighbors-parent/deeplearning4j-nearestneighbor-server/src/main/java/org/deeplearning4j/nearestneighbor/server/NearestNeighbor.java
@@ -31,7 +31,7 @@ public class NearestNeighbor {
 
             if (add.size() != distances.size()) {
                 throw new IllegalStateException(
-                        String.format("add.size(%d) != distances.size(%d)",
+                        String.format("add.size == %d != %d == distances.size",
                                 add.size(), distances.size()));
             }
 
--- a/deeplearning4j-nearestneighbors-parent/deeplearning4j-nearestneighbor-server/src/main/java/org/deeplearning4j/nearestneighbor/server/NearestNeighbor.java
+++ b/deeplearning4j-nearestneighbors-parent/deeplearning4j-nearestneighbor-server/src/main/java/org/deeplearning4j/nearestneighbor/server/NearestNeighbor.java
@@ -28,8 +28,15 @@ public class NearestNeighbor {
             List<DataPoint> add = new ArrayList<>();
             List<Double> distances = new ArrayList<>();
             tree.search(input, record.getK(), add, distances);
-            for (DataPoint dataPoint : add) {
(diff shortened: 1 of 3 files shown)


deeplearning4j-nearestneighbors-parent/deeplearning4j-nearestneighbor-server/src/main/java/org/deeplearning4j/nearestneighbor/server/NearestNeighbor.java
```diff
@@ -28,8 +28,15 @@
             List<DataPoint> add = new ArrayList<>();
             List<Double> distances = new ArrayList<>();
             tree.search(input, record.getK(), add, distances);
-            for (DataPoint dataPoint : add) {
-                results.add(new NearestNeighborsResult(dataPoint.getIndex()));
+
+            if (add.size() != distances.size()) {
+                throw new IllegalStateException(
+                        String.format("add.size(%d) != distances.size(%d)",
```
huitseeker: While I applaud the use of format, this syntax may lead readers to think the value of `add.size()` is the argument to size. 

How about "add.size == %d != %d == distances.size" ?

deeplearning4j-nearestneighbors-parent/deeplearning4j-nearestneighbors-model/src/main/java/org/deeplearning4j/nearestneighbor/model/NearestNeighborsResult.java
```diff
@@ -12,5 +11,5 @@
 @NoArgsConstructor
 public class NearestNeighborsResult {
```
huitseeker: So, as a hack that would work, except as you know, we also have a `NearestNeighborResults` class for a batch of those. Wouldn't it be more logical to have the distance added in `NearestNeighborResults` as a field ? We'd have the distance added in at the list's creation, and would just have `NearestNeighborResult` be a pass-through construction that just unfolds the tuple a list element contains.

emsixteeen: I imagine the same question can be asked with regard to `index` as well. Truth be told: I'm just patching for `distance` here, in line with the way I found it. I guess a discussion can be had about refactoring it, but I think that would include both `index` as well as `distance` (e.g. the whole class). For now though, I think this fits into / doesn't alter the original design & "spirit" of this object.

deeplearning4j-nearestneighbors-parent/deeplearning4j-nearestneighbor-server/src/main/java/org/deeplearning4j/nearestneighbor/server/NearestNeighbor.java
emsixteeen: changed / fixed

huitseeker:
OK to yield to @emsixteeen 's judgement here. If [this thread](https://github.com/deeplearning4j/deeplearning4j/pull/3783#discussion_r131470526) gets ok from @agibsonccc (who might see a quick way to do distances in the reponse that I don't), good to merge.

PR #3795, opened 2017-08-05
Fix ConcurrentModificationException in MathUtils.partitionVariable

joebowbeer:
Proposed fix:

Refactor MathUtils.partitionVariable to use iterator.remove().

Otherwise, the implicit ArrayList iterator in the for loop may throw a ConcurrentModificationException

I encountered the exception in PlayUIServer and then searched for similar bugs. This is the only other occurrence I found.


joebowbeer:
The ConcurrentModificationException is a misnomer, BTW. It is thrown when the for-loop iterator detects that the underlying list has been modified (e.g., by explicit list.remove call).

The public partitionVariable method in MathUtils in unused? Where is it duplicated? (My eyes deceive me.)

huitseeker:
@joebowbeer I gave you the duplication location above : https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/util/MathUtils.java#L606

joebowbeer:
Thanks for clarification.

I created Issue #3797 to capture this information.

huitseeker:
Closing in favor of 3811

PR #3799, opened 2017-08-05
Introduce SKIP_BUILD mode to perform-release.sh script

saudet:

Allow updating the source code for a release before making the final build on the "latest_release" tag.

## How was this patch tested?

Build succeeds with source code modified by the script, with no reference to SNAPSHOT artifacts.

### DIFF
--- a/perform-release.sh
+++ b/perform-release.sh
@@ -9,9 +9,7 @@ fi
 RELEASE_VERSION=$1
 SNAPSHOT_VERSION=$2
 STAGING_REPOSITORY=$3
-if [[ -z ${SKIP_BUILD:-} ]]; then
-    SKIP_BUILD=0
-fi
+SKIP_BUILD=${SKIP_BUILD:-0}
 
 echo "Releasing version $RELEASE_VERSION ($SNAPSHOT_VERSION) to repository $STAGING_REPOSITORY"
 echo "========================================================================================"
--- a/perform-release.sh
+++ b/perform-release.sh
@@ -43,9 +43,9 @@ if [[ "${SKIP_BUILD}" == "0" ]]; then
     source change-scala-versions.sh 2.11
     source change-cuda-versions.sh 8.0
 fi
-git commit -a -m "Update to version $RELEASE_VERSION"
-git tag -a -m "deeplearning4j-$RELEASE_VERSION" "deeplearning4j-$RELEASE_VERSION"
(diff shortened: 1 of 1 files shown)


perform-release.sh
```diff
@@ -9,6 +9,9 @@ fi
 RELEASE_VERSION=$1
 SNAPSHOT_VERSION=$2
 STAGING_REPOSITORY=$3
+if [[ -z ${SKIP_BUILD:-} ]]; then
```
huitseeker: Nitpick : `SKIP_BUILD=${SKIP_BUILD:-0}`

perform-release.sh
```diff
@@ -28,20 +31,23 @@ for f in $(find . -name 'pom.xml' -not -path '*target*'); do
...
-source change-cuda-versions.sh 8.0
-mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -DstagingRepositoryId=$STAGING_REPOSITORY
-source change-spark-versions.sh 2
-mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -DstagingRepositoryId=$STAGING_REPOSITORY -Dspark.major.version=2
-
-source change-spark-versions.sh 1
-source change-scala-versions.sh 2.11
-source change-cuda-versions.sh 8.0
+if [[ "${SKIP_BUILD}" == "0" ]]; then
+    source change-scala-versions.sh 2.10
```
huitseeker: This assumes the current spark version is 1 (which may or may not be true) and does not build for 7.5/2.11 or 8.0/2.10

This should really call buildmultiplescalaversions.sh The script will happily use all these arguments, and then we only have one place in which we have to maintain the versions-switching.

saudet: There are no modules that use both CUDA and Spark, so that's not an issue for that.

I've replied for all the rest at https://github.com/deeplearning4j/Arbiter/pull/93#pullrequestreview-54745066

huitseeker: Yep, but if I use cuda 7.5, where can I find artifacts for 2.11 ?

saudet: There are no artifacts that dependent on both CUDA and Spark.

huitseeker: You mean CUDA and Scala. OK. Fair enough. As I said elsewhere, better to table the logic issue for now.

PR #3837, opened 2017-08-10
Add optional labels for KNN results returned by `/knnnew` endpoint.

turambar:

We need a convenient and reliable way to "identify" the neighbors returned by a KNN search (besides row indeces, that is). To facilitate this, we're making two small changes to KNN functionality:

- add a String field named`label` to the `NearestNeighborResult` class
- add an optional commandline arg to the `NearestNeighborServer` that allows us to pass in one or more text files containing labels for the rows in our points matrix. If provided, we feed these labels to the NearestNeighborResult.

## How was this patch tested?

Mostly manual testing


turambar:
@emsixteeen

emsixteeen:
👍

PR #3878, opened 2017-08-17
An implementation of multi-probe LSH for the cosine distance.

huitseeker:
This can be used as guidelines for a poor man's ANN if the distribution of
the data is well-known enough to be able to predict an average distance
 between ANNs (the radius parameter of Panigraphi's entropy LSH).

However, because if uses a single hash table (as opposed to the several of the
vanilla LSH impelementation), it is relatively reasonable in its memory usage.

In the case of a very large index that would have to be split on several machins
the flavor of LSH that should be used is Layered LSH ( https://arxiv.org/abs/1210.7057 )

WIP, in particular the index creation does not use workspaces for now.

Opened as a request for comments (hi @turambar!) and perhaps an inspiration
to upgrade this to another implementation (hi @turambar!).

From the discard pile of high-performance ANN experiments.

review by @turambar for the math, @agibsonccc and @raver119 for the perf opts

### DIFF
--- a/deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/LSH.java
+++ b/deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/LSH.java
@@ -1,0 +1,77 @@
+package org.deeplearning4j.clustering.lsh;
+
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.api.ops.BaseAccumulation;
+
+/**
+ * This interface gathers the minimal elements for an LSH implementation
+ *
+ * See chapter 3 of :
+ * _Mining Massive Datasets_, Anand Rajaraman and Jeffrey Ullman
+ * http://www.mmds.org/
+ *
+ */
+public interface LSH {
+
+    /**
+     * Returns an instance of the distance measure associated to the LSH family of this implementation.
+     * Beware, hashing families and their amplificaiton constructs are distance-specific.
+     */
+    public String getDistanceMeasure();
+
+    /**
+     * Returns the size of a hash compared against in one hashing bucket, corresponding to an AND construction
+     *
+     * denoting hashLength by h,
+     * amplifies a (d1, d2, p1, p2) hash family into a
+     *                   (d1, d2, p1^h, p2^h)-sensitive one (match probability is decreasing with h)
+     *
+     * @return the length of the hash in the AND construction used by this index
+     */
+    public int getHashLength();
+
+    /**
+     *
+     * denoting numTables by n,
+     * amplifies a (d1, d2, p1, p2) hash family into a
+     *                   (d1, d2, (1-p1^n), (1-p2^n))-sensitive one (match probability is increasing with n)
+     *
+     * @return the # of hash tables in the OR construction used by this index
+     */
+    public int getNumTables();
+
+    /**
+     * @return The dimension of the index vectors and queries
+     */
+    public int getInDimension();
+
+    /**
+     * Popolates the index with data vectors.
+     * @param data the vectors to index
+     */
+    public void makeIndex(INDArray data);
+
+    /**
+     * Returns the set of all vectors that could approximately be considered negihbors of the query,
+     * without selection on the basis of distance or number of neighbors.
+     * @param query a  vector to find neighbors for
+     * @return its approximate neighbors, unfiltered
+     */
+    public INDArray bucket(INDArray query);
+
+    /**
+     * Returns the approximate neighbors within a distance bound.
+     * @param query a vector to find neighbors for
+     * @param maxRange the maximum distance between results and the query
+     * @return approximate neighbors within the distance bounds
+     */
+    public INDArray search(INDArray query, double maxRange);
+
+    /**
+     * Returns the approximate neighbors within a k-closest bound
+     * @param query a vector to find neighbors for
+     * @param k the maximum number of closest neighbors to return
+     * @return at most k neighbors of the query, ordered by increasing distance
+     */
+    public INDArray search(INDArray query, int k);
+}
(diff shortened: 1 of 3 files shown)


deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/LSH.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * This interface gathers the minimal elements for an LSH implementation
+ *
+ * See chapter 3 of :
+ * _Mining Massive Datasets_, Anand Rajaraman and Jeffrey Ullman
+ * http://www.mmds.org/
+ *
+ */
+public interface LSH {
+
```
agibsonccc: Could we make this an op maybe? It looks really similar. It feels like we could fit this in to the Op type.

deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/RandomProjectionLSH.java
```diff
@@ -0,0 +1,230 @@
...
+ *
+ */
+public class RandomProjectionLSH implements LSH {
+
+    @Override
+    public String getDistanceMeasure(){
+        return "cosinedistance";
+    }
+
+    @Getter int hashLength;
```
agibsonccc: private scope?

deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/RandomProjectionLSH.java
```diff
@@ -0,0 +1,230 @@
...
+
+        GaussianDistribution op1 = new GaussianDistribution(res, 0.0, 1.0 / Math.sqrt(shape[0]));
+
+        Nd4j.getExecutioner().exec(op1, rng);
+        return res;
+    }
+
+    public RandomProjectionLSH(int hashLength, int numTables, int inDimension, double radius){
+        this(hashLength, numTables, inDimension, radius, Nd4j.getRandom());
+    }
```
agibsonccc: Builder maybe?

deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/RandomProjectionLSH.java
```diff
@@ -0,0 +1,230 @@
...
+        return queryRes;
+    }
+
+    // data elements in the same entropy bucket as the query,
+    INDArray bucketData(INDArray query){
+        INDArray mask = bucket(query);
+        int nRes = mask.sum(0).getInt(0);
+        INDArray res = Nd4j.create(new int[] {nRes, inDimension});
+        int j = 0;
+        for (int i = 0; i < nRes; i++){
```
agibsonccc: @raver119 could probably do this in c++ in another rev..

deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/RandomProjectionLSH.java
huitseeker: I'm not sure: `hashLength`, `numTables`, `inDimension`, `radius`, are all required and not one has a sensible default.

deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/RandomProjectionLSH.java
raver119: Yes, we can fit anything we want into CustomOp...

deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/lsh/LSH.java
huitseeker: The hashing part maybe. How would you implement the query ?

huitseeker:
@raver119 made sure the random test data would be released in https://github.com/deeplearning4j/deeplearning4j/pull/3878/commits/55380591e8b712561aa423541300ea87dd467e18, as requested

AlexDBlack:
@huitseeker where are we at with this?

huitseeker:
1. IIUC, @agibsonccc suggested making `makeIndex` a custom op in a later round.
2. The suggestion of a builder for the constructor is problematic since any default that would be put here for any of the 4 arguments would be nonsensical. The builder would throw on any missing argument, which means we don't win much w.r.t a constructor, apart from implementing the pattern for the sake of it.  
3. other issues fixed.

AlexDBlack:
I'd be in favor of merging this then (and updating with a custom op later), rather than having the PR sit around open for weeks or months longer...

huitseeker:
@AlexDBlack Completely agree.

raver119:
I'm sorry, what customop you're speaking about? Can i have some details?

huitseeker:
@raver119 The custom op is @agibsonccc 's idea, mentioned here:
https://github.com/deeplearning4j/deeplearning4j/pull/3878#discussion_r140920657
So I don't want to put words in @agibsonccc 's mouth. But my interpretation is that we'd want to accelerate the one computationally significant operation in this, namely : 
https://github.com/deeplearning4j/deeplearning4j/pull/3878/files#diff-eda90c99b9e6b9ffe1c3cb301f48762dR126
It's the meat of the `makeIndex` method.

raver119:
Sure we can do that (after release thought, to avoid CUDA performance hassle there), but performance gains will be neglectable - this method is just 2 operations: gemm + sign transform, both are ok if data size is big enough. And for both cpu and CUDA, i'm afraid, gemm call will stay independent anyway due to performance concerns: cublas, mkl, openblas are pretty much overoptimzed these days. Just not worth trying to beat them to gain 1 additional operation on top of gemm.

huitseeker:
@raver119 Good ! even more of an argument to merge this then :)

PR #3882, opened 2017-08-18
Add support for "no bias" layers

AlexDBlack:
Should be good to go as-is.

Adds a ```.hasBias(boolean)``` config option (default: true, though false for embedding) for the following layers:
* DenseLayer, EmbeddingLayer
* Output (OutputLayer, RnnOutputLayer, CenterLossOutputLayer)
* Convolutional (ConvolutionLayer, Convolution1DLayer)

Of note:
* Config, parameter initializer, forward/backward passes updated to support this
* Added tests + gradient checks
* CuDNN convolutional layer was also tested (using dummy/all 0s 'bias' array)
* EmbeddingLayer: now defaults to having no biases (i.e., hasBias=false by default)


tomthetrainer:
@AlexDBlack 

I tried searching for papers on the importance of this and was not able to find anything. Bias is not a particular discriminating search pattern for papers on this. 

Can you answer some quick questions at some point before or after this is merged so I can see that some documentation is written. 

1. Is there a Paper that describes why this feature is useful? 
 2. Just in general why is it useful. 

No need for burdensome research just a quick brain dump would help. 

Thanks

AlexDBlack:
@tomthetrainer I haven't really seen anything in the literature on "no bias" layers. For most users (99% of the time) they will use layers with biases (which is the default). The only exception there is embedding layer (which now defaults to noBias = true) which is more in line with how word2vec operates.

The main reason we are adding this is for model import support from Keras.

maxpumperla:
@AlexDBlack it's not all that important, but I would vote for quickly refactoring `noBias` to `useBias` or `hasBias`. Two reasons:
a) it would be closer to keras, which, as you say, is the main reason we do this.
b) it's cognitively easier to avoid double negatives (current default is noBias false). It's trivial on its own, but when you build complex models you want to keep arguments as intuitive as possible.

AlexDBlack:
@maxpumperla yeah, that thought did cross my mind. I was on the fence about that when implementing it. I'll switch to ```hasBias``` I think (*use*Bias could be misinterpreted as meaning the bias exists but isn't used... hasBias is the most unambiguous option IMO).

AlexDBlack:
I've switched ```noBias``` to ```hasBias```, should be good to go.
Unless anyone wants to review, this can probably be merged.

Issue #3897, opened 2017-08-20
Unnecessary invocation of builder.regularization(true) when employing regularization


stolsvik:
When using the NeuralNetConfigurationBuilder.Builder, the examples show that you must both set regularization(true), then l1(..) or l2(..) or dropout(..) (others?).

This is hard to understand - does this toggle some special mode? If I do not call that, won't the l2(..) call do anything?

If it is important for l1(..) and the other regularization methods to have some state in the config toggled to true, could they not set it themselves?

The fluency of the API is somewhat hindered by this - you need to read those two methods in conjunction with each other. The method regularization() would make sense if it return a new Regularization.Builder() instance or somesuch, but it doesn't.

I assume the call to regularization(true) does not do *anything* by itself, i.e. without also calling l1(..) or similar? If it does, then it should be javadocced better.

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/3900

Issue #3896, opened 2017-08-20
Layer.Builder.dropout(..) needs better javadoc


stolsvik:
The JavaDoc should be specific where the dropout applies: Is it "above" or "below" this layer? Compared to if you used an actual DropoutLayer. Also, the name of the dropout parameter should be "retainFraction" or similer wording that implies "the number of neurons retained", since this is pretty important to get right. (it is currently "dropout", arguably easy to misunderstand for the exact opposite).

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/3900

Issue #3895, opened 2017-08-20
units(number) alias for nOut(number)


stolsvik:
When using the fluent API with ListBuilder, you actually do not need to specify the nIn(...) (number of neurons in).

The nOut(..) method then seems a tad strange. It would be better if it was named units(..), neurons(..), size(..), number(..). Keras uses "units".

With this change, along with #3888 and #3889, a config could look like this:

                .list()
                .inputType(InputType.feedForward(numInputs))
                .layer(new DenseLayer.Builder().units(1000).build())
                .layer(new DenseLayer.Builder().units(500).dropout(0.5d).build())
                .layer(new DenseLayer.Builder().units(1000).build())
                .layer(new OutputLayer.Builder(LossFunctions.LossFunction.NEGATIVELOGLIKELIHOOD)
                        .units(numOutputs)
                        .activation(Activation.SOFTMAX)
                        .build())
                .build();

.. which reads very compact and fluently. It also more to the point describes that *this layer* has 1000 neurons, which I think fits more with how most people visualize a network in their heads.

AlexDBlack:
Hm... I could probably be persuaded to add a .units(x) overload... Anyone else have an opinion on this?

maxpumperla:
@AlexDBlack I'd personally be in favour of a `units` overload.

tomthetrainer:
@AlexDBlack I would also be in favor of a `units` overload

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/3900

Issue #3889, opened 2017-08-20
ListBuilder should override the setInputType(..) method


stolsvik:
Currently, the setInputType(..) method exists on the MultiLayerConfiguration.Builder. The ListBuilder inherits this. The problem is that setInputType() (obviously) returns MultiLayerConfiguration.Builder, which makes it impossible to write a config as such:

                .list()
                .setInputType(InputType.feedForward(numInputs))
                .layer(0, new DenseLayer.Builder()
                        .nOut(1000)
                        .build())
                .layer(1, new DenseLayer.Builder()
                        .nOut(500)
                        .build())
                .layer(2, new DenseLayer.Builder()
                        .nOut(1000)
                        .build())
                .layer(3, new OutputLayer.Builder(LossFunctions.LossFunction.NEGATIVELOGLIKELIHOOD) //create hidden layer
                        .activation(Activation.SOFTMAX)
                        .nOut(numOutputs)
                        .build())
Notice how this can be "linearly" read as the entire layer-stack, with the inputs specified at top, and the output at bottom. (Combined with #3888, this would be pretty fluent).

You currently have to put the setInputType below the last layer, since you "loose" the ListBuilder type after that.

If the setInputType() was overridden on the ListBuilder class, but narrowing (specializing) the return type to ListBuilder, the above config would be valid.

The narrowing of return type was allowed from Java 5. https://stackoverflow.com/a/20514762/39334

**ADDED: I just realized that it would possibly be better if the method was named or aliased to inputType(..) instead of setInputType(..)**. Also, shortcuts for the most used InputTypes could be added: feedForwardInputs(28*28)...

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/3900

PR #3900, opened 2017-08-21
API Cleanup

AlexDBlack:
Good to go IMO. I'll leave this open a bit longer for reviews.

Addresses:
- https://github.com/deeplearning4j/deeplearning4j/issues/3889 (Fixed override, added proposed convenience method: ```.inputType().convolutional(h,w,d)```)
- https://github.com/deeplearning4j/deeplearning4j/issues/3888 (.layer(Layer) overload now exists, in addition to .layer(int, Layer))
- https://github.com/deeplearning4j/deeplearning4j/issues/3897 (.regularization(boolean) deprecated, use removed internally)
- https://github.com/deeplearning4j/deeplearning4j/issues/3896 (improved dropout javadoc)
- https://github.com/deeplearning4j/deeplearning4j/issues/3895 (.units(int) is equivalent to .nOut(int))

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
@@ -1022,7 +1022,7 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
          * Note 2: This sets the probability per-layer. Care should be taken when setting lower values for
          * complex networks (too much information may be lost with aggressive (very low) dropout values).<br>
          * Note 3: Frequently, dropout is not applied to (or, has higher retain probability for) input (first layer)
-         * layers. Dropout is also often not applied to our output layer. This needs to be handled MANUALLY by the user
+         * layers. Dropout is also often not applied to output layers. This needs to be handled MANUALLY by the user
          * - set .dropout(0) on those layers when using global dropout setting.<br>
          * Note 4: Implementation detail (most users can ignore): DL4J uses inverted dropout, as described here:
          * <a href="http://cs231n.github.io/neural-networks-2/">http://cs231n.github.io/neural-networks-2/</a>
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
@@ -98,7 +98,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
     //whether to constrain the gradient to unit norm or not
     //adadelta - weight for how much to consider previous history
     protected StepFunction stepFunction;
-    protected boolean useRegularization = false;
     protected boolean useDropConnect = false;
     //minimize or maximize objective
     protected boolean minimize = true;
@@ -620,7 +619,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
         protected int numIterations = 1;
         protected int maxNumLineSearchIterations = 5;
         protected long seed = System.currentTimeMillis();
-        protected boolean useRegularization = false;
         protected OptimizationAlgorithm optimizationAlgo = OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT;
         protected StepFunction stepFunction = null;
         protected boolean useDropConnect = false;
@@ -649,7 +647,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
                 maxNumLineSearchIterations = newConf.maxNumLineSearchIterations;
                 layer = newConf.layer;
                 numIterations = newConf.numIterations;
-                useRegularization = newConf.useRegularization;
                 optimizationAlgo = newConf.optimizationAlgo;
                 seed = newConf.seed;
                 stepFunction = newConf.stepFunction;
@@ -853,10 +850,10 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
         }
 
         /**
-         * Whether to use regularization (l1, l2, dropout, etc
+         * @deprecated Now: no-op. Regularization is always used when l1/l2/dropout is > 0
          */
+        @Deprecated
         public Builder regularization(boolean useRegularization) {
-            this.useRegularization = useRegularization;
             return this;
         }
 
@@ -980,7 +977,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
 
         /**
          * L1 regularization coefficient for the weights.
-         * Use with .regularization(true)
          */
         public Builder l1(double l1) {
             this.l1 = l1;
@@ -989,7 +985,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
 
         /**
          * L2 regularization coefficient for the weights.
-         * Use with .regularization(true)
          */
         public Builder l2(double l2) {
             this.l2 = l2;
@@ -998,7 +993,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
 
         /**
          * L1 regularization coefficient for the bias.
-         * Use with .regularization(true)
          */
         public Builder l1Bias(double l1Bias) {
             this.l1Bias = l1Bias;
@@ -1007,7 +1001,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
 
         /**
          * L2 regularization coefficient for the bias.
-         * Use with .regularization(true)
          */
         public Builder l2Bias(double l2Bias) {
             this.l2Bias = l2Bias;
@@ -1280,7 +1273,6 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
             conf.maxNumLineSearchIterations = maxNumLineSearchIterations;
             conf.layer = layer;
             conf.numIterations = numIterations;
-            conf.useRegularization = useRegularization;
             conf.optimizationAlgo = optimizationAlgo;
             conf.seed = seed;
             conf.stepFunction = stepFunction;
@@ -1329,7 +1321,7 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
                     sl.setConvolutionMode(convolutionMode);
                 }
             }
-            LayerValidation.generalValidation(layerName, layer, useRegularization, useDropConnect, dropOut, l2, l2Bias,
+            LayerValidation.generalValidation(layerName, layer, useDropConnect, dropOut, l2, l2Bias,
                             l1, l1Bias, dist);
         }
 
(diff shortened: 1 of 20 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -213,6 +214,7 @@ public double getL2ByParam(String variable) {
      * Fluent interface for building a list of configurations
      */
     public static class ListBuilder extends MultiLayerConfiguration.Builder {
+        private AtomicInteger layerCounter = new AtomicInteger(-1); //Used only for .layer(Layer) method
```
stolsvik: AtomicInteger sounds like ("communicates") you're expecting concurrency, while that is not true - the use of an ordinary map rules out concurrency. You could thus just use an int. 

However, I think it would be wiser to just use the "find max key in map" logic and then just invoke the indexed variant with max+1, as that is very straightforward and immediately understandable for a code-reader, and there is no possibility to start wondering what happens if you mix the methods. Right now, while reviewing this, I am wondering what happens if I e.g. add layers using index in out-of-order fashion, and then add one layer using the non-indexed method. After thinking a bit, I can't immediately see how that would fail - but my point is that this particular thinking should really not be necessary!

Also, JavaDoc, explaining what they do, and how they interact.

I would note that the "same index overrides" has tripped me up over a dozen times already, and I am still just getting to know the API! It should rather have thrown - it makes NO sense that this is possible, IMHO. It should at least be JavaDocced, and could potentially also output a log.warn(..)

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -261,6 +262,19 @@ public ListBuilder layer(Layer layer){
             return layerwise;
         }
 
+        @Override
+        public ListBuilder setInputType(InputType inputType){
+            return (ListBuilder)super.setInputType(inputType);
+        }
+
```
stolsvik: Since my ML-level hasn't reached anything else than feedforward at this point, this might be utterly meaningless: Are you sure it makes sense for a ListBuilder to accept all the different types of InputType? Will the ListBuilder - in this usage scenario - be used in a non-feedforward setting? I had this idea that these other types would be used "within" the network, but not as the top, "receiving" part? As I said, might be totally off base here.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -1008,17 +1008,23 @@ public Builder l2Bias(double l2Bias) {
...
+         * Note 1: This sets the probability per-layer. Care should be taken when setting lower values for
+         * complex networks (too much information may be lost with aggressive dropout values).<br>
+         * Note 2: Frequently, dropout is not applied to input (first layer) our output layer. This needs to be
+         * handled MANUALLY by the user - set .dropout(0) on those layers when using global dropout setting.<br>
+         * Note 3: Implementation detail (most users can ignore): DL4J uses inverted dropout, as described here:
+         * <a href="http://cs231n.github.io/neural-networks-2/">http://cs231n.github.io/neural-networks-2/</a>
          * </p>
          *
-         * @param dropOut Dropout probability (probability of retaining an activation)
+         * @param inputRetainFraction Dropout probability (probability of retaining an input activation for a layer)
```
stolsvik: Point out (again) here in the @param section that 0.0 _disables_ dropout, while 1.0 _effectively disables_ it, since it means that all nodes are retained.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
maxpumperla: e.g. Embedding would be another common type

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -1008,17 +1008,23 @@ public Builder l2Bias(double l2Bias) {
...
+         * Dropout probability. This is the probability of <it>retaining</it> an input activation for a layer. So
+         * dropOut(x) will keep an input activation with probability x, and set to 0 with probability 1-x.<br>
+         * dropOut(0.0) is disabled (default). When {@link #useDropConnect(boolean)} is set to true (false by default),
+         * this method sets the drop connect probability instead.
          * <p>
-         * Note: This sets the probability per-layer. Care should be taken when setting lower values for complex networks.
+         * Note 1: This sets the probability per-layer. Care should be taken when setting lower values for
+         * complex networks (too much information may be lost with aggressive dropout values).<br>
+         * Note 2: Frequently, dropout is not applied to input (first layer) our output layer. This needs to be
+         * handled MANUALLY by the user - set .dropout(0) on those layers when using global dropout setting.<br>
```
stolsvik: Point again out that 0 disables it. Point is that this disable-value is hard to grasp, as it semantically means "keep NO nodes"..

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -1008,17 +1008,23 @@ public Builder l2Bias(double l2Bias) {
         }
 
         /**
-         * Dropout probability. This is the probability of <it>retaining</it> an activation. So dropOut(x) will keep an
-         * activation with probability x, and set to 0 with probability 1-x.<br>
-         * dropOut(0.0) is disabled (default).
+         * Dropout probability. This is the probability of <it>retaining</it> an input activation for a layer. So
+         * dropOut(x) will keep an input activation with probability x, and set to 0 with probability 1-x.<br>
```
stolsvik: I am not sure whether I understand this. Because if I do, I think it is wrong.
Dropout should "zero out _nodes_", and I've come to understand that this means that their OUTPUT will be zero. Which again means that the NEXT layer will see a consistent set of zeroed and non-zeroed node values, i.e. if for Layer L node 0, 2 and 5 is zeroed, then ALL nodes in Layer L+1 will see node 0, 2 and 5 as zeroed.

Reading this JavaDoc, I get the feeling that node 0 in Layer L+1 (which is THIS layer, since we're talking "input activations") might see that the input activation is zeroed for node 1, 2, 3 from Layer L, while node 1 in Layer L+1 could see zeroed nodes for 0, 2 and 3 - i.e. independently zeroed.

If the latter is the case, then I believe this dropout is implemented incorrectly. But I probably misunderstand the wording.

Also, with dropout, one should have this scaling thing. This should be mentioned in the JavaDoc. Here's TensorFlow's dropout doc:
https://www.tensorflow.org/api_docs/python/tf/nn/dropout

"With probability keep_prob, outputs the input element scaled up by 1 / keep_prob, otherwise outputs 0. The scaling is so that the expected sum is unchanged."

Which brings us to the next thing that should be javadocced, and that is that dropout should only be applied in the training phase, not the evaluation. In particular, this also goes for the scaling. Since when using TensorFlow, you as user needs to set the keep_prob to your wanted value when training, while in evaluation mode, **you need to set it to 1**. Using e.g. "0.5", this means that when training, half of the nodes are zeroed, but the output is scaled by 2 for the remaining. But in evaluation, no nodes are silenced, and there is no scaling.

Is this handled automatically by sl4j? Would be nice to have it here in the JavaDoc.

Regarding whether it is "before" or "after" this layer: This could be explicitly explained like "setting a dropout on this layer is the same as adding a {@link DropoutLayer} _in front_ of this layer".

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -1008,17 +1008,23 @@ public Builder l2Bias(double l2Bias) {
...
+         * dropOut(0.0) is disabled (default). When {@link #useDropConnect(boolean)} is set to true (false by default),
```
stolsvik: Something like "dropout(0) is a special value that _disables_ the dropout mechanism, which effectively is the same as dropout(1.0) since that means to retain all nodes."

Problem is that 0 _semantically_ would mean to drop _all_ nodes, which can hinder understanding unless being very explicit about it.

maxpumperla: keras' definition is not exactly cleaner btw: 

rate: float between 0 and 1. Fraction of the input units to drop.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
stolsvik: Regarding input layer. The "dropout paper" (_the original!_) suggests 0.8 retain on input, 0.5 on hidden layers. So yes, it should definitely be different, but not necessarily "retain all".

AGAIN the point about "0": I consistently get a "huh?!"-feeling seeing that number - which implies that it should always be pointed out something like _".. thus disabling the dropout for this layer"_, or something.

Yes, I realize that I really don't like that special value, as it so directly contradicts the semantic of what the number means.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
AlexDBlack: Changed to int.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
AlexDBlack: Yes, it does make sense to accept all input types.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
AlexDBlack: Additional javadoc changes + changed to inputRetainProbability

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
AlexDBlack: Dropout zeros out activations, by definition.
The "scaling thing" is mentioned in the inverted dropout implementation.
Added note on train vs. test.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -1008,17 +1008,23 @@ public Builder l2Bias(double l2Bias) {
...
+         * this method sets the drop connect probability instead.
          * <p>
-         * Note: This sets the probability per-layer. Care should be taken when setting lower values for complex networks.
+         * Note 1: This sets the probability per-layer. Care should be taken when setting lower values for
+         * complex networks (too much information may be lost with aggressive dropout values).<br>
+         * Note 2: Frequently, dropout is not applied to input (first layer) our output layer. This needs to be
```
sorensf: Note 2: Frequently, dropout is not applied to input (first layer) _our_ output layer.

"our" should probably be "or". And since the java doc is the same for Layer.java any changes here should be reflected there as well.

PR #3902, opened 2017-08-21
Yolo2 object detection implementation

AlexDBlack:
Should be good to go.

YOLO v2 model implementation. https://arxiv.org/abs/1612.08242
Done:
    - Design, data layout
    - Config + scoring done (needs debugging), very basic tests passing
    - Backprop
    - Gradient checks
    - Convenience methods for getting predictions
    - Data pipeline code: https://github.com/deeplearning4j/DataVec/pull/411


crockpotveggies:
Did you need a model to the zoo that utilizes this? :)

Issue #2950, opened 2017-03-01
pls add AtrousConvolution to dl4J and(or) add AtrousConvolution2D to imported layers from keras


Apostol77794:
Hello. I'm trying to import a model from ssd https://github.com/rykov8/ssd_keras Keras and facing a problem. In dl4j no AtrousConvolution2D https://keras.io/layers/convolutional/#atrousconvolution2d . Implement it please in dl4j and in keras imports.

turambar:
@Apostol77794 Got it! This is on our radar now. @AlexDBlack FYI.

kmotohas:
Hello. I'd like to use SSD on DL4J too. Any progress on this topic?

turambar:
Reminder: we need to add a DL4J implementation of this layer before we can import it from Keras. So perhaps not @maxpumperla's purview?

maxpumperla:
as a side note: atrous conv layers have been deprecated in keras 2 (dilation now happens as additional argument through regular conv layers).

@AlexDBlack @turambar at some point we should discuss how to go about this. see #3413 for this as well

maxpumperla:
Since #3922 has been merged, we can next include this in model import. @Apostol77794 will let you know once it's done.

maxpumperla:
#3938

PR #3922, opened 2017-08-23
Add support for dilated convolutions

AlexDBlack:
Complete, should be good to go (maybe pending review).

- Adds dilation (atrous convolutions) for both ConvolutionLayer and subsampling layer
- Unlike Keras: no restriction on combinations of strides and dilation (Keras requires stride 1 with dilation > 1)
- Gradient checks: added and passing
- CuDNN Convolution layers support dilation (subsampling don't). Added + tested

https://github.com/deeplearning4j/nd4j/pull/2064
https://github.com/deeplearning4j/libnd4j/pull/439


maxpumperla:
see #2950

PR #3929, opened 2017-08-24
Keras Model & layer refactoring

maxpumperla:
Now that we have a bit more test coverage, I wanted to tear down the excessively large `KerasModel` and `KerasLayer` abstractions. I created a few utils classes with static helper functions that are individually testable.

Note that one thing that did change high-level is this:

Before we could write
```
new KerasModel.ModelBuilder...
```

not it's:
```
new KerasModel().modelBuilder()...
```
Since "everything" is equipped with a configuration now, it's getting more difficult to keep it all static. Are we OK with this change? @agibsonccc, what do you think? I'll adapt the documentation as well.

## How was this patch tested?

Tests were adapted accordingly


maxpumperla:
@agibsonccc also, this is the last "big" change I wanted to make. The module is now in a shape I can live with and it's, at least for me, easier to maintain and add new features.

Issue #3898, opened 2017-08-20
Add max-norm regularization


AlexDBlack:
http://jmlr.org/papers/volume15/srivastava14a/srivastava14a.pdf

Idea: constrain the L2 norm of the weights into a each hidden unit to  be less than or equal to a constant C.

maxpumperla:
@AlexDBlack maybe we can have a module similar to keras' `constraints`, see https://keras.io/constraints/

mbkennel:
There are multiple potential constraints which can be useful, constraining the average squared weight of all weights of a layer, i.e. one constraint for the layer, or a number of constraints equal to the number of the number of hiddens (could be outputs potentially) for that layer.   L2 constraining is very simple too, project to the ball's surface by the obvious scaling.   

Only question is how these may affect the various adaptive gradient algorithms and Nesterov momentum, and where to implement in the appropriate place.  E.g. for momentum, do we consider the "previous step" to be where the unconstrained step would have taken it, or where it ended up with the constraint?  These issues should be explicitly recognized and not silent accidents of implementation.

AlexDBlack:
Implemented here: https://github.com/deeplearning4j/deeplearning4j/pull/3957

Issue #3894, opened 2017-08-20
epochStart() - Way to notify the model that a new learning epoch starts


stolsvik:
Since epochs aren't really a part of the learning (you only get that implicit by the fit(iterator)), a method whereby the user  could notify the model that an epoch had started could maybe be of use.  E.g. model.epochStart().

This was brought up in #3503.

This could also be of use for things like ScoreIterationListener, as it would then know the difference between a mini batch, and an epoch.

The model should keep a counter of the epochs, and pass that to any  listeners. Note that when epoch 0 was started, this is before any learning had taken place. I personally find it nice to know how the score and accuracy etc are right after weight inits etc.

This method would obviously be invoked automatically when using the  fit(iterator), while if you used fit(dataset), you'd have to signal to  the model yourself when whatever you consider an epoch starts.

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/3957

Issue #2505, opened 2016-12-15
Regularization via constraint on weights instead (or in addition to) penalty term. 


mbkennel:
Request for enhancement or information for me to implement such an enhancement:

I would like the ability to apply a regularization constraint on weights.  This means that after taking a learning step in weight space, the weights are constrained to not exceed a certain L2 or L1 norm.  I

It is called 'max norm constraint' here: http://www.kdnuggets.com/2015/04/preventing-overfitting-neural-networks.html/2

I would like the values of the constraint parameter as supplied by the user to be on a 'per-weight' scale, and not a total sum scale.  So if one says the L2 constraint is 0.5, it means that 1/N \sum W_ij^2 <= 0.5 (or 0.5^2) with N the number of weights in W_ij == width * height.   If an L1 constraint it means  1/N \sum |W_ij|^2 <= 0.5 (or whatever the parameter is).   

If the constraint is violated (in the too big direction), then there is a truncation operation applied so move the weights as a whole back to the constraint surface, typically by the minimal L2 distance necessary. 

The constraint to L2 norm is simple, projection to a sphere, but constraint to L1 norm is not as obvious.  I know how to do it, but it requires an element-wise sorting procedure.  The question is where in the DL4J architecture this would go.  It cannot be computed on the raw gradient alone as it depends on the weight values after the gradient is applied. 

I would like this constraint to be able to be turned on and off (and its magnitude changed) layer by layer. I would like the constraint not to apply to any offset/bias term. 

The motivation for this enhancement is that typically "good" values of regularization constants are easier to find and tune in this formulation which tend to work well across multiple data sets.   With a penalty-based regularization one needs to balance the value of the error(cost) with the regularization penalty, but the quantitative magnitude of the error is often quite unknown when the modeler has started, and optimal values of regularization may be different to different orders of magnitude.  To the contrary, the typical sizes of weights in a "good" nnet (from prior experience) is often within a range of O(.1) to O(1) across a variety input datasets. 

A secondary consideration is that when starting training from small weights, the regularization is effectively off until training has proceeded further; this may speed training somewhat.

AlexDBlack:
http://cs231n.github.io/neural-networks-2/#reg
"Max norm constraints. Another form of regularization is to enforce an absolute upper bound on the magnitude of the weight vector for every neuron and use projected gradient descent to enforce the constraint."

Looks like there's 2 options here, for enforcing the constraint:
- Similar to L1/L2: just another gradient/update component
[...]

PR #3957, opened 2017-08-28
New functionality (epoch tracking and constraints)

AlexDBlack:
Should be good to go.

- Adds epoch tracking: https://github.com/deeplearning4j/deeplearning4j/issues/3894
- Adds constraints (API + implementations) https://github.com/deeplearning4j/deeplearning4j/issues/3898 https://github.com/deeplearning4j/deeplearning4j/issues/2505

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/layers/LayerConstraint.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/layers/LayerConstraint.java
@@ -1,0 +1,11 @@
+package org.deeplearning4j.nn.api.layers;
+
+import org.deeplearning4j.nn.api.Layer;
+import org.nd4j.shade.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+public interface LayerConstraint {
+
+    void applyConstraint(Layer layer, int iteration, int epoch);
+
+}
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/layers/LayerConstraint.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/layers/LayerConstraint.java
@@ -3,9 +3,13 @@ package org.deeplearning4j.nn.api.layers;
 import org.deeplearning4j.nn.api.Layer;
 import org.nd4j.shade.jackson.annotation.JsonTypeInfo;
 
+import java.io.Serializable;
+
 @JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
-public interface LayerConstraint {
+public interface LayerConstraint extends Cloneable, Serializable {
 
     void applyConstraint(Layer layer, int iteration, int epoch);
+
+    LayerConstraint clone();
 
 }
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
@@ -23,6 +23,10 @@ public abstract class BaseConstraint implements LayerConstraint {
     protected double epsilon = 1e-6;
     protected int[] dimensions;
 
+    protected BaseConstraint(){
+        //No arg for json ser/de
+    }
+
     protected BaseConstraint(boolean applyToWeights, boolean applyToBiases, int... dimensions){
         this(applyToWeights, applyToBiases, DEFAULT_EPSILON, dimensions);
     }
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
@@ -42,12 +42,12 @@ public abstract class BaseConstraint implements LayerConstraint {
         ParamInitializer i = layer.conf().getLayer().initializer();
         for(Map.Entry<String,INDArray> e : paramTable.entrySet()){
             if(applyToWeights && i.isWeightParam(e.getKey()) || applyToBiases && i.isBiasParam(e.getKey())){
-                apply(e.getValue(), i.isBiasParam(e.getKey()));
+                apply(e.getValue());
             }
         }
     }
 
-    public abstract void apply(INDArray param, boolean isBias);
+    public abstract void apply(INDArray param);
 
     public abstract BaseConstraint clone();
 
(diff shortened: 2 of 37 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
```diff
@@ -3183,6 +3185,20 @@ protected void clearLayersStates() {
...
+     * Increment the epoch count (in the underlying {@link MultiLayerConfiguration} by 1).
+     * Note that this is done <i>automatically</i> when using iterator-based fitting methods, such as
+     * {@link #fit(DataSetIterator)} or {@link #fit(MultiDataSet)}. However, when using non-iterator fit methods
+     * (DataSet, MultiDataSet, INDArrays etc), the network has no way to know when one epoch ends and another starts.
+     * In such situations, this method can be used to increment the epoch counter.<br>
+     * Note that the epoch counter is used for situations such as some learning rate schedules, and the like.
+     *
+     * The current epoch count can be obtained using {@code ComputationGraph.getConfiguration().getEpochCount()}
+     */
+    public void incrementEpochCount(){
```
huitseeker: Is this thread-safe, could we envision a situation with concurrent `setEpochCount` calls ?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/layers/LayerConstraint.java
```diff
@@ -0,0 +1,11 @@
+package org.deeplearning4j.nn.api.layers;
+
+import org.deeplearning4j.nn.api.Layer;
+import org.nd4j.shade.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+public interface LayerConstraint {
+
+    void applyConstraint(Layer layer, int iteration, int epoch);
```
huitseeker: I see where the call would come in the optimizer, but again it seems dangerous that this could be applied less than once (or more than once, depending on the constraint) per pass. Now that we're tracking epochs, could we have some epoch checking to make sure two successive calls are idempotent and less than one per epoch log a warning?

AlexDBlack: The way I'll be implementing it is that it'll get called once per iteration, after updates have been applied - it won't be possible to apply it more than once per iteration (and, even if it does: all of the implementations will give the same result with multiple sequential appliacions.
The epoch (and, iteration) number is just informational, to allow flexibility for users to have behaviour that depends on the number of epochs passed.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
AlexDBlack: No, it's not thread safe - but then nothing in MultiLayerNetwork or ComputationGraph is.
Setting can be done directly on the configuration objects.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
```diff
@@ -0,0 +1,64 @@
...
+    public void applyConstraint(Layer layer, int iteration, int epoch) {
+        Map<String,INDArray> paramTable = layer.paramTable();
+        if(paramTable == null || paramTable.isEmpty() ){
+            return;
+        }
+
+        ParamInitializer i = layer.conf().getLayer().initializer();
+        for(Map.Entry<String,INDArray> e : paramTable.entrySet()){
+            if(applyToWeights && i.isWeightParam(e.getKey()) || applyToBiases && i.isBiasParam(e.getKey())){
+                apply(e.getValue(), i.isBiasParam(e.getKey()));
```
maxpumperla: maybe we can replace `i.isBiasParam(e.getKey()` here simply with `true`?! It took me a while to realise that it doesn't matter what to use here, but this seems to indicate it has specifically to to with bias, which it doesn't

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/MaxNormConstraint.java
```diff
@@ -0,0 +1,68 @@
...
+    private MaxNormConstraint(){
+        //No arg for json ser/de
+    }
+
+    /**
+     * Apply to weights but not biases by default
+     *
+     * @param maxNorm        Maximum L2 value
+     * @param dimensions     Dimensions to apply to. For DenseLayer, OutputLayer, RnnOutputLayer, LSTM, etc: this should
+     *                       be dimension 1. For CNNs, this should be dimensions [1,2,3] correspending to last 3 of
```
maxpumperla: typo: corresponding

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/MinMaxNormConstraint.java
```diff
@@ -0,0 +1,87 @@
...
+        //No arg for json ser/de
+    }
+
+    /**
+     * Apply to weights but not biases by default
+     *
+     * @param max            Maximum L2 value
+     * @param min            Minimum L2 value
+     * @param dimensions     Dimensions to apply to. For DenseLayer, OutputLayer, RnnOutputLayer, LSTM, etc: this should
+     *                       be dimension 1. For CNNs, this should be dimensions [1,2,3] correspending to last 3 of
```
maxpumperla: correspending => corresponding. I think this is "everywhere".

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+    }
+
+    public abstract void apply(INDArray param, boolean isBias);
```
maxpumperla: afaict `isBias` is never actually used in the constraints you implemented. Why do we have it?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
AlexDBlack: No - some parameters are neither weights or biases... Consider for example batch normalization statistics, which are technically parameters in DL4J, but neither weights or biases.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
AlexDBlack: Removed.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/MaxNormConstraint.java
AlexDBlack: Fixed

PR #3958, opened 2017-08-28
Support ZeroPadding1D layers

maxpumperla:

Support ZeroPadding1D layers. @AlexDBlack I'd like your input on this, if possible. 

In the `CNN1DGradientCheckTest` test, if I replace `SIGMOID` by `TANH`, _a few_ of the tests will fail. Not sure why that is exactly, but something seems off.


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/ZeroPadding1DLayer.java
```diff
@@ -0,0 +1,148 @@
...
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Zero padding 1D layer for convolutional neural networks.
+ * Allows padding to be done separately for top and bottom.
+ *
+ * @author Alex Black
```
AlexDBlack: Wrong author tag.

AlexDBlack:
As noted in gitter - testCnn1DWithZeroPadding1D() is failing, and this looks like a legitimate issue to me.
The implementation itself looks fine - I'm guessing maybe an edge case in the underlying (existing) code somewhere.

I'll dig into this and debug it over the next couple of days.

maxpumperla:
@AlexDBlack just fyi, while this PR remains open, I'll also integrate the new layer into model import already. Originally wanted to do this as follow up, but makes sense now.

AlexDBlack:
I'd be fine with merging it now, but open an issue and make sure it gets fixed soon... most of the cases are working fine, and I suspect the bug isn't in the new code anyway (which means: we might be hitting the edge case elsewhere already).

Issue #3975, opened 2017-08-30
Gradient check edge cases found for certain padding configurations


maxpumperla:
#### Issue Description

When adding `ZeroPadding1D` layers, we noticed that in `CNN1DGradientCheckTest` test, if we replace `SIGMOID` by `TANH`, a few (about 4 out of 100) of the tests will fail.

After consulting with @AlexDBlack we might have hit an edge case somewhere in the underlying code. Need to further investigate this.

raver119:
Is it fixed now?

AlexDBlack:
Confirmed fixed - passes with TANH now too for that test.

PR #4004, opened 2017-09-05
[DONE] Constraints configurable on per-weight basis, richer API

maxpumperla:

Contraints can now be set on weights only, biases only or both. On request, one can also constrain custom parameters by name. For instance for LSTMs, we can do
```
                    .layer(new LSTM.Builder().nIn(12).nOut(10)
                            .recurrentConstraints(lc).build())
```
to constrain "RW", i.e. recurrent weights. Added more tests for all of this.

Note that concerns of "what constraint to use" and "where to apply it to" have been separated.


deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNN1DGradientCheckTest.java
```diff
@@ -51,7 +51,7 @@ public void testCnn1DWithZeroPadding1D() {
         int zeroPadding = 2;
         int paddedLength = length + 2 * zeroPadding;
 
-        Activation[] activations = {Activation.SIGMOID, Activation.TANH};
```
AlexDBlack: I want this one to keep failing so it doesn't get forgotten/missed :)

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/layers/LayerConstraint.java
```diff
@@ -4,11 +4,24 @@
...
+     * in the provided epoch, after parameters have been updated.
+     *
+     * @param layer org.deeplearning4j.nn.api.Layer
+     * @param iteration given iteration as integer
+     * @param epoch current epoch as integer
+     * @param biasConstraint if constraint is applied to bias parameters
+     * @param weightConstraint if contraint is applied to weight parameters
+     * @param paramNames Parameter names to which to apply the constraints
+     */
+    void applyConstraint(Layer layer, int iteration, int epoch, Boolean biasConstraint,
```
AlexDBlack: I'd use ```boolean``` instead of ```Boolean``` - shouldn't allow null here on the public API IMO.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Layer.java
```diff
@@ -78,12 +79,18 @@
     protected String layerName;
     protected double dropOut;
     protected List<LayerConstraint> constraints;
+    protected Boolean hasBiasConstraints;
+    protected Boolean hasWeightConstraints;
+    protected Set<String> constraintParamNames;
```
AlexDBlack: I don't like this. The parameter names that the constraint applies to should be in the constraints themselves, not on the layer. More generally we should have only *one* field related to constraints, with all relevant information contained within it.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Layer.java
```diff
@@ -278,8 +301,29 @@ public T constraints(LayerConstraint... constraints) {
...
+         *
+         * @param constraints Constraints to apply to this layer
+         * @param hasBiasConstraints if contraints will be applied to bias parameters
+         * @param hasWeightConstraints if constraints will be applied to weight parameters
+         * @param constraintParamNames Set of parameter names that constraints will be applied to. Overrides the two
+         *                             boolean flags before.
+         * @return Constraints
+         */
+        public T constraints(List<LayerConstraint> constraints, boolean hasBiasConstraints,
+                             boolean hasWeightConstraints, Set<String> constraintParamNames) {
```
AlexDBlack: This approach won't work: we can't use ```weightConstraints(x).biasConstraints(y)``` as the latter will override the former (i.e., x will be removed entirely, and only y will remain).
And again, the weight vs. bias vs. recurrent weight selection should IMO be in the constraints themselves, not as fields on the layer.

deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNN1DGradientCheckTest.java
maxpumperla: Since we have an open ticket with you assigned already (https://github.com/deeplearning4j/deeplearning4j/issues/3975) and I'd rather keep the build green, can we agree to leave it like this?

maxpumperla:
@AlexDBlack I think I addressed your concerns. We can now create very customised constraints, but to normal users it looks like a constraint is not actually linked to parameters (which internally they are). So it feels a little like keras, while we get around having to store different sets of constraints internally.

deeplearning4j-core/src/test/java/org/deeplearning4j/nn/conf/constraints/TestConstraints.java
```diff
@@ -21,21 +22,401 @@
...
 
+        LayerConstraint[] constraints = new LayerConstraint[]{
+                new MaxNormConstraint(0.5, 1),
+                new MinMaxNormConstraint(0.3, 0.4, 1.0, 1),
+                new NonNegativeConstraint(),
+                new UnitNormConstraint(1)
+        };
+
+        Set<String> customConstraint = new HashSet<>();
+        customConstraint.add("RW");
```
AlexDBlack: This customConstraint set isn't used anywhere?
We probably want' to check that the constraints are *not* applied to the other parameters too.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/constraint/BaseConstraint.java
```diff
@@ -20,18 +22,18 @@
 
     protected boolean applyToWeights;
     protected boolean applyToBiases;
+    protected Set<String> paramNames = new HashSet<>();
```
AlexDBlack: It should be IMO one or the other - applyToWeights/applyToBiases OR paramNames set - not both. My vote is for the latter only.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/LSTM.java
```diff
@@ -83,6 +85,20 @@ public LayerMemoryReport getMemoryReport(InputType inputType) {
     @AllArgsConstructor
     public static class Builder extends AbstractLSTM.Builder<Builder> {
 
+
+        public Builder recurrentConstraints(LayerConstraint... constraints) {
```
AlexDBlack: I was thinking, and I'm not 100% sure on this, but maybe we name everything ```constraintsX``` where X is {Weights, Recurrent, Bias} etc. That way all constraint methods show up together in the javadoc and code suggestions in IDEs.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
```diff
@@ -1245,8 +1246,14 @@ public Builder convolutionMode(ConvolutionMode convolutionMode) {
...
-            return constraints(Arrays.asList(constraints));
-        }
+            List<LayerConstraint> currentConstraints = this.constraints;
+            for(LayerConstraint c : constraints ){
+                BaseConstraint constraint = (BaseConstraint) c.clone();
+                constraint.setApplyToBiases(true);
+                constraint.setApplyToWeights(true);
+                currentConstraints.add(constraint);
+            }
+            return constraints(currentConstraints);         }
```
AlexDBlack: I'm not sure defaulting  to biases on the .constraints(...) method is a good idea?
I'd probably go for .constraints(LayerContstraint...) method requiring the user to manually specify the weight/bias keys (i.e., the weightConstraints and biasConstraint methods become the 'default' ones to use).

This current approach will also override whatever the user has set...

maxpumperla: @AlexDBlack what do you mean by "what the user has set"? This IS what the user sets. You mean what he sets on individual layers? There has to be some rule for what takes precedence and I think it makes sense that global config overrides local.

maxpumperla:
@AlexDBlack this is done from my side. All keras constraints are supported in model import. The nitpick you have above regarding constraint fields is understandable, but I don't feel like it justifies pouring more time into this right now.

PR #4014, opened 2017-09-06
Update docs for keras model import

maxpumperla:
- Updated original doc
- Added feature support page


tomthetrainer:
I would suggest that we begin to use a template for all documentation. 

Could these pages fit into the template linked below?

Basically the template enforces a structure of: 

Table of contents
Description
Details

If the Doc does not fit into the template structure for some reason, then propose another template for that style of doc. 

https://deeplearning4j.org/doc_templates/template_basic

tomthetrainer:
Besides the Template comment, everything looks good.

maxpumperla:
@tomthetrainer thanks, wasn't aware we already had these templates.

tomthetrainer:
Great work, thanks.

PR #4024, opened 2017-09-07
CompGraph vertices + Layers multiple inputs/outputs

AlexDBlack:
***WIP DO NOT MERGE (OBVIOUSLY)***

The plan here:
- Layers will support multiple inputs/outputs
- GraphVertex supports multiple inputs/outputs
- GraphVertex extends Layer. Conceptually, the idea: a GraphVertex *is* a Layer, but additionally has (at least as currently envisioned):
    - "store arrays and do forward/backward later" methods
    - connection structure information


AlexDBlack:
Obsolete - will be going with a different design than the one proposed here.

PR #4039, opened 2017-09-11
Log information proposing to use cuDNN when appropriate

### LINKED ISSUE #3868: Add warnings for cuda backend + no CuDNN
Idea: If user is...
(a) using CUDA backend, and
(b) using a layer that supports CuDNN (conv, subsampling, LRN, LSTM etc)

Warn the user that they would get much better performance by using CuDNN. And, provide a link on how to set it up.

This should be logged only once per JVM, regardless of how many networks/layer types are running without CuDNN.

saudet:
Fixes #3868

Logs information to suggest using cuDNN when appropriate and the page where to find more information. The output is displayed only once, but once for each layer...

## How was this patch tested?

Running an example with CUDA, but without cuDNN, displays the information, without stopping execution.


huitseeker:
This could use a rebase, but otherwise looks dope!

saudet:
Rebased!

Issue #3998, opened 2017-09-03
CompGraph: Multi-output vertices, multi-input layers


AlexDBlack:
Currently, vertices in ComputationGraph support multiple inputs, but only one output array.
This was an early design decision, as many multi-output cases could be composed of different single-output (and - it hasn't really come up until now). However, it may be advantageous to revisit and support this for  some use cases.
Care needs to be put into the API for this, however: currently we have...
```.addLayer("layer3", <layer>, "input0")``` etc
Perhaps default to first (usually only) output, otherwise use something like:
```.addLayer("layer3", <layer>, new Pair<>("layer1", 2))```

Multi-input layers would also be useful in some cases - for example, this would allow certain output layers (triplet embedding, etc) to be built from one object/layer, instead of from multiple vertices as they are currently.

Alternatively - a more radical proposal would be to align the layer and vertex APIs - refactor things such that a layer _is_ a vertex.

AlexDBlack:
FYI - planning to look at this issue this week.

PR #4050, opened 2017-09-12
API Refactoring, Layer Multi In/Out

AlexDBlack:
***DO NOT MERGE***

https://github.com/deeplearning4j/deeplearning4j/issues/3998

This is a major backend-refactoring PR with a number of goals:
1. Add layer multiple input support (done)
2. Add layer multiple output support (done)
3. Combine GraphVertex, NeuralNetConfiguration *and* Layer - just Layer (done, minus tests & debugging after removing NNC)
4. Clean up API and design (unused methods, deprecated methods/classes, refactor to avoid many methods that aren't applicable - like fit() on DenseLayer, reduce code duplication in a few places) (done)
5. Move to a less 'stateful' design that is cleaner, more maintainable, less error prone (no more setInput, setMask, setLabels, computeGradientAndScore, getGradients, clear sequence of ops -> instead just one computeGradientAndScore(input,labels) etc). Public API methods should by default clear the net of all inputs/gradients etc

Note that majority of these changes are internal API, not user-facing.

Implemented so far:
- Layer interface cleaned up (removed deprecated methods, methods that are never used in practice etc)
- Refactor class heirarchy
    - Model extends Layer (previously: other way around) - no more ```fit``` etc methods on DenseLayer etc; also no iteration listeners on non-fittable layers
    - Model is basically a "fittable Layer"
- Layer API supports multiple inputs and outputs
- Introduce ```Activations``` and ```Gradients``` classes
    - Used in layer API to pass around N activations/gradients (and masks, etc)
    - Activations class is also used to store activations/masks/mask state (instead of arrays)
- Preprocessor API supports multiple inputs/outputs
- Moved preprocessors to within layers, rather than separate (cleaner, and avoids code duplication in every single feed-forward/activate method)
    - Consequence: preprocessors are set directly on the layer config (```inputPreProcessor``` methods on GraphBuilder/ListBuilder are now deprecated)
- Combine activate + feedForwardMaskArrays methods - just activate now
    - One consequence: ```setLayerMaskArrays``` methods have been removed
    - Generally: just cleaner/better/less error prone vs. having to do two separate/parallel forward passes for activations vs. masks (and a bit more flexible/powerful)
- GraphVertex and Layer interfaces are combined - just "Layer" (still needs final cleanup)
- Remove no-arg fit(), feedForward() etc methods
- Cleaned up CompGraph feedForward methods (feedForward, rnnTimeStep, rnnActivateUsingStoredState all share single + simpler code)
- remove .iterations(x) configuration option (very rarely used, confusing for users)
- Output layers are not fittable in isolation (not used in practice anymore, used to be very occasionally used with RBMs)
- Multi-output support done
- NeuralNetConfiguration no longer used in MultiLayerConfiguration
- Added topological sort order to ComputationGraphConfiguration (model serialization)
- Compgraph + configuration support multiple output arrays for layers (underways)

Remaining tasks:
- Lots of backwards compatibility (saved models) work left to do
- Ensure all public API methods clear inputs (+ add unit tests for this) - convention should be to leave net it "clean" state after calling fit/output/feedForward etc
- All tests need to pass (most are back to passing - but some failures remain)
- Lots of general cleaning/polishing
- Model zoo appears to be broken
Regarding multi-output support for compgraph:
- Really the only issue users need to be aware of is how to specify which outputs to use, when multiple outputs are present.
    - Builder API: for first output, append ```/0``` to layer name; for second output append ```/1``` etc.
    - Example: second output from layer "myLayer":  ```.layer("out2", <layer here>, "myLayer/1")```

Breaking API changes:
- LastTimeStepVertex: now works as expected without requiring an input vertex name to be specified (now: just use no-arg constructor)
- fit(), feedForward() etc removed: use other methods with args (fit(DataSet), fit(INDArray) etc)
- .iterations(int) removed from config


maxpumperla:
https://github.com/deeplearning4j/deeplearning4j/blob/404a8a31856df84343e5f6ae974a316695b0ace3/deeplearning4j-core/src/test/java/org/deeplearning4j/optimize/solver/TestOptimizers.java#L192-L198

@AlexDBlack In these all these tests a `NeuralNetConfiguration` conf is used and passed through to specific classes extending `Model` and `Layer`. I'm not sure how to fit this into the new paradigm. For instance, without NNC this isn't available anymore:

```
 conf.addVariable("W");
```

maxpumperla:
@AlexDBlack how do we salvage this? Still think this is immensely useful, but seems to be extremely difficult to bring up to date.

Can we cherry pick some of the essential commits and start anew? Or try to rebase this to current master?

AlexDBlack:
@maxpumperla yeah, this was overly ambitious... I say we cherry pick from this and do it in stages rather than trying to rebase at this point

PR #4053, opened 2017-09-12
Fix imdb lstm config tests

maxpumperla:

We had issues with connecting LSTMs to Embedding layers in import tests. Thanks to @turambar for pointing me to the right place in preprocessor logic.


tomthetrainer:
Which model is this? Are there pointers to the weights and keras description of the model? I looked at fchollet's collection of weights and configs and did not see this. I assume it came from somewhere else.

maxpumperla:
https://github.com/fchollet/keras/blob/master/examples/imdb_lstm.py

@tomthetrainer

tomthetrainer:
@maxpumperla Thanks,

 I was in the habit of using fchollet's subset of examples with weights found here. 
https://github.com/fchollet/deep-learning-models

I was not aware of the other broader resource

PR #4066, opened 2017-09-14
Fix Keras 2 import of BatchNormalization

saudet:

Some attributes changed name for BatchNormalization with Keras 2:
https://github.com/fchollet/keras/wiki/Keras-2.0-release-notes

## How was this patch tested?

BatchNormalization from Darknet now loads...

### DIFF
--- a/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/normalization/KerasBatchNormalization.java
+++ b/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/normalization/KerasBatchNormalization.java
@@ -42,6 +42,8 @@ public class KerasBatchNormalization extends KerasLayer {
     private final String PARAM_NAME_BETA = "beta";
     private final String PARAM_NAME_RUNNING_MEAN = "running_mean";
     private final String PARAM_NAME_RUNNING_STD = "running_std";
+    private final String PARAM_NAME_MOVING_MEAN = "moving_mean";
+    private final String PARAM_NAME_MOVING_VARIANCE = "moving_variance";
 
     /**
      * Pass-through constructor from KerasLayer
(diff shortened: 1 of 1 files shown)


deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/normalization/KerasBatchNormalization.java
```diff
@@ -42,6 +42,8 @@
     private final String PARAM_NAME_BETA = "beta";
     private final String PARAM_NAME_RUNNING_MEAN = "running_mean";
     private final String PARAM_NAME_RUNNING_STD = "running_std";
+    private final String PARAM_NAME_MOVING_MEAN = "moving_mean";
```
maxpumperla: actually, when parameter names differ between keras 1 and 2 we reflect that through `KerasLayerConfiguration`. Please put `PARAM_NAME_MOVING_MEAN` and `PARAM_NAME_MOVING_VARIANCE` there (empty) and overwrite it in `Keras1LayerConfiguration` and `Keras2LayerConfiguration`, respectively. Each `KerasLayer` has a `conf` and you can get these parameters through lombok getters. This way we don't have these if-else statements all over the place.

In your situation it's a little more overhead, but it makes differing configs explicit in one central place.

saudet:
I see, thanks! But is this still relevant?
>  // Missing: keras 2 moving_mean_initializer, moving_variance_initializer

https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/config/KerasLayerConfiguration.java#L116

maxpumperla:
@saudet I've just seen that we already have the configs in place:

https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/config/KerasLayerConfiguration.java#L122-L123

just need to pick them up. And, yes, the initialisers for these parameters we don't reflect yet, so please leave that comment above in.

saudet:
Ok, should be good now :) Thanks

maxpumperla:
thanks @saudet

PR #4069, opened 2017-09-14
Prevent potential crashes in HDF5

saudet:

When calling `.asCommonFG()`, the original objects may get out of scope, and garbage collected prematurely, causing corruption in native memory. Here we make sure to deallocate them manually at the right time.

## How was this patch tested?

All unit tests pass, manually loading a Darknet model works.

### DIFF
--- a/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
+++ b/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
@@ -61,6 +61,21 @@ public class Hdf5Archive {
         this.file = new hdf5.H5File(archiveFilename, H5F_ACC_RDONLY);
     }
 
+    private hdf5.Group[] openGroups(String ... groups) {
+        hdf5.Group[] groupArray = new hdf5.Group[groups.length];
+        groupArray[0] = this.file.asCommonFG().openGroup(groups[0]);
+        for (int i = 1; i < groups.length; i++) {
+            groupArray[i] = groupArray[i - 1].asCommonFG().openGroup(groups[i]);
+        }
+        return groupArray;
+    }
+
+    private void closeGroups(hdf5.Group[] groupArray) {
+        for (int i = 0; i < groupArray.length; i++) {
+            groupArray[i].deallocate();
+        }
+    }
+
     /**
      * Read data set as ND4J array from group path.
      *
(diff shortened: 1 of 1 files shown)


deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
```diff
@@ -61,6 +61,21 @@ public Hdf5Archive(String archiveFilename) {
...
+    private hdf5.Group[] openGroups(String ... groups) {
+        hdf5.Group[] groupArray = new hdf5.Group[groups.length];
+        groupArray[0] = this.file.asCommonFG().openGroup(groups[0]);
+        for (int i = 1; i < groups.length; i++) {
+            groupArray[i] = groupArray[i - 1].asCommonFG().openGroup(groups[i]);
+        }
+        return groupArray;
+    }
+
+    private void closeGroups(hdf5.Group[] groupArray) {
```
huitseeker: What are your thoughts on 
```
hdf5.Group thisGroup = groupArray[i];
thisGroup._close();
```
?

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
```diff
@@ -61,6 +61,21 @@ public Hdf5Archive(String archiveFilename) {
         this.file = new hdf5.H5File(archiveFilename, H5F_ACC_RDONLY);
     }
 
+    private hdf5.Group[] openGroups(String ... groups) {
```
huitseeker: Neat. How do you see the reference pattern to this array evolve? If it's always the last element being accessed, could we make it a stack and save ourselves the potential `OutOfBounds` issues?

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
saudet: Examples from the HDF5 documentation uses the `delete` operator and that's called with `dellocate()`...

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
saudet: Don't know, it hasn't really evolved, maybe that's all we'll ever need from HDF5, @maxpumperla ? @turambar ?

saudet:
Hum, I'm getting other kinds of crashes now sometimes. It might be that HDF5 needs to have everything deallocated in order, including everything called in the other methods...

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
huitseeker: I was inspired by §5.4 here : https://support.hdfgroup.org/HDF5/doc1.6/UG/09_Groups.html

saudet:
Ok, that seems to do it. Not getting any more random crashes for now...

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/Hdf5Archive.java
saudet: That refers to the C API. The C++ API seems to be a bit wonky...

huitseeker: Haha, fair enough.

saudet:
Eh, that's why we should do the least possible in C/C++ :) Thanks!

Issue #4057, opened 2017-09-13
ComputationGraph topological sort is not guaranteed to be consistent across platforms


tail-island:
#### Issue Description

I'm writing an Android app using a neural-network imported from Keras 1.2.2.

Returned values from ComputationGraph#outputSingle() varies by CPU. On x86 Android emulator, output values are correct. But on ARM Android device, they are invalid.

Reproducing program is [here](https://github.com/tail-island/try-dl4j-android). Please run androidTest with Android emulator and ARM Android device. testWithComputationGraph() will fail on ARM only.

#### Version Information

* Deeplearning4j 9.0.1
* JavaCPP Presets for OpenBLAS 0.2.19-1.3
* Ubuntu 16.04
* ASUS ZenFone AR (Android 7.0)
* Android Studio 2.3.3 (Android Emulator 26.1.4, Android SDK Plargorm-Tools 26.0.0, Android SDK Tools 26.0.2)

#### Contributing

I think this issue's reason is below:

* Result of ComputationGraph#topologicalSortOrder() is different between x86 and ARM. The reason of difference is that HashMap iterating order is different between x86 and ARM.
* Neural-network will be initialized with invalid parameters. Because ComputationGraph#init() assigns parameters following topological sort order.
* If neural-network parameters are invalid, output values will be invalid.

Below Android app code outputs invalid values on ARM.

```
private void wrongOutputOnARM() throws IOException, ClassNotFoundException {
    // Adding reference. Because the model is imported from Keras and Reflections does not work on Android...
    NeuralNetConfiguration.reinitMapperWithSubtypes(Arrays.asList(new NamedType(TensorFlowCnnToFeedForwardPreProcessor.class)));

    // Loading model.
    ComputationGraph model = getModel(R.raw.model);

    // Prints output values. Correct on x86 but invalid on ARM...
    System.out.println(model.outputSingle(getInput(R.raw.input1), getInput(R.raw.input2), getInput(R.raw.input3), getInput(R.raw.input4)));
}

// Loading model.
private ComputationGraph getModel(int resourceId) throws IOException {
    try (InputStream stream = getResources().openRawResource(resourceId)) {
        return ModelSerializer.restoreComputationGraph(stream);
    }
}

// Loading input.
private INDArray getInput(int resourceId) throws IOException, ClassNotFoundException {
    try (InputStream stream = getResources().openRawResource(resourceId)) {
        return (INDArray)new ObjectInputStream(stream).readObject();
    }
[...]

PR #4072, opened 2017-09-14
Fix issue #4057.

tail-island:

Fix issue #4057.

ComputationGraph#topologicalSortOrder() is not stable on ARM CPU. Neural-net parameters will be invalid when calling ModelSerializer#restoreComputationGraph()/ComputatonGraph.init() with parameters.

## How was this patch tested?

Just run `mvn test -Dlombok.version=1.16.16 -Dcommonsio.version=2.4 -Dnd4j.version=0.9.2-SNAPSHOT -P test-nd4j-native` on the deeplearning4j-nn directory.

Sorry, I can't add tests because I don't know how to add ARM test.

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
@@ -1215,7 +1215,7 @@ public class ComputationGraph implements Serializable, Model, NeuralNetwork {
             i++;
         }
 
-        Map<Integer, Set<Integer>> inputEdges = new HashMap<>(); //key: vertex. Values: vertices that the key vertex receives input from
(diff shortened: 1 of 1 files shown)


AlexDBlack:
I appreciate the PR, but I'm not to merge this.

Yes, the current topological sort order depends (incorrectly) on the iteration order of the inputEdges hashmap - and a linked hash map should give us a consistent order for all platforms. However, there is no guarantee that the LinkedHashMap iteration order will be the same as the HashMap - and consequently, we could (silently!) break a decent number of saved ComputationGraph networks with this change.

Yes, this change need to occur - but I'm only going to do so when we have a thorough set of tests for backward compatibility with saved networks.
I already have to build exactly those sorts of tests for this pull request anyway: https://github.com/deeplearning4j/deeplearning4j/pull/4050

PR #4079, opened 2017-09-15
Minor changes for VocabConstructor

raver119:
This pr adds:
- few tests to ensure parallel tokenization handled properly
- gets rid of few legacy sleeps
- javadoc updates

### DIFF
--- a/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/paragraphvectors/ParagraphVectors.java
+++ b/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/paragraphvectors/ParagraphVectors.java
@@ -1211,6 +1211,10 @@ public class ParagraphVectors extends Word2Vec {
         /**
          * This method defines whether negative sampling should be used or not
          *
+         * PLEASE NOTE: If you're going to use negative sampling, you might want to disable HierarchicSoftmax, which is disabled by default
+         *
+         * Default value: 0
+         *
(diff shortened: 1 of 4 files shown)


deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/paragraphvectors/ParagraphVectors.java
```diff
@@ -1211,6 +1211,10 @@ public Builder useAdaGrad(boolean reallyUse) {
         /**
          * This method defines whether negative sampling should be used or not
          *
+         * PLEASE NOTE: If you're going to use negative sampling, you might want to disable HierarchicSoftmax, which is disabled by default
```
huitseeker: The comment doesn't make sense on his own: why disable HierarchicSoftmax if it's disabled by default? You mean enabled.

deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/word2vec/Word2Vec.java
```diff
@@ -319,6 +319,10 @@ public Builder useAdaGrad(boolean reallyUse) {
         /**
          * This method defines whether negative sampling should be used or not
          *
+         * PLEASE NOTE: If you're going to use negative sampling, you might want to disable HierarchicSoftmax, which is disabled by default
```
huitseeker: Same as above.

raver119:
ping @huitseeker

PR #4089, opened 2017-09-19
Update URL of local nexus repo, as we moved from Nexus v2 to v3

huitseeker:
make clear the staging repo is inoperant, as Nexus Pro is not deployed


huitseeker:
thanks!

PR #4149, opened 2017-10-05
Add Darknet19 to the zoo

saudet:

Also added unit tests.

## How was this patch tested?

Unit tests pass, and manual test on a sample image, which gets classified as expected with the pretrained models.


deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/Darknet19.java
```diff
@@ -0,0 +1,220 @@
...
+        this.workspaceMode = workspaceMode;
+        this.cudnnAlgoMode = workspaceMode == WorkspaceMode.SINGLE ? ConvolutionLayer.AlgoMode.PREFER_FASTEST
+                        : ConvolutionLayer.AlgoMode.NO_WORKSPACE;
+    }
+
+    @Override
+    public String pretrainedUrl(PretrainedType pretrainedType) {
+        if (pretrainedType == PretrainedType.IMAGENET)
+            if (inputShape[1] == 448 && inputShape[2] == 448)
+                return "http://blob.deeplearning4j.org/models/darknet19_448_dl4j_inference.zip";
```
crockpotveggies: One nit-pick here is we should add a version to the extension such as the other pretrained models. So instead this would be `http://blob.deeplearning4j.org/models/darknet19_448_dl4j_inference.v1.zip`

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/Darknet19.java
```diff
@@ -0,0 +1,220 @@
...
+            else
+                return "http://blob.deeplearning4j.org/models/darknet19_dl4j_inference.zip";
```
crockpotveggies: Same here, filename should have v1 for version: `http://blob.deeplearning4j.org/models/darknet19_dl4j_inference.v1.zip`

saudet:
Good, it's done! Thanks

PR #4202, opened 2017-10-24
Yolov2OutputLayer - class predictions per bounding box

AlexDBlack:
***WIP DO NOT MERGE***

Fixes Yolo2OutputLayer to match paper.
- Current implementation: was as per Yolo v1 loss function, for class predictions (per grid cell)
- New implementation (and as per v2 paper), class predictions are per bounding box, instead of per grid cell

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/objdetect/Yolo2OutputLayer.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/objdetect/Yolo2OutputLayer.java
@@ -647,18 +647,17 @@ public class Yolo2OutputLayer extends AbstractLayer<org.deeplearning4j.nn.conf.l
         int h = networkOutput.size(2);
         int w = networkOutput.size(3);
         int b = layerConf().getBoundingBoxes().size(0);
-        int c = networkOutput.size(1)-5*b;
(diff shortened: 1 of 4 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/objdetect/Yolo2OutputLayer.java
```diff
@@ -661,10 +668,10 @@ public void clearNoiseWeightParams() {
...
                         py += y;
 
-                        if(sm == null){
-                            try (MemoryWorkspace wsO = Nd4j.getMemoryManager().scopeOutOfWorkspaces()) {
-                                sm = softmax.get(point(i), all(), point(x), point(y)).dup();
-                            }
+
+                        INDArray sm;
+                        try (MemoryWorkspace wsO = Nd4j.getMemoryManager().scopeOutOfWorkspaces()) {
+                            sm = softmax.get(point(i), point(box), all(), point(x), point(y)).dup();
```
saudet: I believe x and y are reversed here? Would explain why we had trouble with the classification part... @yquemener

yquemener: Most definitely!

PR #4215, opened 2017-10-27
Add TinyYOLO to the zoo

saudet:

Also added to unit tests.

## How was this patch tested?

Unit tests pass, and manual test on a sample image, which gets classified as expected with the pretrained models.


deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/TinyYOLO.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /** Returns {@code inputShape[2] / 32}, where {@code inputShape[2]} should be a multiple of 32. */
+    public int getGridHeight() {
+        return inputShape[2] / 32;
+    }
+
+/*
+    // Import from a Keras model
+    @Override
```
crockpotveggies: Why is this commented out? Perhaps we should just remove it completely?

saudet: It's basically what I used to import the Keras model, might as well keep it for reference I thought, but if it's not useful, we can remove it, sure.

saudet:
Adding a label utility class is probably a good idea as well, yes. Does it need to be in that strange JSON format or can we just use a normal text file because that's basically all we have from https://github.com/pjreddie/darknet/tree/master/data ?

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/TinyYOLO.java
crockpotveggies: Let's move it to documentation on either gh-pages or Basecamp.

crockpotveggies: Perhaps the model-zoo page? https://deeplearning4j.org/model-zoo

crockpotveggies:
We don't have a preference for any format so the text file is just fine.

saudet:
So, I've basically refactored the whole label utility into multiple classes, now a lot more useful. And I've moved the reference code for TinyYOLO in the Javadoc of the class. What do you think?

PR #4217, opened 2017-10-28
new dl4j tools

clavvis:

(These are new tools for dl4j users. Tools show several class content and show other informations.
User can save informations in file. Later I can add other functions.

Later I'd like to add some dl example using these tools.

## How was this patch tested?

Includes unit tests.

### DIFF
--- a/deeplearning4j-tools/.gitignore
+++ b/deeplearning4j-tools/.gitignore
@@ -1,0 +1,4 @@
+/.settings/
+/logs/
+/iris.txt
+graph-tmp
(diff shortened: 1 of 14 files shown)


agibsonccc:
Right so I think I mentioned this once already. This stuff can go in nd4j-common instead.

clavvis:
As a org/nd4j/tools?

agibsonccc:
Put it under nd4j-common instead. https://github.com/deeplearning4j/nd4j/tree/master/nd4j-common

PR #4235, opened 2017-11-02
Deconvolution layer

maxpumperla:

Deconvolution layer with the usual tests (setup + gradient checks). 

Update: We decided with @raver119 and @AlexDBlack to exclude "Same" from accepted border modes until we find the root cause. The layer should still be useful like this. Keras import is supported as well now and tests are passing.


maxpumperla:
@huitseeker review would be highly appreciated. Any hints for deconvolution arithmetic for same border mode to fix remaining test would be cool. I'm a little stuck :)

deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNNGradientCheckTest.java
```diff
@@ -645,35 +646,70 @@ public void testCnnZeroPaddingLayer() {
...
+        net.init();
+        net.getLayer(0).activate(Nd4j.rand(10, 7 * 7 * 3));
+    }
+
     @Test
-    public void testSeparableConv2D() {
+    public void testDeconvolution2D() {
         int nOut = 2;
 
         int minibatchSize = 3;
```
AlexDBlack: Minibatch size 1 is an important edge case (I've seen bugs appear only on that before).
Maybe check both 1 and 3

deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNNGradientCheckTest.java
```diff
@@ -719,6 +754,74 @@ public void testSeparableConv2D() {
                 }
             }
         }
+    }
+
+    @Test
+    public void testSeparableConv2D() {
+        int nOut = 2;
+
+        int minibatchSize = 3;
```
AlexDBlack: As above, test 1 & 3 (I realize this was already the case before this PR, but might as well change it here)

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/config/KerasLayerConfiguration.java
```diff
@@ -70,6 +70,7 @@
     private final String LAYER_CLASS_NAME_UPSAMPLING_1D = "UpSampling1D";
     private final String LAYER_CLASS_NAME_UPSAMPLING_2D = "UpSampling2D";
     private final String LAYER_CLASS_NAME_SEPARABLE_CONVOLUTION_2D = ""; // 1: SeparableConvolution2D, 2: SeparableConv2D
+    private final String LAYER_CLASS_NAME_DECONVOLUTION_2D = ""; // 1: Deconvolution2D, 2: Conv2DTranspose
```
AlexDBlack: Can't comment on the correctness of this, but it *seems* like an empty string wouldn't be correct here?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Deconvolution2D.java
```diff
@@ -0,0 +1,344 @@
...
+        if (builder.stride.length != 2)
+            throw new IllegalArgumentException("Stride should include stride for rows and columns (a 2d array)");
+        this.stride = builder.stride;
+        if (builder.padding.length != 2)
+            throw new IllegalArgumentException("Padding should include padding for rows and columns (a 2d array)");
+        this.padding = builder.padding;
+        this.cudnnAlgoMode = builder.cudnnAlgoMode;
+        this.cudnnFwdAlgo = builder.cudnnFwdAlgo;
+        this.cudnnBwdFilterAlgo = builder.cudnnBwdFilterAlgo;
+        this.cudnnBwdDataAlgo = builder.cudnnBwdDataAlgo;
```
AlexDBlack: All of these builder args (like, last 16 lines) should be handled in superclass constructor already... was there a reason for this?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Deconvolution2D.java
```diff
@@ -0,0 +1,344 @@
...
+            throw new IllegalStateException("Invalid input for Convolution layer (layer name=\"" + getLayerName()
+                    + "\"): Expected CNN input, got " + inputType);
+        }
+
+        return InputTypeUtil.getOutputTypeDeconvLayer(inputType, kernelSize, stride, padding, dilation,
+                convolutionMode, nOut, layerIndex, getLayerName(), Deconvolution2DLayer.class);
+    }
+
+
+    public static class Builder extends BaseConvBuilder<Builder> {
```
AlexDBlack: All of the duplicate methods heres hould be unnecessary. Was there a reason for it?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/Deconvolution2DLayer.java
```diff
@@ -0,0 +1,288 @@
...
+        INDArray bias = getParamWithNoise(DeconvolutionParamInitializer.BIAS_KEY, training);
+        INDArray weights = getParamWithNoise(DeconvolutionParamInitializer.WEIGHT_KEY, training);
+
+        //Input validation: expect rank 4 matrix
+        if (input.rank() != 4) {
+            String layerName = conf.getLayer().getLayerName();
+            if (layerName == null)
+                layerName = "(not named)";
+            throw new DL4JInvalidInputException("Got rank " + input.rank()
+                    + " array as input to SeparableConvolution2D (layer name = " + layerName + ", layer index = "
```
AlexDBlack: Wrong layer name (should be deconv, not separable)

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/Deconvolution2DLayer.java
```diff
@@ -0,0 +1,288 @@
...
+        } else {
+            pad = layerConf().getPadding();
+            outSize = ConvolutionUtils.getDeconvolutionOutputSize(input, kernel, strides, pad, convolutionMode, dilation); //Also performs validation
+        }
+
+        int outH = outSize[0];
+        int outW = outSize[1];
+
+
+        if (helper != null) {
```
AlexDBlack: Does CuDNN support deconv? If so, we need separate tests for it.
If not, we should probably override initializeHelper() and make it a no-op.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/params/DeconvolutionParamInitializer.java
```diff
@@ -0,0 +1,85 @@
...
+import org.deeplearning4j.nn.conf.distribution.Distributions;
+import org.deeplearning4j.nn.weights.WeightInitUtil;
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.api.rng.distribution.Distribution;
+import org.nd4j.linalg.indexing.NDArrayIndex;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class DeconvolutionParamInitializer extends ConvolutionParamInitializer {
```
AlexDBlack: I see DeconvolutionParamInitializer.WEIGHT_KEY (and bias) used elsewhere, but these are not defined here??

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/config/KerasLayerConfiguration.java
maxpumperla: no, that's fine. will get overwritten in keras 1&2 specific files.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/Deconvolution2DLayer.java
maxpumperla: good catch! :)

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/params/DeconvolutionParamInitializer.java
maxpumperla: The only difference is in the _shape_ of those weights, compare in `createWeightMatrix`:

deconv:
```
int[] weightsShape = new int[] {inputDepth, outputDepth, kernel[0], kernel[1]};
```

conv:
```
 int[] weightsShape = new int[] {outputDepth, inputDepth, kernel[0], kernel[1]};

```

maxpumperla: For this reason I inherit from `ConvolutionParamInitializer`, to keep all other things equal.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Deconvolution2D.java
maxpumperla: damn, you're right. I can delete the whole Builder...

maxpumperla:
@AlexDBlack I addressed all your points. Feel free to merge, unless I missed something.

Issue #4232, opened 2017-11-01
CompGraph Evaluation (ROC?) bug with loop vs. doEvaluation


AlexDBlack:
Confirmed to occur on 0.9.1, also reported to occur on master.

Uncomment the reset in the following code to get different evaluation behaviour.

https://gist.github.com/AlexDBlack/b927aebf2e38906e7942111b12f07b5e

AlexDBlack:
As discussed in gitter:
https://gist.github.com/AlexDBlack/4042ced2b88cdd10070b0f3837147e14

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/4239

PR #4239, opened 2017-11-03
ADSI/AMDSI reset fix

raver119:
AMDSI/ADSI should check for hasNext() before calling reset() at construction time


raver119:
Fix for: https://github.com/deeplearning4j/deeplearning4j/issues/4232

PR #4247, opened 2017-11-05
Update memory.md

stolsvik:

Some information that would have helped me understand (and enable to Google the problem) the situation I ended up in.

## How was this patch tested?

Doc change, n/a.

### DIFF
--- a/memory.md
+++ b/memory.md
@@ -21,7 +21,7 @@ Despite the differences between these two approaches, the idea stays the same: o
 
 # Configuring Memory Limits
 
-With DL4J/ND4J, there are two types of memory limits to be aware of and configure: The on-heap JVM memory limit, and the off-heap memory limit. Both limits are controlled via Java command line arguments:
+With DL4J/ND4J, there are two types of memory limits to be aware of and configure: The on-heap JVM memory limit, and the off-heap memory limit, where NDArrays live. Both limits are controlled via Java command line arguments:
 
 `-Xms` - this option defines how much memory JVM heap will use at application start.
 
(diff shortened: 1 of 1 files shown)


memory.md
```diff
@@ -35,8 +35,11 @@ Example: Configuring 1GB initial on-heap, 2GB max on-heap, 8GB off-heap:
 
 ```-Xms1G -Xmx2G -Dorg.bytedeco.javacpp.maxbytes=8G -Dorg.bytedeco.javacpp.maxphysicalbytes=8G```
 
+**Notice**: With GPU systems, the maxbytes and maxphysicalbytes settings currently also effectively defines the memory limit for the GPU, since the off-heap memory is mapped (via NDArrays) to the GPU - read more about this in the GPU-section below.
+
```
agibsonccc: Maybe we might want to add a link to: http://bytedeco.org/javacpp/apidocs/org/bytedeco/javacpp/Pointer.html#maxPhysicalBytes  http://bytedeco.org/javacpp/apidocs/org/bytedeco/javacpp/Pointer.html#maxBytes

Another thing mentioned was clarifying how gpu memory was allocated..the problem I had was how to express this without going down a rabbit hole of how gpu memory works.

stolsvik: There is evidently a difference between those two values, though. In my current setting, I want a large heap, so I used 
-Xms8G -Xmx16G -Dorg.bytedeco.javacpp.maxbytes=8G -Dorg.bytedeco.javacpp.maxphysicalbytes=8G
This resulting in that dl4j wouldn't even "boot" (crashed on this line: Nd4j.getMemoryManager().setAutoGcWindow(10000) in my code)
The reason was that physical size (total of JVM) at init point was 10GB, while "maxphysicalbytes" was 8GB.

This at least made the thing run:
-Xms8G -Xmx14G -Dorg.bytedeco.javacpp.maxbytes=8G -Dorg.bytedeco.javacpp.maxphysicalbytes=16G

saudet: Yes, physical memory usage of the whole process includes JVM heap. Please add that clarification as well since that would have helped you! Thanks

crockpotveggies:
LGTM. Regarding the *how*, feel free to treat content as usual. Will begin web docs overhaul after we get tutorials launched this week.

PR #4253, opened 2017-11-07
Basic EMR client for Spark Jobs

huitseeker:
Adds a basic EMR client 

## How was this patch tested?
Manually.


saudet:
With samediff we do now...

wmeddie:
Ah ok.

huitseeker:
This PR does not introduce a Java8 dependency. The `Optional` used in this PR is not the Java8 `Optional`.

deeplearning4j-scaleout/deeplearning4j-aws/pom.xml
```diff
@@ -46,6 +46,11 @@
                         <artifactId>jsch</artifactId>
                         <version>0.1.51</version>
                 </dependency>
+            <dependency>
+                <groupId>org.apache.spark</groupId>
+                <artifactId>spark-core_2.11</artifactId>
+                <version>2.1.0</version>
```
agibsonccc: Don't we want a ${spark.version} here? That or at least use an inherited version somewhere? Would rather not have multiple spark version declarations.

deeplearning4j-scaleout/deeplearning4j-aws/src/main/java/org/deeplearning4j/aws/emr/EmrConfig.java
```diff
@@ -0,0 +1,31 @@
...
+    Configuration toAwsConfig() {
+        Configuration config = new Configuration().withClassification(classification).withProperties(properties);
+        List<Configuration> subConfigs = new ArrayList<>();
+        for (EmrConfig conf : configs){
+            subConfigs.add(conf.toAwsConfig());
+        }
+        return config.withConfigurations(subConfigs);
+    }
+
+    protected String classification;
```
agibsonccc: Fields at top is normal convention?

deeplearning4j-scaleout/deeplearning4j-aws/src/main/java/org/deeplearning4j/aws/emr/S3Url.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Getter
+    private String bucket;
+    @Getter
+    private String key;
+
+    public S3Url(String url){
+        if (!url.startsWith("s3://") || url.length() < 5){
+            throw new IllegalArgumentException("S3 URL should start with s3://");
+        }
```
agibsonccc: Could we use URI.create/parse here instead?

PR #4255, opened 2017-11-07
components for a simple learnable element wise layer

Liujish931:

Added components for a simple element wise layer with parameters to upgrade 
       x \circ w + b 

the input and the output size must be the same !!
new ElementWiseMultiplicationLayer.Builder().nIn(n).nOut(n).build()

## How was this patch tested?
build, run, save and load passed with computationGraph

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
@@ -3,9 +3,9 @@ package org.deeplearning4j.nn.api.CustomParamInitializer;
 import org.apache.commons.math3.util.FastMath;
 import org.deeplearning4j.nn.api.ParamInitializer;
 import org.deeplearning4j.nn.conf.NeuralNetConfiguration;
-import org.deeplearning4j.nn.conf.distribution.Distributions;
 import org.deeplearning4j.nn.conf.layers.FeedForwardLayer;
 import org.deeplearning4j.nn.conf.layers.Layer;
+import org.deeplearning4j.nn.params.DefaultParamInitializer;
 import org.deeplearning4j.nn.weights.WeightInit;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.api.rng.distribution.Distribution;
@@ -20,22 +20,14 @@ import java.util.Map;
 import static org.deeplearning4j.nn.weights.WeightInitUtil.DEFAULT_WEIGHT_INIT_ORDER;
 
 /**
- * extended by jingshu
+ * created by jingshu
  */
-public class ElementWiseParamInitializer implements ParamInitializer {
+public class ElementWiseParamInitializer extends DefaultParamInitializer implements ParamInitializer {
 
     private static final ElementWiseParamInitializer INSTANCE = new ElementWiseParamInitializer();
 
     public static ElementWiseParamInitializer getInstance() {
         return INSTANCE;
-    }
-
-    public final static String WEIGHT_KEY = "W";
-    public final static String BIAS_KEY = "b";
-
-    @Override
-    public int numParams(NeuralNetConfiguration conf) {
-        return numParams(conf.getLayer());
     }
 
     @Override
@@ -111,41 +103,13 @@ public class ElementWiseParamInitializer implements ParamInitializer {
 
         return out;
     }
-    protected INDArray createBias(NeuralNetConfiguration conf, INDArray biasParamView, boolean initializeParameters) {
-        org.deeplearning4j.nn.conf.layers.FeedForwardLayer layerConf =
-                (org.deeplearning4j.nn.conf.layers.FeedForwardLayer) conf.getLayer();
-        return createBias(layerConf.getNOut(), layerConf.getBiasInit(), biasParamView, initializeParameters);
-    }
-
-    protected INDArray createBias(int nOut, double biasInit, INDArray biasParamView, boolean initializeParameters) {
-        if (initializeParameters) {
-            INDArray ret = Nd4j.valueArrayOf(nOut, biasInit);
-            biasParamView.assign(ret);
-        }
-        return biasParamView;
-    }
-
-
-    protected INDArray createWeightMatrix(NeuralNetConfiguration conf, INDArray weightParamView,
-                                          boolean initializeParameters) {
-        org.deeplearning4j.nn.conf.layers.FeedForwardLayer layerConf =
-                (org.deeplearning4j.nn.conf.layers.FeedForwardLayer) conf.getLayer();
-
-        if (initializeParameters) {
-            Distribution dist = Distributions.createDistribution(layerConf.getDist());
-            return createWeightMatrix(layerConf.getNIn(), layerConf.getNOut(), layerConf.getWeightInit(), dist,
-                    weightParamView, true);
-        } else {
-            return createWeightMatrix(layerConf.getNIn(), layerConf.getNOut(), null, null, weightParamView, false);
-        }
-    }
-
+    
     protected INDArray createWeightMatrix(int nIn, int nOut, WeightInit weightInit, Distribution dist,
                                           INDArray weightParamView, boolean initializeParameters) {
         int[] shape = new int[] {nIn};
 
         if (initializeParameters) {
-            INDArray ret = initWeights(nIn, //Fan in
+            INDArray ret = ElementWiseParamInitializer.initWeights(nIn, //Fan in
                     nOut, //Fan out
                     shape, weightInit, dist, DEFAULT_WEIGHT_INIT_ORDER,weightParamView);
             return ret;
@@ -215,3 +179,4 @@ public class ElementWiseParamInitializer implements ParamInitializer {
 
 
 }
+
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
@@ -1,4 +1,4 @@
-package Utils.Utils_dpl4j.CustomParamInitializer;
+package org.deeplearning4j.nn.api.CustomParamInitializer;
 
 import org.apache.commons.math3.util.FastMath;
 import org.deeplearning4j.nn.api.ParamInitializer;
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
@@ -1,0 +1,217 @@
+package Utils.Utils_dpl4j.CustomParamInitializer;
+
+import org.apache.commons.math3.util.FastMath;
+import org.deeplearning4j.nn.api.ParamInitializer;
+import org.deeplearning4j.nn.conf.NeuralNetConfiguration;
+import org.deeplearning4j.nn.conf.distribution.Distributions;
+import org.deeplearning4j.nn.conf.layers.FeedForwardLayer;
+import org.deeplearning4j.nn.conf.layers.Layer;
+import org.deeplearning4j.nn.weights.WeightInit;
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.api.rng.distribution.Distribution;
+import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.linalg.indexing.NDArrayIndex;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static org.deeplearning4j.nn.weights.WeightInitUtil.DEFAULT_WEIGHT_INIT_ORDER;
+
+/**
+ * extended by jingshu
+ */
+public class ElementWiseParamInitializer implements ParamInitializer {
+
+    private static final ElementWiseParamInitializer INSTANCE = new ElementWiseParamInitializer();
+
+    public static ElementWiseParamInitializer getInstance() {
+        return INSTANCE;
+    }
+
+    public final static String WEIGHT_KEY = "W";
+    public final static String BIAS_KEY = "b";
+
+    @Override
+    public int numParams(NeuralNetConfiguration conf) {
+        return numParams(conf.getLayer());
+    }
+
+    @Override
+    public int numParams(Layer layer) {
+        FeedForwardLayer layerConf = (FeedForwardLayer) layer;
+        int nIn = layerConf.getNIn();
+        return nIn*2; //weights + bias
+    }
+
+    /**
+     * Initialize the parameters
+     *
+     * @param conf             the configuration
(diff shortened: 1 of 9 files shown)


AlexDBlack:
Do you have a reference for this? For example, is this layer implemented in another deep learning library, or is there a paper describing this architecture? I don't believe I've come across this specific architecture before.

Also, as a general rule all layer in DL4J need to be implemented with gradient checks. Otherwise, the implementation could be incorrect and not learn.
https://github.com/deeplearning4j/deeplearning4j/tree/master/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck

Liujish931:
Thanks for reviewing :)
it is just a common layer where we add one weight for each feature (scaling), for example we can use it to implement a element wise addition vertex (add this layer to each input vertex before applying the addition vertex ) because we cannot add weights in the current version of element wise addition vertex .  This kind of op is used in many custom architectures like Wide&Deep (there is a weighted addition layer)

I will look at the gradient checks.

agibsonccc:
@Liujish931 we need tests like this: https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTests.java

Liujish931:
guys i returned to this pull request today and i found this error 
org.nd4j.linalg.exception.ND4JIllegalStateException: op.X dataType is [FLOAT] instead of expected [DOUBLE]
when i try to do 
boolean gradOK = GradientCheckUtil.checkGradients(mln, DEFAULT_EPS, DEFAULT_MAX_REL_ERROR,
                                    DEFAULT_MIN_ABS_ERROR, PRINT_RESULTS, RETURN_ON_FIRST_FAILURE, input, labels);

Is there something that i missed ?

AlexDBlack:
@Liujish931 yes - as per the error, gradient checks need to be run using double precision datatype for ND4J. This should be the default if running within DL4J unit tests. Otherwise, if you are running it elsewhere, you can use the same arg (-Ddtype=double) when running your test...

https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/pom.xml#L28-L38

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
```diff
@@ -0,0 +1,182 @@
...
+                    nOut, //Fan out
+                    shape, weightInit, dist, DEFAULT_WEIGHT_INIT_ORDER,weightParamView);
+            return ret;
+        } else {
+//            System.out.println(weightParamView.shapeInfoToString());
+            return weightParamView;
+        }
+    }
+
+    public static INDArray initWeights(double fanIn, double fanOut, int[] shape, WeightInit initScheme,
```
AlexDBlack: Is there a reason why the code from WeightInitUtil has to be copied to here?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/CustomLayer/ElementWiseMultiplicationLayer_impl.java
```diff
@@ -0,0 +1,160 @@
...
+import org.nd4j.linalg.indexing.INDArrayIndex;
+import org.nd4j.linalg.indexing.NDArrayIndex;
+import org.nd4j.linalg.primitives.Pair;
+
+import java.util.Arrays;
+
+/**
+ * created by jingshu
+ */
+public class ElementWiseMultiplicationLayer_impl extends BaseLayer<ElementWiseMultiplicationLayer> {
```
AlexDBlack: Rename to ElementWiseMultiplicationLayer for consistency with all other layers.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/CustomLayer/ElementWiseMultiplicationLayer_impl.java
```diff
@@ -0,0 +1,160 @@
+package org.deeplearning4j.nn.layers.CustomLayer;
```
AlexDBlack: Please move to org.deeplearning4j.nn.layers.feedforward.elementwise package

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/CustomLayer/ElementWiseMultiplicationLayer_impl.java
```diff
@@ -0,0 +1,160 @@
...
+        INDArray weightError = weightErrorSignal.mul(activation);
+        ret.gradientForVariable().put(ElementWiseParamInitializer.WEIGHT_KEY, weightError);
+        INDArray biasGradient = weightError.mean(0);
+        ret.gradientForVariable().put(ElementWiseParamInitializer.BIAS_KEY, biasGradient);
+
+        return ret;
+    }
+
+    @Override
+    public Pair<Gradient, INDArray> backpropGradient(INDArray epsilon) {
```
AlexDBlack: Commented out code in this method needs to be cleaned up.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/CustomLayer/ElementWiseMultiplicationLayer_impl.java
```diff
@@ -0,0 +1,160 @@
...
+        INDArray biasGrad = gradientViews.get(ElementWiseParamInitializer.BIAS_KEY);
+        delta.sum(biasGrad, 0); //biasGrad is initialized/zeroed first
+
+        ret.gradientForVariable().put(ElementWiseParamInitializer.WEIGHT_KEY, weightGrad);
+        ret.gradientForVariable().put(ElementWiseParamInitializer.BIAS_KEY, biasGrad);
+
+        INDArray epsilonNext = Nd4j.zeros(params.get(ElementWiseParamInitializer.WEIGHT_KEY).shape());
+
+        for(int row =0;row<delta.rows();row++){
+            epsilonNext.addi(params.get(ElementWiseParamInitializer.WEIGHT_KEY).mul(delta.getRow(row)));
```
AlexDBlack: This for loop is inefficient.
I believe it can be replaced by:
```epsilonNext = delta.mulRowVector(params.get(ElementWiseParamInitializer.WEIGHT_KEY).sum(0)```

File: deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/CustomLayer/ElementWiseMultiplicationLayer_impl.java [hunk 6]
```diff
@@ -0,0 +1,160 @@
...
+     * Returns true if the layer can be trained in an unsupervised/pretrain manner (VAE, RBMs etc)
+     *
+     * @return true if the layer can be pretrained (using fit(INDArray), false otherwise
+     */
+    @Override
+    public boolean isPretrainLayer() {
+        return false;
+    }
+
+    public INDArray preOutput(boolean training) {
```
AlexDBlack: Also remove commented out code here.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/CustomLayer/ElementWiseMultiplicationLayer_impl.java
```diff
@@ -0,0 +1,160 @@
...
+        if (maskArray != null) {
+            applyMask(ret);
+        }
+
+        return ret;
+    }
+
+
+    @Override
+    public INDArray activationMean() {
```
AlexDBlack: This method does not exist anymore on Layer API:
https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/Layer.java

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/CustomLayer/GradientCheckTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private static final boolean PRINT_RESULTS = true;
+    private static final boolean RETURN_ON_FIRST_FAILURE = false;
+    private static final double DEFAULT_EPS = 1e-6;
+    private static final double DEFAULT_MAX_REL_ERROR = 1e-3;
+    private static final double DEFAULT_MIN_ABS_ERROR = 1e-8;
+
+    private static Logger log = (Logger) LoggerFactory.getLogger(GradientCheckTest.class);
+
+    @Test
+    public void elementWiseMultiplicationLayerTest(){
```
AlexDBlack: This gradient check is a god start - but is insufficient. To properly test it, we require a layer below it (otherwise the epsilonNext component could be incorrect, and we would never know from this test). Something like a DenseLayer below it should be fine for that.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/api/CustomParamInitializer/ElementWiseParamInitializer.java
Liujish931: yes, the  matrix  w for the element wise layer is little different with the other, this is just for initialising the w  in the right form, which is a 1-d vector and the length = inputFeatureSize, otherwise it will initialise a 2-d matrix nIn*nOut

AlexDBlack: The ```int[] shape``` argument specifies the shape. You should be able to pass [1,featureSize] to the existing code, to get the correct shape weights.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/customLayerTest/GradientCheckTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.slf4j.LoggerFactory;
+
+import static org.deeplearning4j.gradientcheck.GradientCheckUtil.checkGradients;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * created by jingshu
+ */
+
+public class GradientCheckTest {
```
AlexDBlack: This file is in ```deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/customLayerTest/```
It should be in ```deeplearning4j-core/src/main/test/org/deeplearning4j/gradientcheck/```

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/feedforward/elementwise/ElementWiseMultiplicationLayer.java
```diff
@@ -0,0 +1,122 @@
...
+    public ElementWiseMultiplicationLayer(NeuralNetConfiguration conf){
+        super(conf);
+    }
+
+    public ElementWiseMultiplicationLayer(NeuralNetConfiguration conf, INDArray input) {
+        super(conf, input);
+    }
+
+    @Override
+    public Gradient error(INDArray errorSignal) {
```
AlexDBlack: This method no longer exists.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/feedforward/elementwise/ElementWiseMultiplicationLayer.java
```diff
@@ -0,0 +1,122 @@
...
+        INDArray W = getParam(ElementWiseParamInitializer.WEIGHT_KEY);
+        Gradient nextLayerGradient = new DefaultGradient();
+        INDArray wErrorSignal = errorSignal.mul(W);
+        nextLayerGradient.gradientForVariable().put(ElementWiseParamInitializer.WEIGHT_KEY, wErrorSignal);
+        return nextLayerGradient;
+    }
+
+    @Override
+    public Gradient calcGradient(Gradient layerError, INDArray activation) {
```
AlexDBlack: This method no longer exists.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/feedforward/elementwise/ElementWiseMultiplicationLayer.java
```diff
@@ -0,0 +1,122 @@
...
+
+
+import org.deeplearning4j.nn.params.ElementWiseParamInitializer;
+import org.deeplearning4j.exception.DL4JInvalidInputException;
+import org.deeplearning4j.nn.conf.NeuralNetConfiguration;
+import org.deeplearning4j.nn.gradient.DefaultGradient;
+import org.deeplearning4j.nn.gradient.Gradient;
+import org.deeplearning4j.nn.layers.BaseLayer;
+import org.deeplearning4j.nn.params.DefaultParamInitializer;
+import org.deeplearning4j.util.Dropout;
```
AlexDBlack: The import that this class refers to no longer exists.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/feedforward/elementwise/ElementWiseMultiplicationLayer.java
```diff
@@ -0,0 +1,122 @@
...
+
+        if ( input.columns() != W.columns()) {
+            throw new DL4JInvalidInputException(
+                    "Input size (" + input.columns() + " columns; shape = " + Arrays.toString(input.shape())
+                            + ") is invalid: does not match layer input size (layer # inputs = "
+                            + W.shapeInfoToString() + ") " + layerId());
+        }
+
+        if (conf.isUseDropConnect() && training && layerConf().getDropOut() > 0) {
+            W = Dropout.applyDropConnect(this, DefaultParamInitializer.WEIGHT_KEY);
```
AlexDBlack: This dropout section is obsolete. Replace if statement and content with ```applyDropOutIfNecessary(training);```

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/ElementWiseMultiplicationLayer.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public LayerMemoryReport getMemoryReport(InputType inputType) {
+        InputType outputType = getOutputType(-1, inputType);
+
+        int numParams = initializer().numParams(this);
+        int updaterStateSize = (int) getIUpdater().stateSize(numParams);
+
+        int trainSizeFixed = 0;
+        int trainSizeVariable = 0;
+        if (getDropOut() > 0) {
```
AlexDBlack: Method no longer exists. Replace condition with ```if (getIDropout() != null) {```

Liujish931: ah sorry yes i was testing on another project, i test this time on deeplearning4j/master and i cannot find this method .

Liujish931: for the moment i tried to build with the build stack script but it tried to test cuda which i dont have , anyway i pushed the modification you mentioned , but i dont know if it works on master

AlexDBlack:
Merged here, with modifications: https://github.com/deeplearning4j/deeplearning4j/pull/4414

PR #4266, opened 2017-11-08
Fix for Issue 4249, "RuntimeException: "Can't allocate [HOST] memory"…

stolsvik:
… when using UIServer with StatsListener"

Using an explicit memory Workspace for the statistics-gathering. This
ensures that we don't end up in GC-collecting-mode for the NDArrays that
are duplicated, which evidently is a problem on my machine.


raver119:
@AlexDBlack do you remember, why exactly we're excluding UI from workspace during training? There was something...

stolsvik:
One thing I realized when committing this evident non-fix, was that what's the use of doing all that copying/duplication of the NDArrays? Why not just calculate the mean, std-dev and mean magnitudes right there in the onForwardPass(..) (x2) and onGradientCalculation(..)? Unless I am missing something, that would be a much better solution, alleviating the need to keep those NDArrays around. I guess it should both reduce memory footprint and increase speed too.

Notice also that I did some micro changes to the "if" within the onForwardPass - it was different between the two, which didn't make immediate sense. Also, the use of the two HashMaps storing the activations and gradient had slightly different semantics between them: Seemed like a better idea to just use one instance, and clear them as fast as possible after use - not in front of the next use.

AlexDBlack:
Closing this: this PR should have fixed the memory issues: https://github.com/deeplearning4j/deeplearning4j/pull/4292

PR #4272, opened 2017-11-09
NearestNeighboursServer logging improvements

AlexDBlack:


AlexDBlack:
Right, that's a fair point. It's probably possible using relative paths or something... Though I guess if we are doing that - we might as well do it with logging configuration (logback.xml) too?

huitseeker:
As far as I'm concerned, auxiliary files (applicaiton,conf, logback ...):
- can arguably be duplicated in the main,
- should be minimized to stubs or relative in the test.

AlexDBlack:
Hm, though if application.conf and logback.xml for test resources are defined in relative terms, they won't be on the classpath (which, alone, won't be sufficient).  I agree with the motivation to reduce duplication though... I think our best option might be to:

(a) Have a single copy in say deeplearning4j-core or -nn (unless somewhere else would be better?)

(b) Configure maven surefure plugin to set relevant system properties for each library that requires them
http://maven.apache.org/surefire/maven-surefire-plugin/examples/system-properties.html
Play: -Dconfig.file=<path here>
Logback: -Dlogback.configurationFile=<path here>

https://www.playframework.com/documentation/2.4.x/ProductionConfiguration#Specifying-an-alternate-configuration-file
https://logback.qos.ch/manual/configuration.html

AlexDBlack:
@huitseeker thanks for the review, I've changed it to use the existing one instead (via system property). Seems to be finding it ok using that approach.

For logback.xml: I've opened this issue https://github.com/deeplearning4j/deeplearning4j/issues/4288

PR #4279, opened 2017-11-10
fit(DataSetIterator) without pretrain

clavvis:

MultiLayerNetwork method fit( DataSetIterator) do in one epoch pretraining and backpropagation too. It is not suitable.

Training NN needs some epochs only pretraining and later some epochs only backpropagation.

I suggest:
- do pretraning for DataSetIterator in method pretrain(DataSetIterator)
- in method fit(DataSetIterator) do only backpropagation

In pretrain(DataSetIterator) method I added for TrainingListener's set onEpochStart(this).

## How was this patch tested?

I checked manually code.


clavvis:
I changed ComputationGraph and added java doc.

PR #4282, opened 2017-11-10
MultiLayerNetwork  - add getTrainingListeners

clavvis:

In MultiLayerNetwork I created getTrainingListeners. It is preparation for solving problem: For pretraing could be necessary other listeners than for backpropagation.

## How was this patch tested?

I checked manually code.

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
@@ -1573,7 +1573,7 @@ public class MultiLayerNetwork implements Serializable, Classifier, Layer, Neura
     *
     * @return trainingListeners
     */
-   public Collection<trainingListeners> getTrainingListeners() {
+   public Collection<TrainingListener> getTrainingListeners() {
        return trainingListeners;
    }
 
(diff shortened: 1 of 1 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java
```diff
@@ -1563,12 +1563,20 @@ protected void truncatedBPTTGradient() {
...
     public Collection<IterationListener> getListeners() {
         return listeners;
     }
 
+    /**
+    *
+    * @return trainingListeners
+    */
+   public Collection<trainingListeners> getTrainingListeners() {
+       return trainingListeners;
```
agibsonccc: This doesn't look right. Shouldn't this be camel case? We don't have all lower case classes.

clavvis: I don't see. First word (get) should have small letter. Next word (Training) should have capital letter to separate sense. Last word (Listeners) should have capital letter to separate sense and it's first capital letter correspond with collection name. For me it is standard name.
In MultiLayerNetwork I see a lot of similar.

AlexDBlack: The method name is fine, but ```Collection<trainingListeners>``` won't compile. It should be ```Collection<TrainingListener>``` instead.

clavvis:
My mistake. I have repaired it.

PR #4284, opened 2017-11-10
MultiLayerNetwork - pretrainLayer - allow/block, check

clavvis:

1) Method pretrainLayer(int layerIdx, INDArray features) in MultiLayerNetwork includes code:
        layer.conf().setPretrain(true);
        layer.conf().setPretrain(false);
  It is allow pretrain and block pretrain.

1a) Allow pretrain is not good idea.
  If user doesn't allow pretrain, code  may not allow it. Not pretrain is not usual, but it is user option and/or problem.

1b) Block pretrain is not good idea.
  Pretraining usualy includes more epochs and it is never sure this epoch is the last.

2) Methods pretrainLayer should check if pretran is alowed. If is not allowed get info and return.

  I created method doPretrain() in NeuralNetConfiguration. This method allow this checking.

## How was this patch tested?

I checked manually code.

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
@@ -1135,4 +1135,11 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
         }
     }
 
+    /**
+     * If pretrain == true pretrain is allowed
+     * @return pretrain
+     */
+    public boolean doPretrain() {
+    	return pretrain;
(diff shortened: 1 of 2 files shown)


clavvis:
I have tried to remove MLN changes, but empty line + - left. How can I remove all file from commits?

AlexDBlack:
Actually, on reflection: this getter is redundant.
We have the lombok ```@Data``` annotation, which should add an ```isPretrain() : boolean``` method automatically.
https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java#L75

Issue #4271, opened 2017-11-08
Reduce memory requirements in BaseStatsListener


AlexDBlack:
After introducing workspaces, BaseStatsListener started copying gradients and activations, for calculation of stats.
This results in considerably greater memory requirements than if the various stats were calculated directly in the *onX(Model)* methods.

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/4292

PR #4292, opened 2017-11-13
Refactor BaseStatsListener to do some calculations in-line (requires less memory)

AlexDBlack:
Addresses: https://github.com/deeplearning4j/deeplearning4j/issues/4271


deeplearning4j-ui-parent/deeplearning4j-ui-model/src/main/java/org/deeplearning4j/ui/stats/BaseStatsListener.java
```diff
@@ -222,59 +228,71 @@ public void onEpochEnd(Model model) {
     @Override
     public void onForwardPass(Model model, List<INDArray> activations) {
         int iterCount = getModelInfo(model).iterCount;
-        if (storeActivations() && (iterCount == 0 || iterCount % updateConfig.reportingFrequency() == 0)) {
+        if (calcFromActivations() && (iterCount == 0 || iterCount % updateConfig.reportingFrequency() == 0)) {
```
crockpotveggies: Is this handled on a separate thread similar/same to as before?

AlexDBlack: No, it's handled in the current thread, same as before. It just does it when the arrays are available, rather than duping + calculating all at one later.

agibsonccc:
Anything stopping us from merging this?

crockpotveggies:
Nope...

Issue #4302, opened 2017-11-15
CompGraph: Array leaking out of workspace in computeGradientAndScore


AlexDBlack:
See test below. With training workspace enabled, the result is a JVM crash:
https://gist.github.com/AlexDBlack/bc761268f9115ab08ffaaf2130e317c3

```
public static void main(String[] args) throws Exception {
        ComputationGraph c = createResnet50();
//        c.getConfiguration().setTrainingWorkspaceMode(WorkspaceMode.NONE);
        c.getConfiguration().setTrainingWorkspaceMode(WorkspaceMode.SEPARATE);

        org.nd4j.linalg.dataset.DataSet ds = new org.nd4j.linalg.dataset.DataSet();
        ds.load(new File("dataSet.bin"));

        System.out.println("Train workspace: " + c.getConfiguration().getTrainingWorkspaceMode());
        System.out.println("Inference workspace: " + c.getConfiguration().getInferenceWorkspaceMode());

        c.setInputs(ds.getFeatures());
        c.setLabels(ds.getLabels());

        c.computeGradientAndScore();
    }



    public static ComputationGraph createResnet50() throws Exception {
        int labelSize = 5;

        ZooModel zooModel = new ResNet50(
                labelSize,
                12345,
                1,
                WorkspaceMode.NONE
        );

        Map<Integer,Double> lrsch = new LinkedHashMap<>();
        lrsch.put(0,0.0005);
        lrsch.put(500, 0.001);
        lrsch.put(1000, 0.0005);
        lrsch.put(2000, 0.0001);
        lrsch.put(4000, 0.00008);
        lrsch.put(6000, 0.00002);
        lrsch.put(8000, 0.00001);
        lrsch.put(10000, 0.000005);
        ISchedule lrSchedule = new MapSchedule(ScheduleType.ITERATION,lrsch);
        
            ComputationGraph zooModelOriginal = (ComputationGraph) zooModel.initPretrained(PretrainedType.IMAGENET);
            FineTuneConfiguration fineTuneConf = new FineTuneConfiguration.Builder()
                    .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
                    .l1(0.0)
                    .l2(0.0)
                    .updater(new Adam.Builder().learningRateSchedule(lrSchedule).build())
                    .seed(12345)
                    .trainingWorkspaceMode(WorkspaceMode.NONE)
[...]

PR #4320, opened 2017-11-21
#4302: Initialize workspace in computeGradientAndScore

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/4302


raver119:
Yes it fails, and that's fine. I'll pm.

PR #4336, opened 2017-11-28
Add ReverseTimeSeriesVertex to allow backward processing of time series

Broele:
We propose a new vertex type for usage with RNNs that reverts the order of time series.
This allows to backward process times series. 
We used this in one project to establish two stacks of layers - one for forward and one for backward processing - that were combined at the end of the computation pipeline.

## How was this patch tested?
- The vertex is used in one project and were tested manually and on base of the overall system output.
- Some changes were made to improve code quality for this pull request. These were only tested manually.

This is our first pull request to this project (hence a small class) and we are looking forward to some feedback.

### DIFF
--- a/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTestsComputationGraph.java
+++ b/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTestsComputationGraph.java
@@ -11,6 +11,7 @@ import org.deeplearning4j.nn.conf.distribution.UniformDistribution;
 import org.deeplearning4j.nn.conf.graph.*;
 import org.deeplearning4j.nn.conf.graph.rnn.DuplicateToTimeSeriesVertex;
 import org.deeplearning4j.nn.conf.graph.rnn.LastTimeStepVertex;
+import org.deeplearning4j.nn.conf.graph.rnn.ReverseTimeSeriesVertex;
 import org.deeplearning4j.nn.conf.inputs.InputType;
 import org.deeplearning4j.nn.conf.layers.*;
 import org.deeplearning4j.nn.conf.preprocessor.CnnToFeedForwardPreProcessor;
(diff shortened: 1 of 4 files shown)


AlexDBlack:
Thanks for the PR. I'd be happy to have this in DL4J, one it's ready.

Some initial thoughts:
1. We'll need gradient checks for this (to ensure correctness, we gradient check all layers and vertices): https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTestsComputationGraph.java
2. ```INDArray out = input.dup();``` can be ```INDArray out = Nd4j.create(input.shape(), 'f');``` (don't need to copy values, plus f array order for better memory striding with DL4J's time series layout)
3. To implement this most efficiently, we probably want to push this permute/reverse behaviour to the C++ layer, rather than the doubly-nested loops in Java. One way we can approximate this in the mean time is shown below (it's not optimal, but should be faster)

```
INDArray aTimeSeries = input.get(point(s), all(), interval(0,lastIdx));
INDArray reversed = Nd4j.pullRows(aTimeSeries, 1, new int[]{lastIdx-1, lastIdx-2, ..., 1, 0});
out.put(new INDArray[]{point(s), all(), interval(0,lastIdx)}, reversed)
```

Broele:
@AlexDBlack thank you for your reply.
About the issues:
1. I will look at the tests and see that I can add a gradient check.
2. I was not totally sure about masks the padding in inputs and dup() looked like the safe way. But looking at it again, create seems to be just fine.
3. Actually, I was looking for such a way. I tried interval(lastIdx, -1, firstIndex-1), which did not work, here.
I did not see pullRows. I will have a look for that.

The triple loops are actually for some edge cases where the mask has "holes", e.g. [1, 1, 0, 1, 0, 1].
Though this is not the typical case, I could imagine cases where we would have such a mask. 
This actually prevented me from using `BooleanIndexing` (or `INDArray::sum` - which I first thought of)

Do you think we should support this case?
In either way, I will try to move as much as possible to the C++ layer.

Broele:
@AlexDBlack thank you for merging. I also worked less than I intended on this.
I was thinking about an optimized reverse, but with the general case of masks with possible "holes" in mind (see above), this is complicated and I did not find a better way then these loops (at least without extending ND4j).
If you have some idea, I am happy to discuss and work on improvements.

Issue #4342, opened 2017-11-29
Pearson Correlation Coefficient confused with Coefficient of Determination


slang03:
#### Issue Description
Hi, just had a quick look at the `correlationR2` metric implementation [here](https://github.com/deeplearning4j/deeplearning4j/blob/cd9e03e26e4130d69d9a0ec52c1f8ed87252d0f2/deeplearning4j-nn/src/main/java/org/deeplearning4j/eval/RegressionEvaluation.java#L324).
This translates to the following formula (from [Wikipedia](https://en.wikipedia.org/wiki/Pearson_correlation_coefficient#For_a_sample)):

![Pearson Correlation Coefficient for Samples (Wikipedia)](https://wikimedia.org/api/rest_v1/media/math/render/svg/d5ff8bcff99d80ea955677080d97a00cd5fb6d3b)

In the [evaluation documentation](https://deeplearning4j.org/evaluation#Regression) it is described as *R^2 Coefficient of Determination* - which is wrong. The actual formula for the *Coefficient of Determination* is:

![](https://wikimedia.org/api/rest_v1/media/math/render/svg/fed29779d54adeccdec58f0894870c680f3d6b5b)
with

![](https://wikimedia.org/api/rest_v1/media/math/render/svg/aec2d91094ee54fbf0f7912d329706ff016ec1bd) 
and

![](https://wikimedia.org/api/rest_v1/media/math/render/svg/2669c9340581d55b274d3b8ea67a7deb2225510b)

respectively.

#### Possible Fix
Either
1. Change the documentation and description to *Pearson Correlation Coefficient for Samples*
2. Change the implementation to the correct formula of *Correlation of Determination*
3. Implement both methods `pearsonCorrelation` and `r2score` and document them accordingly

No matter which option will be chosen, there is going to be confusion for the users since their values which they have previously collected were calculated with the Pearson Correlation formula but they interpreted the score as R2 score (assuming they have read the documentation).

As soon as a decision is made I can make a PR for the fix.

#### Version Information
* Deeplearning4j version: 0.9.2-SNAPSHOT

AlexDBlack:
I guess seeing as we have pearson correlation already implemented, we might as well keep it. Furthermore, R2 is common enough that we will want to support it.
In short: option 3 is probably best. A PR for this would be great!

While we are at it: we should probably check both vs. say apache commons math, on some test data.

PR #4354, opened 2017-12-01
Add coefficient of determination metric

slang03:
Distinguish between Pearson Correlation and Coefficient of Determination as for the implementation of `correlationR2` actually calculated the Pearson Correlation. Additionally the `rSquared` method was added, which calculates the Coefficient of Determination.

Fixes: #4342

## How was this patch tested?
Test values were added in `RegressionEvalTest` and computed with [sklearn.metrics.r2_score](http://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html)


slang03:
Additionally, the documentation [here](https://github.com/deeplearning4j/deeplearning4j/blame/gh-pages/evaluation.md#L128) has to be updated. The values for the "shellfish" dataset have to be computed. Does anyone have a link to that dataset?

PR #4377, opened 2017-12-06
Add RnnLossLayer

AlexDBlack:
Should be good to go. Basically the RNN version of LossLayer.

Also removes an old loss function method deprecated in 0.6.0

### DIFF
--- a/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/OutputLayerGradientChecks.java
+++ b/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/OutputLayerGradientChecks.java
@@ -1,0 +1,119 @@
+package org.deeplearning4j.gradientcheck;
+
+import org.deeplearning4j.nn.conf.MultiLayerConfiguration;
+import org.deeplearning4j.nn.conf.NeuralNetConfiguration;
+import org.deeplearning4j.nn.conf.distribution.NormalDistribution;
+import org.deeplearning4j.nn.conf.layers.LSTM;
+import org.deeplearning4j.nn.conf.layers.RnnLossLayer;
+import org.deeplearning4j.nn.multilayer.MultiLayerNetwork;
+import org.deeplearning4j.nn.weights.WeightInit;
+import org.junit.Test;
+import org.nd4j.linalg.activations.Activation;
+import org.nd4j.linalg.api.buffer.DataBuffer;
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.api.ops.random.impl.BernoulliDistribution;
+import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.linalg.learning.config.NoOp;
+import org.nd4j.linalg.lossfunctions.ILossFunction;
+import org.nd4j.linalg.lossfunctions.impl.LossMCXENT;
+import org.nd4j.linalg.lossfunctions.impl.LossMSE;
+
+import java.util.Random;
+
+import static org.junit.Assert.assertTrue;
+
+public class OutputLayerGradientChecks {
+
+    private static final boolean PRINT_RESULTS = true;
+    private static final boolean RETURN_ON_FIRST_FAILURE = false;
+    private static final double DEFAULT_EPS = 1e-6;
+    private static final double DEFAULT_MAX_REL_ERROR = 1e-3;
+    private static final double DEFAULT_MIN_ABS_ERROR = 1e-8;
+
+    static {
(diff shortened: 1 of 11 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/RnnLossLayer.java
```diff
@@ -0,0 +1,130 @@
...
+import org.nd4j.linalg.lossfunctions.LossFunctions.LossFunction;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Recurrent Neural Network Loss Layer.<br>
+ * Handles calculation of gradients etc for various objective functions.<br>
+ * NOTE: Unlike {@link RnnOutputLayer} this RnnLossLayer does not have any parameters - i.e., there is no time
+ * distributed dense component here. Consequently, the output activations size is equail to the input size.<br>
```
maxpumperla: typo: equail -> equal

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/RnnLossLayer.java
```diff
@@ -0,0 +1,256 @@
...
+import org.nd4j.linalg.primitives.Pair;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Recurrent Neural Network Loss Layer.<br>
+ * Handles calculation of gradients etc for various objective functions.<br>
+ * NOTE: Unlike {@link RnnOutputLayer} this RnnLossLayer does not have any parameters - i.e., there is no time
+ * distributed dense component here. Consequently, the output activations size is equail to the input size.<br>
```
maxpumperla: equail

deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/OutputLayerTest.java
```diff
@@ -441,4 +445,81 @@ public void testRnnOutputLayerIncEdgeCases() {
...
+        int miniBatchSize = 3;
+
+        MultiLayerConfiguration conf1 =
+                new NeuralNetConfiguration.Builder().seed(12345L)
+                        .updater(new NoOp())
+                        .list()
+                        .layer(new LSTM.Builder().nIn(nIn).nOut(layerSize).activation(Activation.TANH)
+                                .weightInit(WeightInit.DISTRIBUTION).dist(new NormalDistribution(0, 1.0))
+                                .updater(new NoOp()).build())
+                        .layer(new DenseLayer.Builder().nIn(layerSize).nOut(nOut).activation(Activation.IDENTITY).build())
```
maxpumperla: @AlexDBlack I'm curious: going forward, do you think it would be possible to write something like this instead:

```
     .layer(new DenseLayer.Builder().nIn(layerSize).nOut(nOut).activation(Activation.SOFTMAX).build())
     .loss(new RnnLossLayer.Builder(LossFunction.MCXENT).build())
```
or even:
```
     .layer(new DenseLayer.Builder().nIn(layerSize).nOut(nOut).activation(Activation.SOFTMAX).build())
     .loss(LossFunction.MCXENT)
```
i.e. inferring loss layer type?

Sorry, I know this is vastly beyond the scope of this PR. Just want to discuss.

maxpumperla: btw, this test was very useful for understanding what you try to achieve here. 👍

deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/OutputLayerGradientChecks.java
```diff
@@ -0,0 +1,119 @@
...
+                for (int j = 0; j < timeSeriesLength; j++) {
+                    int idx = r.nextInt(nOut);
+                    labels.putScalar(new int[]{i, idx, j}, 1.0);
+                }
+            }
+
+            INDArray labelMask;
+            String mt;
+            switch (maskType){
+                case 0:
```
maxpumperla: 👍 very extensive setup, I like it

deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/OutputLayerTest.java
AlexDBlack: That's a good idea, open an issue? Not sure on the exact semantics/names, but something like that would be good.

PR #4396, opened 2017-12-12
Add bidirectional RNN wrapper layer

AlexDBlack:
Essentially the same as Keras' Bidirectional layer.

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/recurrent/Bidirectional.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/recurrent/Bidirectional.java
@@ -4,17 +4,24 @@ import lombok.Data;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
 import org.deeplearning4j.nn.api.ParamInitializer;
+import org.deeplearning4j.nn.api.layers.RecurrentLayer;
 import org.deeplearning4j.nn.conf.InputPreProcessor;
 import org.deeplearning4j.nn.conf.NeuralNetConfiguration;
 import org.deeplearning4j.nn.conf.inputs.InputType;
 import org.deeplearning4j.nn.conf.layers.BaseRecurrentLayer;
 import org.deeplearning4j.nn.conf.layers.Layer;
 import org.deeplearning4j.nn.conf.memory.LayerMemoryReport;
+import org.deeplearning4j.nn.layers.recurrent.BidirectionalLayer;
 import org.deeplearning4j.nn.params.BidirectionalParamInitializer;
 import org.deeplearning4j.optimize.api.IterationListener;
 import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.learning.config.IUpdater;
 
 import java.util.Collection;
+import java.util.Map;
+
+import static org.nd4j.linalg.indexing.NDArrayIndex.interval;
+import static org.nd4j.linalg.indexing.NDArrayIndex.point;
 
 @NoArgsConstructor
 @Data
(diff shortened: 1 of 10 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/recurrent/Bidirectional.java
```diff
@@ -0,0 +1,180 @@
...
+import static org.nd4j.linalg.indexing.NDArrayIndex.interval;
+import static org.nd4j.linalg.indexing.NDArrayIndex.point;
+
+/**
+ * Bidirectional is a "wrapper" layer: it wraps any uni-directional RNN layer to make it bidirectional.<br>
+ * Note that multiple different modes are supported - these specify how the activations should be combined from
+ * the forward and backward RNN networks. See {@link Mode} javadoc for more details.<br>
+ * Parameters are not shared here - there are 2 separate copies of the wrapped RNN layer, each with separate parameters.
+ * <br>
+ * Usage: {@code .layer(new Bidirecitonal(new LSTM.Builder()....build())}
```
maxpumperla: Bidirecitonal --> Bidirectional

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/recurrent/Bidirectional.java
```diff
@@ -0,0 +1,180 @@
...
+     * ADD: out = forward + backward (elementwise addition)<br>
+     * MUL: out = forward * backward (elementwise multiplication)<br>
+     * AVERAGE: out = 0.5 * (forward + backward)<br>
+     * CONCAT: Concatenate the activations.<br>
+     * Where 'forward' is the activations for the forward RNN, and 'backward' is the activations for the backward RNN.
+     * In all cases except CONCAT, the output activations size is the same size as the standard RNN that is being wrapped
+     * by this layer. In the CONCAT case, the output activations size (dimension 1) is 2x larger than the standard RNN's
+     * activations array.
+     */
+    public enum Mode {ADD, MUL, AVERAGE, CONCAT}
```
maxpumperla: just a remark: in keras you can also specify "NONE", which wouldn't combine the output at all, but rather return two outputs (as python list). I've personally never seen an application of this, but if it's not too difficult to achieve, might be good (esp. for keras import).

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/memory/LayerMemoryReport.java
```diff
@@ -137,6 +138,31 @@ public String toString() {
         return "LayerMemoryReport(layerName=" + layerName + ",layerType=" + layerType.getSimpleName() + ")";
     }
 
+    /**
+     * Multiply all memory usage by the specified scaling factor
```
maxpumperla: 👍

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/BidirectionalLayer.java
```diff
@@ -0,0 +1,546 @@
...
+import static org.nd4j.linalg.indexing.NDArrayIndex.interval;
+import static org.nd4j.linalg.indexing.NDArrayIndex.point;
+
+/**
+ * Bidirectional is a "wrapper" layer: it wraps any uni-directional RNN layer to make it bidirectional.<br>
+ * Note that multiple different modes are supported - these specify how the activations should be combined from
+ * the forward and backward RNN networks. See {@link Bidirectional.Mode} javadoc for more details.<br>
+ * Parameters are not shared here - there are 2 separate copies of the wrapped RNN layer, each with separate parameters.
+ * <br>
+ * Usage: {@code .layer(new Bidirecitonal(new LSTM.Builder()....build())}
```
maxpumperla: same typo

AlexDBlack:
@maxpumperla Should be fixed. Can't re-test DL4J yet (see gitter channel)
As for 'NONE' mode (multi-output): that'll have to wait for our big PR (that adds multi in/out support) before we can add that.

AlexDBlack:
@maxpumperla I'll close this - both the bidirectional and simple RNN layers are here: https://github.com/deeplearning4j/deeplearning4j/pull/4398
(I rebased the simple RNN on this one as there was a bit of overlap)

Issue #3482, opened 2017-06-05
Evaluation on long time series: use split feed-forward mechanism


AlexDBlack:
Example: a time series of length 10k.
You might train with TBPTT length of 100 - but then at evaluation time, we just naively feed-forward the whole time series. In the 10k/100 example, that means activation memory requirements are 100x higher at test time vs. at train time.

Proposal:
MultiLayerNetwork/ComputationGraph.doEvaluation() automatically uses a similar "partial/segmented time series feed-forward" mechanism, similar to TBPTT, for long time series.
This may be as simple as doing it using the TBPTT lengths, and only when TBPTT is enabled (if we can *train* without requiring this, then testing without it should also be possible).

AlexDBlack:
Added here: https://github.com/deeplearning4j/deeplearning4j/pull/4405

PR #4405, opened 2017-12-13
Make RNN evaluation more memory efficient when using TBPTT

AlexDBlack:
Addresses: https://github.com/deeplearning4j/deeplearning4j/issues/3482


AlexDBlack:
@raver119 thanks, and not yet. I'll run some tests.

PR #4410, opened 2017-12-13
Support bidirectional layer wrapper keras import

maxpumperla:
Support import of Bidirectional layer wrapper from keras. Pretty straightforward overall, but this test still fails:

https://github.com/deeplearning4j/deeplearning4j/blob/c7b15db6444235a1f3d7a3f45485278ccf4ececd/deeplearning4j-modelimport/src/test/java/org/deeplearning4j/nn/modelimport/keras/configurations/Keras2ModelConfigurationTest.java#L42

which boils down to this guy not being properly initialized:

https://github.com/deeplearning4j/deeplearning4j/blob/0d9f7dab18c99249c76fefacb957e60035e1ade2/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/recurrent/Bidirectional.java#L100-L101

namely throwing:
```
org.deeplearning4j.exception.DL4JInvalidConfigException: LSTM (index=1, name=layer1) nIn=0, nOut=64; nIn and nOut must be > 0

	at org.deeplearning4j.nn.conf.layers.LayerValidation.assertNInNOutSet(LayerValidation.java:37)
	at org.deeplearning4j.nn.conf.layers.LSTM.instantiate(LSTM.java:79)
	at org.deeplearning4j.nn.conf.layers.recurrent.Bidirectional.instantiate(Bidirectional.java:101)
	at org.deeplearning4j.nn.multilayer.MultiLayerNetwork.init(MultiLayerNetwork.java:629)
	at org.deeplearning4j.nn.multilayer.MultiLayerNetwork.init(MultiLayerNetwork.java:548)
	at org.deeplearning4j.nn.modelimport.keras.configurations.Keras2ModelConfigurationTest.runSequentialConfigTest(Keras2ModelConfigurationTest.java:141)
	at org.deeplearning4j.nn.modelimport.keras.configurations.Keras2ModelConfigurationTest.bidirectionalLstmConfigTest(Keras2ModelConfigurationTest.java:43)
```

@AlexDBlack it seems that although `fwd` and `bwd` are cloned from the _same_ configuration `conf`, the latter has `nOut=0` and the former `nOut=128`. I checked all other properties, they're equal except for `nOut`. Very strange. Ideas?

### DIFF
--- a/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/config/KerasLayerConfiguration.java
+++ b/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/config/KerasLayerConfiguration.java
@@ -72,7 +72,7 @@ public class KerasLayerConfiguration {
     private final String LAYER_CLASS_NAME_UPSAMPLING_2D = "UpSampling2D";
     private final String LAYER_CLASS_NAME_SEPARABLE_CONVOLUTION_2D = ""; // 1: SeparableConvolution2D, 2: SeparableConv2D
     private final String LAYER_CLASS_NAME_DECONVOLUTION_2D = ""; // 1: Deconvolution2D, 2: Conv2DTranspose
-
+    private final String LAYER_CLASS_NAME_BIDIRECTIONAL = "Bidirectional";
 
     /* Partially shared layer configurations. */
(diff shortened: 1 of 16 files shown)


AlexDBlack:
@maxpumperla Only thing I can think of is maybe clone is broken... happy to run this in the debugger if you want me to take a look.

PR #4447, opened 2017-12-21
Add .travis.yml and build script for Linux

saudet:

Let Travis CI make builds and deploy SNAPSHOT artifacts to Sonatype OSS automatically.

## How was this patch tested?

Builds pass on Travis CI and SNAPSHOT artifacts are deployed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+    - master
 dist: trusty
 sudo: false
 cache:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,23 @@
+dist: trusty
```
agibsonccc: Same concern re: distro.

saudet: No choice, only thing supported by Travis.

PR #4457, opened 2018-01-02
keras 2 merge layers

maxpumperla:
Simply adding explicit keras 2 merge layers, e.g. `Add` vs. `Merge(mode='add')`.

### DIFF
--- a/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasLayerUtils.java
+++ b/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasLayerUtils.java
@@ -18,6 +18,7 @@
 package org.deeplearning4j.nn.modelimport.keras.utils;
 
 import lombok.extern.slf4j.Slf4j;
+import org.deeplearning4j.nn.conf.graph.ElementWiseVertex;
 import org.deeplearning4j.nn.conf.layers.Layer;
 import org.deeplearning4j.nn.modelimport.keras.KerasLayer;
 import org.deeplearning4j.nn.modelimport.keras.config.KerasLayerConfiguration;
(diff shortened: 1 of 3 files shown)


deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasLayerUtils.java
```diff
@@ -225,6 +226,24 @@ public static KerasLayer getKerasLayerFromConfig(Map<String, Object> layerConfig
...
+                layerClassName.equals(conf.getLAYER_CLASS_NAME_FUNCTIONAL_SUBTRACT())) {
+            layer = new KerasMerge(layerConfig, ElementWiseVertex.Op.Subtract, enforceTrainingConfig);
+        } else if (layerClassName.equals(conf.getLAYER_CLASS_NAME_AVERAGE()) ||
+                layerClassName.equals(conf.getLAYER_CLASS_NAME_FUNCTIONAL_AVERAGE())) {
+            layer = new KerasMerge(layerConfig, ElementWiseVertex.Op.Average, enforceTrainingConfig);
+        } else if (layerClassName.equals(conf.getLAYER_CLASS_NAME_MULTIPLY()) ||
+                layerClassName.equals(conf.getLAYER_CLASS_NAME_FUNCTIONAL_MULTIPLY())) {
+            layer = new KerasMerge(layerConfig, ElementWiseVertex.Op.Product, enforceTrainingConfig);
+        } else if (layerClassName.equals(conf.getLAYER_CLASS_NAME_ADD()) ||
+                layerClassName.equals(conf.getLAYER_CLASS_NAME_ADD())) {
```
AlexDBlack: Isn't the add case already handled earlier?

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasLayerUtils.java
```diff
@@ -225,6 +226,24 @@ public static KerasLayer getKerasLayerFromConfig(Map<String, Object> layerConfig
...
+            layer = new KerasMerge(layerConfig, ElementWiseVertex.Op.Add, enforceTrainingConfig);
+        } else if (layerClassName.equals(conf.getLAYER_CLASS_NAME_DOT()) ||
+                layerClassName.equals(conf.getLAYER_CLASS_NAME_FUNCTIONAL_DOT())) {
+            layer = new KerasMerge(layerConfig, null, enforceTrainingConfig);
```
AlexDBlack: If mergeMode is null, wan't this use MergeVertex, and not implement dot?

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasLayerUtils.java
maxpumperla: yes, thanks!

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasLayerUtils.java
maxpumperla: yep, that should be concatenate... :/

PR #4472, opened 2018-01-04
DataSetIterator splitters

raver119:
***WIP; DO NOT MERGE;***

This PR adds abstraction to provide virtual DataSetIterator splitter into test/train parts.
And tests.


raver119:
@AlexDBlack i've added simple check for equality for first elements between epochs. How about this approach? :)

PR #4491, opened 2018-01-09
Fixed support of loading the binary Google model

mossaab0:
After reading the word String and the floats, we need to skip the return line. Otherwise that return line is read as a new empty String, crashing the loader.


raver119:
I'm afraid this PR should be reverted.

All you do here is mimic of resetLine=true option. So, there comes CR/LF difference between operating systems. I'll file an issue, and investigate this change.

PR #4499, opened 2018-01-11
Add command line flags for file based stats storage.

crockpotveggies:

Add custom flags for setting StatsFileStorage and custom path if necessary.

## How was this patch tested?

Since this is UI-related, will need to rebuild JAR and deploy on a cluster to actually test effectiveness.

### DIFF
--- a/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java
+++ b/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java
@@ -22,6 +22,7 @@ import org.deeplearning4j.ui.module.tsne.TsneModule;
 import org.deeplearning4j.ui.play.misc.FunctionUtil;
 import org.deeplearning4j.ui.play.staticroutes.Assets;
 import org.deeplearning4j.ui.play.staticroutes.I18NRoute;
+import org.deeplearning4j.ui.storage.FileStatsStorage;
 import org.deeplearning4j.ui.storage.InMemoryStatsStorage;
 import org.deeplearning4j.ui.storage.impl.QueueStatsStorageListener;
 import org.nd4j.linalg.primitives.Pair;
@@ -32,6 +33,7 @@ import play.api.routing.Router;
 import play.routing.RoutingDsl;
 import play.server.Server;
 
+import java.io.File;
 import java.util.*;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java
```diff
@@ -79,13 +81,19 @@
...
 
     private Thread uiEventRoutingThread;
-    @Parameter(names = {"-r", "-enableRemote"}, description = "Whether to enable remote or not", arity = 1)
-    private boolean enableRemote;
 
+    @Parameter(names = {"-r", "--enableRemote"}, description = "Whether to enable remote or not", arity = 1)
+    private boolean enableRemote;
 
-    @Parameter(names = {"--uiPort"}, description = "Whether to enable remote or not", arity = 1)
+    @Parameter(names = {"-p", "--uiPort"}, description = "Whether to enable remote or not", arity = 1)
```
AlexDBlack: Was already wrong - but let's fix this description?

deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java
```diff
@@ -79,13 +81,19 @@
...
     private int port = DEFAULT_UI_PORT;
 
+    @Parameter(names = {"-fs", "--enableFileStorage"}, description = "Use file-based stats storage", arity = 1)
+    private boolean enableFileStorage;
+
+    @Parameter(names = {"--customStatsFile"}, description = "Path to create custom stats file", arity = 1)
```
AlexDBlack: arity=1 I think is only valid for boolean args. Also improve description - " for remote listener support. If set, also enables remote listener."

deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java
```diff
@@ -222,7 +230,20 @@ public void runMain(String[] args) {
...
+                File tempStatsFile = File.createTempFile("dl4j", "UIstats");
+                tempStatsFile.deleteOnExit();
+                enableRemoteListener(new FileStatsStorage(tempStatsFile), true);
+            } catch(Exception e) {
+                System.out.println("Failed to create temporary file for stats storage");
+                System.exit(0);
+            }
+        }
+        else if(enableRemote)
             enableRemoteListener();
```
AlexDBlack: On second thought - let's merge enableRemote and enableFileStorage - i.e., default to FileStorage only (temp dir unless path is set). File has better performance and less memory use...

deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java
crockpotveggies: incorrect, the javadoc says

> How many parameter values this parameter will consume. For example, an arity of 2 will allow "-pair value1 value2".

PR #4508, opened 2018-01-16
Standalone module for UI server

crockpotveggies:

Shade for standalone JAR on UI remote instances.

## How was this patch tested?

In progress.


AlexDBlack:
Looks like this is essentially good to go:
image

Two known issues:
1. There seems to be a character encoding issue with the internationalization (Japanese etc doesn't render properly, but the contents in the JAR is still correct, when viewed as utf-8)
2. The logs default to <jar_location>/logs/ directory... if running from say maven, this may not be ideal as it'll be in the users .m2 directory.

crockpotveggies:
Thank you very much!

PR #4514, opened 2018-01-17
Keras import: a few tiny end to end tests to check proper weight setting

maxpumperla:
This in particular fixes the problems we seemed to be having with keras 2 and theano, thanks to @tomthetrainer for pointing this out.

@AlexDBlack: there's one e2e test I had to relax, namely `importMnistCnnThKeras1()` below. This fails on master too, I checked. Do you know if something changed underlying our CNN impl? Other than that I can't explain why it suddenly fails. Thanks for your feedback in any case.


AlexDBlack:
@maxpumperla yes, there were recent changes to our CNN implementation. DM me for details.

tomthetrainer:
@maxpumperla I could run a few tests, I have lots of Keras Models laying around, 

When will this be in a snapshot, or would I have to build from source ?

AlexDBlack:
@tomthetrainer snapshots are up currently, so it *should* be in there by now

PR #4524, opened 2018-01-19
Update to JavaCPP 1.4 and add support for CUDA 9.1

saudet:

Update to the latest release versions of JavaCPP and CUDA.

## How was this patch tested?

Unit tests pass, but let's wait after the CI server for the build scripts...


raver119:
TL/DR: we'll support 3 CUDA versions for upcoming release: 8.0, 9.0 and 9.1

saudet:
@raver119 Maybe, but it doesn't cost anything to leave 3 in the source code for now :)

PR #4538, opened 2018-01-23
ActivationLayer: deal with different mask ranks

Broele:
Using an activation layer on TimeSeries Data fails when using masks.
The layer was modified to deal with different masks (rank 1, 2 or 3) and different inputs (rank 2 or 3)

## How was this patch tested?
I tested it with a small example for time series data:
https://gist.github.com/Broele/c704b46715da1a6aec1ffdc472a6b311


AlexDBlack:
Thanks

PR #4550, opened 2018-01-25
add feed forward to layer for comp graph

agibsonccc:

Adds feed forward to layer for comp graph.
Test:
https://github.com/deeplearning4j/deeplearning4j/blob/6a6d5044419a6b6cea36333ef7d8f065eae46cfe/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/graph/TestComputationGraphNetwork.java#L80
## How was this patch tested?

## Quick checklist

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
@@ -1576,6 +1576,14 @@ public class ComputationGraph implements Serializable, Model, NeuralNetwork {
 
         //Do forward pass according to the topological ordering of the network
         for (int i = 0; i < topologicalOrder.length; i++) {
+/*
+            if(layerFeedForwardIdx >= 0 && i > layerFeedForwardIdx + 1) {
+                break;
+            }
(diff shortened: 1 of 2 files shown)


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
```diff
@@ -1488,6 +1576,12 @@ public void computeGradientAndScore() {
 
         //Do forward pass according to the topological ordering of the network
         for (int i = 0; i < topologicalOrder.length; i++) {
+            if(layerFeedForwardIdx >= 0 && i > layerFeedForwardIdx + 1) {
+                break;
```
AlexDBlack: I'm not sure this is guaranteed to be correct in all possible cases.
Specifically, suppose we have 4 vertices - A,B,C,D... and we want to feed forward to vertex C with index 2.
Though in most cases it will be fine, I don't think we have a guarantee that C's forward pass will be done by i=(2+1).

A better condition: at the very end of the for(int i=0...) loop, do something like this:
```
if(current.getVertexIndex() == layerFeedForwardIdx) break;
```

PR #4581, opened 2018-01-31
set output even though an exception happens

jerry-sjtu:
## The InferenceWorker will set the output even though an exception happens

When an exception happened when executing the statement `output = ((MultiLayerNetwork) replicatedModel).output(request.getInput()[0])` ,  the `InferenceWorker` will not call  the method  `request.setOutput(output)`. Then `observer.waitTillDone()` will wait all the time.

## How was this patch tested?

This code was tested manually.

## Quick checklist


raver119:
I don't think it's good. I think we should propagate exception to the upper level. 
Because in some cases exception thrown on native level followed by java exception means: process is dead, and should be reinstantiated. Especially in case of CUDA.

raver119:
I mean - the problem you've spotted is real one, thanks for highlighting it. But it's not a fix, it's attempt to mask problem. And that's very bad for CUDA use case at very least.

agibsonccc:
Closing due to inactivity.

PR #4599, opened 2018-02-02
Fix for obtaining own IP by master and nodes in shared training

tomkozicki:

Parameter Server master now translates SPARK_PUBLIC_DNS to IP if it's just hostname.

## How was this patch tested?

Ran MNIST parameter server example after building this and analogous nd4j change with expected result


tomkozicki:
Should be considered along with:
https://github.com/deeplearning4j/nd4j/pull/2594/commits/d6bc465626bdd4d9b8a1354fc23534be078ce23f

PR #4604, opened 2018-02-02
demo test case re: original/restored model output difference

cpoerschke:
I locally observed differences (curiously not all of the time) between original model output and restored model output. This pull request is intended to demonstrate/reproduce the issue. The model is admittedly trivial and perhaps it's setup/usage is incomplete/incorrect (I am new to deeplearning4j) but even then it's odd that the issue happens only part of the time?

### DIFF
--- a/deeplearning4j-modelexport-solr/pom.xml
+++ b/deeplearning4j-modelexport-solr/pom.xml
@@ -1,0 +1,118 @@
+<!--
+  ~ /*
+  ~  * Copyright 2016 Skymind,Inc.
+  ~  *
+  ~  *    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


agibsonccc:
@cpoerschke could you clarify a bit? What was the exact problem? You should probably file a specific issue first so we can clarify. Appreciate the pull request though!

AlexDBlack:
Thanks for the pull request - at first glance it looks like you've identified a legitimate issue / edge case.
I'll close this PR in favor of this issue - https://github.com/deeplearning4j/deeplearning4j/issues/4609 - and I will push up a modified version of the test here once it's fixed.

For future reference, you can just open an issue with a unit test, rather than going to the trouble of a full PR.

Issue #4639, opened 2018-02-09
"CUDA 8" maven profile is actually CUDA 9


AlexDBlack:
This should be renamed to avoid confusion. Perhaps a CUDA 8 profile should also be added, unless we standardize on CUDA 8 only?

https://github.com/deeplearning4j/deeplearning4j/blob/master/pom.xml#L383-L404

saudet:
Or perhaps we should have the change-cuda-versions.sh script change the name of the profile as well?

AlexDBlack:
That's... an interesting idea. A simple (similar) alternative is just changing it to say test-nd4j-cuda, and yeah, letting the change cuda version script do the switching. Single profile might make it simpler for CI if/when we test different CUDA versions, vs. "rename the profile"

saudet:
Not sure it would make it simpler for CI to have no versions in the profile, @sshepel ? It'd certainly make it easier to test the wrong version of CUDA though.

sshepel:
> Not sure it would make it simpler for CI to have no versions in the profile, @sshepel ? It'd certainly make it easier to test the wrong version of CUDA though.

@saudet you are right, I would say just to fix profile name and not remove the version in it, "Explicit is better than implicit."

PR #4649, opened 2018-02-12
Add CUDA 9 test profile

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/4639

Changes test-nd4j-cuda-8.0 to to actually be CUDA 8 (not CUDA 9)
Also adds a test-nd4j-cuda-9.0 profile.

Tested on some of the tests, runs OK. ```change-cuda-versions.sh``` script seems to work correctly.

### DIFF
--- a/change-cuda-versions.sh
+++ b/change-cuda-versions.sh
@@ -82,6 +82,14 @@ echo "Updating CUDA versions in pom.xml files to CUDA $1";
 
 BASEDIR=$(dirname $0)
 
+#Artifact ids, ending with "-8.0", "-9.0", etc. nd4j-cuda, etc.
+find "$BASEDIR" -name 'pom.xml' -not -path '*target*' \
+  -exec bash -c "sed_i 's/\(artifactId>nd4j-cuda-\)...<\/artifactId>/\1'$VERSION'<\/artifactId>/g' {}" \;
+
+#Profiles ids, ending with "-8.0", "-9.0", etc. test-nd4j-cuda, etc.
+find "$BASEDIR" -name 'pom.xml' -not -path '*target*' \
+  -exec bash -c "sed_i 's/\(test-nd4j-cuda-\)...</\1'$VERSION'</g' {}" \;
+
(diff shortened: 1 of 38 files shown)


saudet:
The problem with this is that the script will change the CUDA version in both profiles. My suggestion would be to change the version of the profile with the script too, such that if we're building for CUDA 8.0, test-nd4j-cuda-9.0 would not exist at all and would cause an error if the user tried to use it. That sounds more robust and comes with less maintenance work, but you feel that this isn't right?

AlexDBlack:
> The problem with this is that the script will change the CUDA version in both profiles. 

I did already test this... 
> change-cuda-versions.sh script seems to work correctly.

See for yourself:
https://gist.github.com/AlexDBlack/8dfd47f434726f7db7a95f6002b0e87e

Unless I'm missing something here?

saudet:
Ok, I guess the script doesn't change the version of CUDA in the profile. What if it did change the version of CUDA in the profile, and the name of the profile as well?

AlexDBlack:
That's an alternative. Though I figured this approach was more in line with @sshepel 's comment in the issue:
> I would say just to fix profile name and not remove the version in it, "Explicit is better than implicit."

I don't care much either way... @sshepel do you have a preference?

saudet:
I think that's what he meant as well. Testing an artifact that isn't being built doesn't make much sense IMO.

saudet:
Ok, I've "reverted" this PR to put my suggested changes on top (but we can easily "rerevert"...) 
@sshepel, which one shall we pick?

sshepel:
> Ok, I've "reverted" this PR to put my suggested changes on top (but we can easily "rerevert"...)
> @sshepel, which one shall we pick?

> I don't care much either way... @sshepel do you have a preference?

Well, I would rather agree with @saudet, and stick to one tag of CUDA version that could be changed depending on needs.

Lets try current implementation and see how it goes.

PR #4660, opened 2018-02-15
[WIP] Mask zero activations for RNNs

mboyanov:
The proposed wrapper looks at the activations of the input and sets the mask to zero for a given timestep if the input activation is zero. 
The idea is to replicate the mask_zero=True behaviour emerging from the Keras Embedding layer, but it should also be applicable in DL4J as a whole.

 
## How was this patch tested?
For now, it has only been tested manually by importing a keras embedding -> keras lstm -> output model and checking in the debugger that the hidden memory state is not updated for the zero timesteps.

## Quick checklist

Not yet
Haven't run them yet


mboyanov:
Thanks for the feedback. 
I implemented a basic unit test when calling instantiate manually. I'd like to extend the functionality to work with `ComputationGraph` and I'm wondering what would be the better approach to creating the corresponding entries in the conf package: to extend `GraphVertex` or to extend `Layer`?

AlexDBlack:
> I'd like to extend the functionality to work with ComputationGraph and I'm wondering what would be the better approach to creating the corresponding entries in the conf package: to extend GraphVertex or to extend Layer?

There's no need to make any changes for ComputationGraph... you can use the layer as-is there.

AlexDBlack:
I've merged this - looks good now. Thanks!

PR #4675, opened 2018-02-19
Add samediff layer support (forward only, no output layers)

AlexDBlack:
Starting from: https://github.com/deeplearning4j/deeplearning4j/pull/4426 - removes the backprop components and output/loss layers... I'll tackle those components separately.

Adds support for SameDiff layers:
- Forward pass only for now, non-output layers only. Backprop is partially implemented, but disabled for now
- I've tried to keep everything to a *single class only*: that means, implementing a SameDiff layer requires *only* extending BaseSameDiffLayer, and that's it (no implementation or param initializer classes, for example).
    - 2 complete examples of samediff layers can be found in this PR (both in test directory) - SameDiffConv and SameDiffDense. Param initialization and forward pass match the DL4J layers.
    - See also MinimalSameDiffDense for a minimal version (least number of methods required to be implemented)


AlexDBlack:
Something we will want to support: importing a layer definition (serialized SameDiff, TF, etc) into a DL4J layer. The current layer API may not easily support this in all cases.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/samediff/BaseSameDiffLayer.java
```diff
@@ -0,0 +1,67 @@
...
+    protected BaseSameDiffLayer(Builder builder){
+        super(builder);
+        this.weightInit = builder.weightInit;
+    }
+
+    protected BaseSameDiffLayer(){
+        //No op constructor for Jackson
+    }
+
+    public abstract List<String> defineLayer(SameDiff sameDiff, SDVariable layerInput, Map<String,SDVariable> paramTable);
```
wmeddie: I felt this would be better defined as a List<SDVariable> instead of List<String>.  That way you aren't just throwing away objects, and it's impossible to make a typo.

wmeddie: Here's an idea oh how it could work (and a working unit test)

https://gist.github.com/wmeddie/809c602311503a2ebb2d2cd957daf09e

Issue #4692, opened 2018-02-22
The method humanReadableInt in StringUtils.java has a race condition.


cragkhit:
#### Issue Description

Hi. I found that the `humanReadableInt(long number)` method in the `deeplearning4j-nn/src/main/java/org/deeplearning4j/util/StringUtils.java` contains code which has a race condition as shown in Hadoop (issue tracking ID [HADOOP-9252](https://issues.apache.org/jira/browse/HADOOP-9252)). I couldn't find a call to the method anywhere else in the code. But it might be worth to fix.

#### Version Information

I looked at the latest version of the code.

#### Contributing

In Hadoop, they changed this method in their [StringUtils.java](https://github.com/apache/hadoop/blob/532bbf4602e418276961cfc4f6b5e3f70e9cedc0/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/util/StringUtils.java) to call `TraditionalBinaryPrefix.long2String(number, "", 1);` instead.
```java
  /**
   * Given an integer, return a string that is in an approximate, but human 
   * readable format. 
   * @param number the number to format
   * @return a human readable form of the integer
   *
   * @deprecated use {@link TraditionalBinaryPrefix#long2String(long, String, int)}.
   */
  @Deprecated
  public static String humanReadableInt(long number) {
    return TraditionalBinaryPrefix.long2String(number, "", 1);
  }
```

saudet:
Sounds good! Please send a pull request if possible.

AlexDBlack:
Fixed here: https://github.com/deeplearning4j/deeplearning4j/pull/4725

PR #4710, opened 2018-02-24
Fixed a race condition in StringUtils.java 

cragkhit:
Hi. I found that the `humanReadableInt(long number)` method in the `deeplearning4j-nn/src/main/java/org/deeplearning4j/util/StringUtils.java` contains code which has a race condition as shown in Hadoop (issue tracking ID [HADOOP-9252](https://issues.apache.org/jira/browse/HADOOP-9252)). I couldn't find a call to the method anywhere else in the code. But it might be worth to fix.

This pull request is linked to issue [#4692](https://github.com/deeplearning4j/deeplearning4j/issues/4692).

I updated the `humanReadableInt` method by following the fix in [HADOOP-9252](https://issues.apache.org/jira/browse/HADOOP-9252). 

## How was this patch tested?

I added a test class, `TestStringUtils.java` adopted from Apache Hadoop code. It tests the functionality of the newly introduced `TraditionalBinaryPrefix` enum (which is called by `humanReadableInt`).

## Quick checklist

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/StringUtils.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/StringUtils.java
@@ -75,27 +75,19 @@ public class StringUtils {
     /**
      * Given an integer, return a string that is in an approximate, but human
      * readable format.
-     * It uses the bases 'k', 'm', and 'g' for 1024, 1024**2, and 1024**3.
      * @param number the number to format
      * @return a human readable form of the integer
+     *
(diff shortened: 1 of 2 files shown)


saudet:
LGTM, but if it's not used anywhere in the code, should we just remove this? @AlexDBlack @agibsonccc

wmeddie:
I've used the DL4J StringUtils in my projects.  Even if it isn't used in our code we probably shouldn't remove it if it was in a release.

AlexDBlack:
Closing this - thanks for the PR and flagging the issue, but I'll do some refactoring and cleanup (plus fix license headers) separately to this, that will handle the reported issue.

PR #4716, opened 2018-02-26
[WIP] Zero masking model import

mboyanov:

The idea is to support the mask_zero parameter in the Keras embedding layer. 
When a new `KerasLSTM` layer is constructed, the constructor checks the inbound layers and if any of them are a `KerasEmbedding` with `mask_zero=True`, it applies the mask zero wrapper.
This functionality should still work with `return_sequences = False`, so sometimes we get double wrappers. 
## How was this patch tested?
Unit test for the constructor functionality.
Experiments with importing trained keras models for the rest. 

## WIP
I'd like to run a few more imports to make sure it works as expected. Also, I'd like to change the warning that `mask_zero` is not supported.


ickyfariz:
Icky Fariz

maxpumperla:
@mboyanov this looks very promising, thanks a lot! I'll do a full review once you're done. just ping me (I think you know there's still stuff missing).

passing previous layers into a Keras layer constructor is a good idea. we still have some minor issues with Embedding layers in other scenarios as well. 👍

mboyanov:
@maxpumperla I've added a toy json configuration in a unit test and a unit test to check the `KerasEmbedding` functionality, but I may still be missing something?
Looking forward to any feedback

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/embeddings/KerasEmbedding.java
```diff
@@ -66,12 +76,10 @@ public KerasEmbedding(Map<String, Object> layerConfig, boolean enforceTrainingCo
...
 
-        boolean hasZeroMasking = KerasLayerUtils.getZeroMaskingFromConfig(layerConfig, conf);
+        this.hasZeroMasking = KerasLayerUtils.getZeroMaskingFromConfig(layerConfig, conf);
         if (hasZeroMasking)
-            log.warn("Masking in keras and DL4J work differently. We do not support mask_zero flag" +
-                    "on Embedding layers. If you want to have this behaviour for your imported model" +
-                    "in DL4J, apply masking as a pre-processing step to your input." +
-                    "See https://deeplearning4j.org/usingrnns#masking for more on this.");
+            log.warn("Zero Masking for the Embedding layer only works with unidirectional LSTM for now."
+                    + "For Bidirectional layers, you can set the mask yourself.");
```
maxpumperla: Could we maybe put the last two rows of the old docs back in? Please be a little more explicit what you mean by setting it yourself.

maxpumperla:
@mboyanov really well done, valuable addition.

mboyanov:
Glad you like it :)
I improved the documentation a bit, but it might still need some more clarification. I removed the "setting the mask yourself" bit because I think the docs can explain it in detail.

Can you be a bit more specific on the test checking that masking for the LSTM works as well? There is a  test which checks if  `kerasLstm.getLayer() instanceof MaskZeroLayer`

maxpumperla:
@mboyanov cool, you're right! going to merge this now, thanks for your contribution.

PR #4723, opened 2018-02-27
Polish: constant names should comply with a naming convention

AlexDBlack:
Starts from following PR but updates to fix merge conflict: https://github.com/deeplearning4j/deeplearning4j/pull/4654


saudet:
BTW, GitHub now allows us to push commits to other people's forks by default, so you could simply have modified pull #4654.

AlexDBlack:
Ah, thanks @saudet - good to know. I wasn't aware of that.

PR #4739, opened 2018-02-28
[WIP] SpaceToDepth test fix

maxpumperla:
@AlexDBlack still don't get why the test failed. I've simplified the setup a little. when I have a conv layer right before STD, the test fails. I'm pretty sure the backward pass for STD is ok, though. the respective grad check test in ND4J passes, too.

Ideas?


agibsonccc:
Can we get this merged before release?

maxpumperla:
will investigate again and open a new pr

Issue #4736, opened 2018-02-28
Failing model zoo unit tests - CUDA


AlexDBlack:
CPU: All passing (occasional memory issues aside)

CUDA:
![image](https://user-images.githubusercontent.com/2360237/36774806-9e638744-1cb3-11e8-9fd8-075cf579a555.png)

testImageNetLabels: expected golden retriever, got:
```
o.d.z.TestImageNet - Predictions for batch  :
	0.605528%, bassinet
	0.589057%, tub
	0.565464%, paper_towel
	0.555924%, cradle
	0.526733%, book_jacket
```
CPU assigns 0.9751 probability to "golden retriever"
GPU assigns 0.0013 probability to "golden retriever"

Full probablities from each backend: https://gist.github.com/AlexDBlack/c3cf0404e202974fa2c800899fcfc164

testDarknetLabels:
```
org.junit.ComparisonFailure: 
Expected :golden retriever
Actual   :torch
```

AlexDBlack:
Looks like a bug in first subsampling layer - I'm getting a bunch of 0s for some of the activations during CUDA forward pass... but not all of them. Before that, activations are pretty close for both backends.

AlexDBlack:
Confirmed fixed here: https://github.com/deeplearning4j/libnd4j/pull/818

PR #4758, opened 2018-03-05
Add full YOLO2 (COCO) to the zoo along with a DarknetHelper class

saudet:

Also added a cudnnAlgoMode property to NeuralNetConfiguration so that we can set the default value for all ConvolutionLayer more easily.

## How was this patch tested?

The updated unit tests pass with nd4j-native. (nd4j-cuda is currently broken, see issue #4736.)


crockpotveggies:
Is the codeclimate check trustworthy yet? If no, I'll merge this.

saudet:
Well, check out what it says by clicking on it. It's all really superficial IMO, but do you care about some of them?

Issue #4781, opened 2018-03-09
Improper handling of InterruptedExecption


jentfoo:
I noticed in `MultiThreadUtils` that you are catching an InterruptedException but not reseting the status.

https://github.com/deeplearning4j/deeplearning4j/blob/ec2ed9e71f82ccce1c4f16dba9f339c0b61a27f9/deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/util/MultiThreadUtils.java#L69

By catching the exception the status is cleared, you should re-interrupt the thread before throwing the wrapped exception so that other potentially blocking tasks wont block and prevent things like jvm shutdown.

You may want to reference `Don't swallow interrupts` here: https://www.ibm.com/developerworks/library/j-jtp05236/

saudet:
Sounds good! Could you send a pull request to fix this?

Issue #4766, opened 2018-03-07
KMeansClustering cause thread leak


gogo111007:
#### Issue Description

i use KMeansClustering like KMeansTest.java
every client request, i get new kmeans object by KMeansClustering.setup()
but there is no way to destroy threads allocate by KMeansClustering 
after thousands or more request, too many threads cause crash

#### Version Information

* deeplearning4j ver newest
* windows 7 and centos 6.x

jentfoo:
I recently stumbled across this too.  The issue is in `BaseClusteringAlgorithm`.  It constructs an `ExecutorService` but provides no way to free it.

https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-nearestneighbors-parent/nearestneighbor-core/src/main/java/org/deeplearning4j/clustering/algorithm/BaseClusteringAlgorithm.java#L70

I would be happy to help provide a PR to solve this if the maintainers give me some direction as to how this should be resolved.

the jdk attempts to implement a finalizer in `ThreadPoolExecutor` to shutdown the pool if it is garbage collected, however this does not work correctly if threads are started since they also hold a reference to their parent pool (and thus the finalizer is never invoked even if BaseClusteringAlgorithm is garbage collected).

If you wanted to rely on garbage collection I see two possible options:
1) Selfish plug of using a better thread pool.  Thread pools from my `threadly` project do not suffer from this issue.  They can be collected normally, or you can use `CentralThreadlyPool` as a way to avoid needing to manage the lifecycle at all.
2) You can implement a finalizer in `BaseClusteringAlgorithm` which invokes a shutdown to it's retained pool.

I personally don't like depending on finalizers.  My personal preference (though potentially unpopular) would be to have people supply an `Executor` (`Executor` is way better than `ExecutorService` if you can), and then put the thread pool management on the user.

If you are willing to depend on threadly, if an executor is not provided using a threadly executor can provide the ability to fall back to garbage collection for releasing the created threads.

jentfoo:
I guess after looking at `ExecutorServiceProvider` maybe the original intention was to set `instance` (which currently goes unused) and reuse that instance.

saudet:
@jentfoo Thanks for the input! Using threadly looks like a good technical solution, but I think the Mozilla license is a problem, @agibsonccc ?

agibsonccc:
[...]

Issue #3501, opened 2017-06-08
Update dev documentation on how to run tests 


huitseeker:
#### Issue Description

https://deeplearning4j.org/devguide

Developer documentation should include how to run tests from the command line:

mvn clean test -Ptest-nd4j-native
 or mvn clean test -Ptest-nd4j-native -Dtest=FooBar -DfailIfNoTests=false for one test
 or mvn clean test -Ptest-nd4j-native -pl myProject -amd for a subproject

An important test dependency is https://github.com/deeplearning4j/dl4j-test-resources (we have it separate because of how big it is), and the doc should emphasize how importnat it is to clone & install it.

PR #4794, opened 2018-03-13
Concurrency changes, primary around thread interrupted status and pooling

jentfoo:
This PR is designed to address issues #4766 and #4781.  The commit messages should provide fairly detailed descriptions about what I was changing.

Let me know what you think and what you would like to see changed for this PR.  I think this should be fairly safe, though I can't seem to get the compile script to fully work (it fails on cuda for me).  I do have some ideas for some minor potential performance improvements, but I wanted to make sure I do my hello world PR first.

### DIFF
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
@@ -19,7 +19,6 @@ import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.LockSupport;
 
 /**
  * Async prefetching iterator wrapper for MultiDataSetIterator implementations
@@ -385,7 +384,7 @@ public class AsyncDataSetIterator implements DataSetIterator {
         private BlockingQueue<DataSet> queue;
         private DataSetIterator iterator;
         private DataSet terminator;
-        private AtomicBoolean isShutdown = new AtomicBoolean(false);
+        private boolean isShutdown = false; // locked around `this`
         private WorkspaceConfiguration configuration = WorkspaceConfiguration.builder().minSize(10 * 1024L * 1024L)
                         .overallocationLimit(prefetchSize + 1).policyReset(ResetPolicy.ENDOFBUFFER_REACHED)
                         .policyLearning(LearningPolicy.FIRST_LOOP).policyAllocation(AllocationPolicy.OVERALLOCATE)
@@ -450,13 +449,24 @@ public class AsyncDataSetIterator implements DataSetIterator {
                 //log.info("Trying destroy...");
                 //if (useWorkspace)
                 //Nd4j.getWorkspaceManager().getWorkspaceForCurrentThread(workspaceId).destroyWorkspace();
-                isShutdown.set(true);
+                synchronized (this) {
+                    isShutdown = true;
+                    this.notifyAll();
+                }
             }
         }
 
         public void shutdown() {
-            while (!isShutdown.get())
-                LockSupport.parkNanos(100L);
+            synchronized (this) {
+                while (! isShutdown) {
+                    try {
+                        this.wait();
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        throw new RuntimeException(e);
+                    }
+                }
+            }
 
             if (workspace != null) {
                 log.debug("Manually destroying ADSI workspace");
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncMultiDataSetIterator.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncMultiDataSetIterator.java
@@ -18,8 +18,6 @@ import org.nd4j.linalg.factory.Nd4j;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.locks.LockSupport;
 
 /**
  * Async prefetching iterator wrapper for MultiDataSetIterator implementations
@@ -312,8 +310,7 @@ public class AsyncMultiDataSetIterator implements MultiDataSetIterator {
         private BlockingQueue<MultiDataSet> queue;
         private MultiDataSetIterator iterator;
         private MultiDataSet terminator;
-        private AtomicBoolean isShutdown = new AtomicBoolean(false);
-        private AtomicLong internalCounter = new AtomicLong(0);
+        private boolean isShutdown = false; // locked around `this`
         private WorkspaceConfiguration configuration = WorkspaceConfiguration.builder().minSize(10 * 1024L * 1024L)
                         .overallocationLimit(prefetchSize + 1).policyReset(ResetPolicy.ENDOFBUFFER_REACHED)
                         .policyLearning(LearningPolicy.FIRST_LOOP).policyAllocation(AllocationPolicy.OVERALLOCATE)
@@ -382,13 +379,24 @@ public class AsyncMultiDataSetIterator implements MultiDataSetIterator {
                 //log.info("Trying destroy...");
                 //if (useWorkspaces)
                 //Nd4j.getWorkspaceManager().getWorkspaceForCurrentThread(workspaceId).destroyWorkspace();
-                isShutdown.set(true);
+                synchronized (this) {
+                    isShutdown = true;
+                    this.notifyAll();
+                }
             }
         }
 
         public void shutdown() {
-            while (!isShutdown.get())
-                LockSupport.parkNanos(100L);
+            synchronized (this) {
+                while (! isShutdown) {
+                    try {
+                        this.wait();
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        throw new RuntimeException(e);
+                    }
+                }
+            }
 
             if (workspace != null) {
                 log.debug("Manually destroying AMDSI workspace");
--- a/deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/EncodedGradientsAccumulator.java
+++ b/deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/EncodedGradientsAccumulator.java
@@ -5,6 +5,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.deeplearning4j.exception.DL4JInvalidConfigException;
 import org.deeplearning4j.nn.api.Model;
 import org.deeplearning4j.optimize.api.StepFunction;
+import org.deeplearning4j.util.ThreadUtils;
 import org.nd4j.linalg.api.memory.MemoryWorkspace;
 import org.nd4j.linalg.api.memory.conf.WorkspaceConfiguration;
 import org.nd4j.linalg.api.memory.enums.*;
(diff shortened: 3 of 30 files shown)


saudet:
Looks good to me! Thanks. Have you run the tests though to make sure nothing breaks? Just the CPU ones with -Ptest-nd4j-native is fine. There might be some that don't run on your machine anyway, so don't worry too much about those. :)

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
```diff
@@ -385,7 +384,7 @@ public void remove() {
         private BlockingQueue<DataSet> queue;
         private DataSetIterator iterator;
         private DataSet terminator;
-        private AtomicBoolean isShutdown = new AtomicBoolean(false);
+        private boolean isShutdown = false; // locked around `this`
```
AlexDBlack: This doesn't seem like a great idea to me?
We've gone from a lock free thread safe implementation, to something that's not lock free and not strictly thread safe? i.e., need volatile keyword here, at the very least?

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncMultiDataSetIterator.java
```diff
@@ -312,8 +310,7 @@ protected void externalCall() {
         private BlockingQueue<MultiDataSet> queue;
         private MultiDataSetIterator iterator;
         private MultiDataSet terminator;
-        private AtomicBoolean isShutdown = new AtomicBoolean(false);
-        private AtomicLong internalCounter = new AtomicLong(0);
+        private boolean isShutdown = false; // locked around `this`
```
AlexDBlack: As per previous comment (re: volatile).

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/EncodedGradientsAccumulator.java
```diff
@@ -165,7 +165,7 @@ public void registerConsumers(int numConsumers) {
                 log.info("Master thread locks at RC");
 
             while (registered.get()) {
-                LockSupport.parkNanos(100L);
+                ThreadUtils.uncheckedSleep(1);
```
AlexDBlack: Not sure it matters too much in practice, but we are changing the wait time by a few orders of magnitude here...

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
jentfoo: I am glad you brought this up.  There is a LOT of polling designs in this code base.  This was an example of where I switched a polling design to a `wait`/`notify`.

First this is correct, there is no need for a `volatile` keyword because this is only updated or read from inside a synchronized block of `this`.  If this variable needed to be read without synchronized, then we could add `volatile` to achieve that, but right now it's not necessary.

As for no-locking to locking, you are sorta correct.  The synchronized blocks will be held very short, only held while they are updating the value, and sending the notify call.  Or held while checking the value and waking up from their wait.  In this specific case, we will only have one thread doing the update.  And likely only one checking the update, but we can even assume more and this still should perform better than before.

When this would be problematic is if we have a lot of threads in the `wait`.  When the notify occurs those threads would need to acquire the lock one by one as they wake up, then release it.  So I would not suggest changing it to a `wait` / `notify` if there is going to be a lot of waiters, or high thread contention, I just don't think that is the case here.

What's wrong with polling you might ask?  Several things, but the biggest in my mind is the context switching that must occur.  The thread must wakeup, load it's state into the cpu cache, and then check a small boolean before it must again `park` itself.  The more threads you have doing this, the more thrashing occurs.

In short, no solution is the right solution for all cases.  I would like to provide future PR's with benchmarks and prove performance gains.  But these seemed like some obvious and safe points where I could start those changes.

Let me know if you have any more questions or thoughts about this.

deeplearning4j-nn/src/main/java/org/deeplearning4j/optimize/solvers/accumulation/EncodedGradientsAccumulator.java
jentfoo: I would argue this is better (I guess that's obvious since I provided the PR).

Few OS / schedulers will be able to park a thread for only 100 nanoseconds.  IME on linux you are looking at about 1000 nanoseconds in the least, and 1 millisecond at the most.  On windows 1ms tends to actually be optimistic.

I tend to target 1 millisecond because that will provide consistent behavior.  In addition by waking up more frequently you are furthering the context switching / cache miss problem that I described above.

jentfoo:
> Just the CPU ones with -Ptest-nd4j-native is fine.

Do you have any docs on how I run these?  I have been using the `.sh` files in the github, but they fail once they hit cuda.  I also was unsure if they are just compiling, or running tests too.

saudet:
Since you don't need to build ND4J, try to just use the snapshots: https://deeplearning4j.org/snapshots
Then you'll just need to run `mvn -Ptest-nd4j-native ...` inside deeplearning4j. You'll probably need to clone and install https://github.com/deeplearning4j/dl4j-test-resources too though (issue #3501), but that should be about it.

deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/AsyncDataSetIterator.java
AlexDBlack: Thanks for the explanation, I think I'm happy with that. 👍 
Thinking about this more, given we really only have 2 threads operating here, we shouldn't really be worried about performance and lock contention anyway. And, as you say, context switching is a potential (likely?) performance issue anyway with the current design.

jentfoo:
@saudet I got the needed dependencies without too much trouble.  However the master unit tests are not passing on my computer:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project nearestneighbor-core: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: java.lang.RuntimeException:  100, 1, 0, 1, 99]: For input string: " 99]" -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project nearestneighbor-core: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: java.lang.RuntimeException:  100, 1, 0, 1, 99]
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:128)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:307)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:193)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:106)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:863)
	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:288)
	at org.apache.maven.cli.MavenCli.main(MavenCli.java:199)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced(Launcher.java:289)
	at org.codehaus.plexus.classworlds.launcher.Launcher.launch(Launcher.java:229)
	at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode(Launcher.java:415)
	at org.codehaus.plexus.classworlds.launcher.Launcher.main(Launcher.java:356)
Caused by: org.apache.maven.plugin.PluginExecutionException: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: java.lang.RuntimeException:  100, 1, 0, 1, 99]
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:145)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:207)
	... 20 more
Caused by: java.lang.RuntimeException: java.lang.RuntimeException:  100, 1, 0, 1, 99]
	at org.apache.maven.plugin.surefire.booterclient.output.ThreadedStreamConsumer.close(ThreadedStreamConsumer.java:127)
	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:569)
	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:460)
	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:229)
	at org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:201)
	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeProvider(AbstractSurefireMojo.java:1026)
	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.executeAfterPreconditionsChecked(AbstractSurefireMojo.java:862)
	at org.apache.maven.plugin.surefire.AbstractSurefireMojo.execute(AbstractSurefireMojo.java:755)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
	... 21 more
Caused by: java.lang.RuntimeException:  100, 1, 0, 1, 99]
	at org.apache.maven.plugin.surefire.booterclient.output.ForkClient.createReportEntry(ForkClient.java:286)
	at org.apache.maven.plugin.surefire.booterclient.output.ForkClient.processLine(ForkClient.java:152)
	at org.apache.maven.plugin.surefire.booterclient.output.ForkClient.consumeLine(ForkClient.java:115)
	at org.apache.maven.plugin.surefire.booterclient.output.ThreadedStreamConsumer$Pumper.run(ThreadedStreamConsumer.java:70)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NumberFormatException: For input string: " 99]"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:569)
	at java.lang.Integer.valueOf(Integer.java:740)
	at java.lang.Integer.decode(Integer.java:1197)
	at org.apache.maven.plugin.surefire.booterclient.output.ForkClient.createReportEntry(ForkClient.java:278)
	... 4 more
```

Any thoughts as to what I might be missing?

I am guessing since it passed on the CI this is just an issue with my setup still, but I want to make sure I get my local tests working before I attempt more changes.

saudet:
Some tests are flaky, so if the same exact tests fail with and without your patch, we'll be reasonably sure it's not breaking anything :) We can ignore failing test and continue testing with `mvn -Ptest-nd4j-native -Dmaven.test.failure.ignore=true ...`. The CI server isn't in a good enough shape to run the tests for us yet, unfortunately. :(

jentfoo:
@saudet I tried that, but it still is failing.  If you look at the stack I don't even see deeplearnin4j in it.  So I am not sure what the issue is.

AlexDBlack:
Hm... maybe I'll run tests manually on this branch/PR (vs. master) and make a determination. There's a small number of known issue with some tests on master.

AlexDBlack:
Apologies for not getting back to this sooner. I've run some of the tests manually and compared vs. master... not everything is passing on master, but I'm not seeing any new failures on this branch/PR - so I'll merge it.
Thanks again @jentfoo !

jentfoo:
@AlexDBlack No worries at all!  Thank you guys, I am sure I will have some more PR's in the future :)

PR #4855, opened 2018-03-27
Dl4j iterators

agibsonccc:
Split out dl4j iterators from deeplearning4j-core (but still keep same transitive deps)

## How was this patch tested?

## Quick checklist


agibsonccc:
@AlexDBlack when you review this just look at the last few commits, I deleted all the changes related to the PR and split them out. The datavec iterators and specialized iterators are the main focus of this PR.

agibsonccc:
@AlexDBlack go ahead and do another round

deeplearning4j-iterators/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+        <artifactId>deeplearning4j-parent</artifactId>
+        <groupId>org.deeplearning4j</groupId>
+        <version>0.9.2-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>deeplearning4j-data</artifactId>
+    <packaging>pom</packaging>
+
+    <name>deeplearning4j-iterators</name>
```
AlexDBlack: name should match artifactId (i.e., both should be deeplearning4j-data)

deeplearning4j-nn/pom.xml
```diff
@@ -12,11 +12,13 @@
...
 
-    <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.deeplearning4j</groupId>
+            <artifactId>deeplearning4j-utility-iterators</artifactId>
+            <version>${project.version}</version>
```
AlexDBlack: Do we want deeplearning4j-nn relying on deeplearning4j-utility-iterators?
My first thought here would be no - i.e., ```-nn``` should be very minimal.
That's why I was suggesting we keep the async iterators in -nn but move everything else - I *think* the async ones are the only ones that are used directly in -nn. If we move them to -nn, we can drop this dependency here?
Not sure how strict we want to be with the "-nn should be minimal" design...

deeplearning4j-core/pom.xml
```diff
@@ -65,6 +65,22 @@
...
 
     <dependencies>
+        <dependency>
+            <groupId>org.deeplearning4j</groupId>
+            <artifactId>deeplearning4j-tsne</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.deeplearning4j</groupId>
+            <artifactId>deeplearning4j-specialized-iterators</artifactId>
```
AlexDBlack: Update to ```deeplearning4j-datasets```

deeplearning4j-core/pom.xml
```diff
@@ -65,6 +65,22 @@
...
+            <version>${project.version}</version>
+            <scope>test</scope>
```
AlexDBlack: I'd be inclined to make this default (i.e., compile) scope not test scope to make things easier for users - my thinking is that deeplearning4j-core is "all the usual stuff you might need" - things like MNIST iterator included.
It also saves users wondering what's going on when they upgrade after release - otherwise where did their MNIST iterator go, and now how do they find out about the change?

deeplearning4j-nn/pom.xml
agibsonccc: The reason we do this is because the neural net api itself requires the async iterators and the like.

PR #4879, opened 2018-03-31
Fixed missing log class declaration

esantana:
Signed-off-by: Eduardo Sant'Ana da Silva <eduardo.santanadasilva@gmail.com>

Declaring log variable missing instance that was causing compilation errors.

## How was this patch tested?

It was just a compilation error, the log declaration follows other dl4j log declaration.


saudet:
Those are generated by Lombok at compile time, but IntelliJ IDEA doesn't support it without the plugin, so please install it.

Issue #4761, opened 2018-03-06
Windows - CUDA gradient check failures


AlexDBlack:
Test platform: Windows, CUDA 9.0
Edit: Windows CUDA 9.1 fails with exact same symptoms.
Fails when run through intellij or maven

These tests fail on CUDA when all CNNGradientCheckTests are run, but pass:
(a) On CPU
(b) When run in isolation (i.e., one test at a time)
(c) on Linux, CUDA 9.1 (Kubuntu 17.10, GTX 970s)

![image](https://user-images.githubusercontent.com/2360237/37014124-9b42dfd0-2152-11e8-814a-a5c05e1e6057.png)

Note that it's inconsistent - different tests fail during different runs...
![image](https://user-images.githubusercontent.com/2360237/37014593-d5f1dc82-2155-11e8-862e-7883d6b28e67.png)

![image](https://user-images.githubusercontent.com/2360237/37014618-03d67d92-2156-11e8-9ea7-17b502bbefa4.png)

Results when run through maven: https://gist.github.com/AlexDBlack/ca782023187fa0786054b733c6e10960

Not sure what's going on here or why...

Edit: so far I've ruled out:
1. Workspaces (setting WorkspaceMode.NONE makes no difference)
2. RNG seeds (setting before each test - no difference)
3. Destroying workspaces before each gradient check is run

There seems to be some degree of randomness - and possibly multiple issues...
1. Some tests clearly fail due to subsampling layer backprop
2. Other tests fail with a single, very wrong gradient (see below)
3. Other tests just fail outright - all gradients fail, but other tests (very similar params/config) all pass
4. Some tests I can run for a thousand times, without any issues :/

```
o.d.g.GradientCheckUtil - Param 1812 (2_W) passed: grad= 0.04007660303681882, numericalGrad= 0.04007660303617655, relError= 8.01304140042311E-12
o.d.g.GradientCheckUtil - Param 1813 (2_W) FAILED: grad= 0.0034388891218970277, numericalGrad= 48836.31517098053, relError= 0.9999998591667342, scorePlus=0.693761944443137, scoreMinus= 0.5960893141011759
o.d.g.GradientCheckUtil - Param 1814 (2_W) passed: grad= -0.021406821076271437, numericalGrad= -0.021406821093528094, relError= 4.030644680906441E-10
```

AlexDBlack:
Update here: this is still occuring on Windows, still passing consistently on Linux.
[...]

PR #4883, opened 2018-04-03
Force CuDNN for Windows users for now

AlexDBlack:
***DO NOT MERGE***

Forces Windows users to use CuDNN for Convolutional/Subsampling layers.

To be merged only if the following issue is not resolved before release: https://github.com/deeplearning4j/deeplearning4j/issues/4761


AlexDBlack:
No longer required - bug was (finally!) fixed.

PR #4921, opened 2018-04-11
Cropping1d

maxpumperla:


deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNN1DGradientCheckTest.java
```diff
@@ -32,6 +33,83 @@
...
+                        if (PRINT_RESULTS) {
+                            System.out.println(msg);
+                            for (int j = 0; j < net.getnLayers(); j++)
+                                System.out.println("Layer " + j + " # params: " + net.getLayer(j).numParams());
+                        }
+
+                        boolean gradOK = GradientCheckUtil.checkGradients(net, DEFAULT_EPS, DEFAULT_MAX_REL_ERROR,
+                                DEFAULT_MIN_ABS_ERROR, PRINT_RESULTS, RETURN_ON_FIRST_FAILURE, input, labels);
+
+                        assertTrue(msg, gradOK);
```
AlexDBlack: These days: I'm tending to add a ```TestUtils.testModelSerialization(net);```  at the end of gradient checks as an extra check (if only to ensure a bunch of different configurations are tested *somewhere* for every layer)

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/convolutional/Cropping1D.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public InputPreProcessor getPreProcessorForInputType(InputType inputType) {
+        Preconditions.checkArgument(inputType != null, "Invalid input for Cropping1D layer (layer name=\""
+                + getLayerName() + "\"): InputType is null");
+        return InputTypeUtil.getPreProcessorForInputTypeCnnLayers(inputType, getLayerName());
+    }
+
+    @Override
+    public LayerMemoryReport getMemoryReport(InputType inputType) {
+        return null;
```
AlexDBlack: Return empty memory report perhaps.

PR #4927, opened 2018-04-13
[WIP] conv 3d layer

maxpumperla:


deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNN3DGradientCheckTest.java
```diff
@@ -0,0 +1,113 @@
...
+    private static final double DEFAULT_EPS = 1e-6;
+    private static final double DEFAULT_MAX_REL_ERROR = 1e-3;
+    private static final double DEFAULT_MIN_ABS_ERROR = 1e-8;
+
+    static {
+        Nd4j.setDataType(DataBuffer.Type.DOUBLE);
+    }
+
+    @Test
+    public void testCnn3DPlain() {
```
AlexDBlack: IMO this test needs to be more robust:
1. Test at least some different input d/h/w - with different values for both. Why? In case there's some issue with native code (using wrong arg/value, which won't be picked up with all same values)
2. Ditto for testing multile kernels/strides - I'd do something like 1/1/1, 2/2/2, 1/2/3

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/ConvolutionMode.java
```diff
@@ -60,6 +60,6 @@
  */
 public enum ConvolutionMode {
 
-    Strict, Truncate, Same
+    Strict, Truncate, Same, Valid
```
AlexDBlack: Why add valid? and how does it differ from truncate (unless it's an alias)?
If it is necessary (or just an alias) - update the javadoc.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/inputs/InputType.java
```diff
@@ -176,7 +205,15 @@ public int arrayElementsPerExample() {
     public static class InputTypeConvolutional extends InputType {
         private int height;
         private int width;
-        private int depth;
+        private int channels;
+
+        public int getDepth() {
```
AlexDBlack: Add deprecated tag + javadoc

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/inputs/InputType.java
```diff
@@ -176,7 +205,15 @@ public int arrayElementsPerExample() {
...
+            return channels;
+        }
+
+        public void setDepth(int depth) {
```
AlexDBlack: Add deprecated tag + javadoc

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/convolutional/KerasConvolution3D.java
```diff
@@ -0,0 +1,155 @@
...
+import static org.deeplearning4j.nn.modelimport.keras.utils.KerasActivationUtils.getActivationFromConfig;
+import static org.deeplearning4j.nn.modelimport.keras.utils.KerasInitilizationUtils.getWeightInitFromConfig;
+import static org.deeplearning4j.nn.modelimport.keras.utils.KerasLayerUtils.getHasBiasFromConfig;
+import static org.deeplearning4j.nn.modelimport.keras.utils.KerasLayerUtils.getNOutFromConfig;
+
+
+/**
+ * Imports a 3D Convolution layer from Keras.
+ *
+ * @author dave@skymind.io
```
AlexDBlack: Author :)

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution3D.java
```diff
@@ -0,0 +1,193 @@
...
+ * @author Max Pumperla
+ */
+@Data
+@NoArgsConstructor
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+public class Convolution3D extends ConvolutionLayer {
+
+    private ConvolutionMode mode = ConvolutionMode.Same;  // in libnd4j: 0 - same mode, 1 - valid mode
+    private boolean isNCDHW = true; // in libnd4j: 1 - NCDHW, 0 - NDHWC
```
AlexDBlack: Would it be clearer to have an enumeration here instead of a boolean? Also a bit more flexible in case we ever have to change (or add another) format.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution3D.java
```diff
@@ -0,0 +1,193 @@
...
+
+        public Builder(int[] kernelSize, int[] stride) {
+            super(kernelSize, stride);
+        }
+
+        public Builder(int... kernelSize) {
+            super(kernelSize);
+        }
+
+        public Builder() {
```
AlexDBlack: Different defaults here to the field defaults...

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution3D.java
```diff
@@ -0,0 +1,193 @@
...
+            super(new int[]{1, 1, 1},
+                    new int[]{1, 1, 1},
+                    new int[]{0, 0, 0},
+                    new int[]{0, 0, 0});
+        }
+
+
+        public Builder kernelSize(int... kernelSize) {
```
AlexDBlack: Maybe: add a preconditions check on (a) not null, and (b) length 3 - applies for all methods here.
FYI we have this now: https://github.com/deeplearning4j/nd4j/pull/2845

Also: javadoc on these methods. And make sure you specify the order (h,w,d?) for the dimensions in said javadoc.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/ConvolutionLayer.java
```diff
@@ -89,14 +89,14 @@ protected ConvolutionLayer(BaseConvBuilder<?> builder) {
         this.hasBias = builder.hasBias;
         this.convolutionMode = builder.convolutionMode;
         this.dilation = builder.dilation;
-        if (builder.kernelSize.length != 2)
-            throw new IllegalArgumentException("Kernel size of should be rows x columns (a 2d array)");
+//        if (builder.kernelSize.length != 2)
+//            throw new IllegalArgumentException("Kernel size of should be rows x columns (a 2d array)");
```
AlexDBlack: Keep the checks - but maybe add a numDimensions() on the builder or something... so builder.kernelSize.length != numDimensions() or whatever.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/preprocessor/Cnn3DToFeedForwardPreProcessor.java
```diff
@@ -0,0 +1,163 @@
...
+                    + "for numChannels: " + numChannels + ", inputDepth " + inputDepth + ", inputHeight " + inputHeight
+                    + " and inputWidth " + inputWidth + ", but got "
+                    + Arrays.toString(input.shape()));
+        }
+
+
+        //Assume input is standard rank 5 activations out of CNN3D layer
+        //First: we require input to be in c order. But c order (as declared in array order) isn't enough;
+        // also need strides to be correct
+        if (input.ordering() != 'c' || !Shape.strideDescendingCAscendingF(input))
```
AlexDBlack: These days: I'm using this method instead of strideDescending... as it's a bit more strict (and catches more invalid cases)
https://github.com/deeplearning4j/nd4j/blob/5d4e8a0a873eed92980910552767ed2626d1270b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/shape/Shape.java#L2609-L2622

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/preprocessor/Cnn3DToFeedForwardPreProcessor.java
```diff
@@ -0,0 +1,163 @@
...
+        int[] inShape = input.shape();
+        int[] outShape = new int[]{inShape[0], inShape[1] * inShape[2] * inShape[3] * inShape[4]};
+
+        return input.reshape('c', outShape);
+    }
+
+    @Override
+    public INDArray backprop(INDArray epsilons, int miniBatchSize) {
+        //Epsilons are 2d, with shape [miniBatchSize, outChannels*outD*outH*outW]
+        if (epsilons.ordering() != 'c' || !Shape.strideDescendingCAscendingF(epsilons))
```
AlexDBlack: As above.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/preprocessor/FeedForwardToCnn3DPreProcessor.java
```diff
@@ -0,0 +1,176 @@
...
+    public FeedForwardToCnn3DPreProcessor(int inputDepth, int inputWidth, int inputHeight) {
+        this.inputDepth = inputDepth;
+        this.inputHeight = inputHeight;
+        this.inputWidth = inputWidth;
+        this.numChannels = 1;
+    }
+
+    @Override
+    public INDArray preProcess(INDArray input, int miniBatchSize) {
+        if (input.ordering() != 'c' || !Shape.strideDescendingCAscendingF(input))
```
AlexDBlack: As per earlier comment - I'm preferring Shape.hasDefaultStridesForShape these days

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/Convolution3DLayer.java
```diff
@@ -0,0 +1,257 @@
...
+        return new Pair<>(retGradient, reshapedEpsilon);
+    }
+
+
+    @Override
+    public INDArray preOutput(boolean training) {
+        return preOutput(training, false).getFirst();
+    }
+
+    protected Pair<INDArray, INDArray> preOutput(boolean training, boolean forBackprop) {
```
AlexDBlack: There should be an applyDropoutIfNecessary call here somewhere?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/Convolution3DLayer.java
```diff
@@ -0,0 +1,257 @@
...
+                kernel[0], kernel[1], kernel[2],
+                strides[0], strides[1], strides[2],
+                pad[0], pad[1], pad[2],
+                dilation[0], dilation[1], dilation[2],
+                mode == ConvolutionMode.Same ? 1 : 0,
+                isNCDHW ? 0 : 1
+        };
+
+        CustomOp op;
+        if (layerConfig.hasBias()) {
```
AlexDBlack: Minor thing with this if statement: just have an INDArray[] inputArgs = ... in the if statement. Everything else in the builder there is exactly the same for both cases, and can be pulled out. That's a bit cleaner IMO.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/Convolution3DLayer.java
```diff
@@ -0,0 +1,257 @@
...
+        int outD = outSize[0];
+        int outH = outSize[1];
+        int outW = outSize[2];
+
+        INDArray output = Nd4j.create(miniBatch * outWeightChannels * outH * outW * outD);
+        INDArray reshapedOutput;
+        if (isNCDHW)
+            reshapedOutput = output.reshape('c', miniBatch, outWeightChannels, outD, outH, outW);
+        else
+            reshapedOutput = output.reshape('c', miniBatch, outD, outH, outW, outWeightChannels);
```
AlexDBlack: You don't need to do this "create and reshape" routine. Just do an Nd4j.create with the appropriate shape in the first place.

deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/CNN3DGradientCheckTest.java
maxpumperla: @AlexDBlack of course, I'll do that once this first sanity check works.

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/convolutional/KerasConvolution3D.java
maxpumperla: I try to attribute the ugly stuff to @turambar

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/Convolution3DLayer.java
maxpumperla: that's taken care of in the `activate` method in the super class.

PR #4954, opened 2018-04-19
Keras: optional input shape (for no-top models)

maxpumperla:
Models, especially Keras applications like ResNet etc., can come without input shape, e.g. when `include_top=False` in applications. For this case we can now optionally provide input shape info in all relevant places.


maxpumperla:
comes with Xception model as full-blown test for the mechanism.

Issue #4951, opened 2018-04-19
Use `TrainingListener` in `MultiLayerNetwork.addListeners`


holgerbrandl:
Currently (1.0.0-alpha), the `MultiLayerNetwork.addListeners` expects `IterationListeners`s. However, by doing so,  it's not clear that/if `addListeners` is the correct API to register the more general `TrainingListener`s. The internal upcast is not obvious.

It would be better design if the signature would be `MultiLayerNetwork.addListeners(TrainingListener… listeners)`.

PR #4975, opened 2018-04-25
Use TrainingListener instead of IterationListener in Model.addListeners

holgerbrandl:
As discussed on gitter, it would be a better design to register training callbacks with `org.deeplearning4j.nn.api.Model.addListeners(TrainingListener… listeners)`. 
Currently the API suggests that only `IterationListener`s can be registered with `addListeners`. The internal upcast is not obvious.

This PR implements that change and fixes all usage sites (like classes extending `Model`).


holgerbrandl:
I've tried to enable the CI but it fails with 
image

holgerbrandl:
This PR fixes https://github.com/deeplearning4j/deeplearning4j/issues/4951

AlexDBlack:
OK, so this is a *much* more extensive change than I had in mind.
specifically (based on our gitter discussion) I thought the intent here was just to add some addListener(TrainingListener...) overloads. That would be a pretty minor change but would improve usability a bit.

This PR basically makes IterationListener obsolete and unusable. Whereas previously both IterationListener and TrainingListener instances could be set via setListeners(IterationListener...) etc methods, now IterationListener can't be set or used anywhere due to the signature change (i.e., you can't an IterationListener to setListeners(TrainingListener...) or similar methods).

I guess it comes down to the following question: do we simply remove IterationListener entirely, and have only one listener type (TrainingListener)?
That's certainly cleaner and more maintainable long-term - but I'm not sure how I feel about the breaking API change.

@agibsonccc @maxpumperla @raver119 any thoughts?

sshepel:
> I've tried to enable the CI but it fails with

@holgerbrandl CI enable by default, and will be run for any changes in your PR.
The screenshot that you have posted says the you have no permissions to view the log of the build.
Will try to fix it ASAP (maybe your help will be needed).

Thanks for your observation.

holgerbrandl:
Indeed @AlexDBlack `IterationListener` would be obsolete. And yes, a second `addListeners` would have been a very pragmatic solution, but would have made the API more confusing because `IterationListener`s could be added by both `addListeners` which felt not correct to me.

The mental picture which I had in mind when suggesting the change was the callback API in keras https://keras.io/callbacks/. In its current state  users might wonder why dl4j has no such general API to intercept training here and there (except for the rather limited `IterationListener`), because it is not discoverable from an API perspective. The internal upcast (all across the internal implementations which was removed in the PR) is hiding the fact, that dl4h has a flexible callback API as well.

holgerbrandl:
@sshepel I have now received an email saying
> Changes: 4975, Use TrainingListener instead of IterationListener in Model.addListeners
> Author: holgerbrandl (null)
but the included link to the jenkins is still saying "permission denied" as before.

sshepel:
@holgerbrandl thanks for heads up! Will check one more thing.

AlexDBlack:
Thanks for the PR: rebased, updated and merged here: https://github.com/deeplearning4j/deeplearning4j/pull/5014

PR #4985, opened 2018-04-26
Zoo API updates and fixes

crockpotveggies:

Model selector is no longer useful. This also contains API changes which will facilitate future benchmarking and easier instantiation via a builder method. YOLO2 is also updated.

## How was this patch tested?

Tests passing.

## Quick checklist

- [X] Reviewed the [Contributing Guidelines](https://github.com/deeplearning4j/deeplearning4j/blob/master/CONTRIBUTING.md) and followed the steps within.
- [X] Ran mvn formatter:format (see [formatter instructions](http://code.revelc.net/formatter-maven-plugin/examples.html#Setting_Source_Files) for targeting your specific files).


deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/YOLO2.java
```diff
@@ -70,51 +70,37 @@
...
+    @Builder.Default private IUpdater updater = new Adam(1e-3);
+    @Builder.Default private CacheMode cacheMode = CacheMode.DEVICE;
+    @Builder.Default @Deprecated private WorkspaceMode workspaceMode = WorkspaceMode.ENABLED;
+    @Builder.Default private ConvolutionLayer.AlgoMode cudnnAlgoMode = ConvolutionLayer.AlgoMode.PREFER_FASTEST;
 
     @Override
     public String pretrainedUrl(PretrainedType pretrainedType) {
         if (pretrainedType == PretrainedType.IMAGENET)
-            return "http://blob.deeplearning4j.org/models/yolo2_dl4j_inference.v1.zip";
+            return "http://blob.deeplearning4j.org/models/yolo2_dl4j_inference.v2.zip";
```
saudet: What changed?

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/ZooModel.java
```diff
@@ -1,5 +1,6 @@
 package org.deeplearning4j.zoo;
 
+import lombok.Builder;
```
AlexDBlack: Unused import?

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/AlexNet.java
```diff
@@ -38,26 +40,17 @@
...
-        this.workspaceMode = workspaceMode;
-        this.cudnnAlgoMode = workspaceMode == WorkspaceMode.ENABLED ? ConvolutionLayer.AlgoMode.PREFER_FASTEST
-                        : ConvolutionLayer.AlgoMode.NO_WORKSPACE;
-    }
+    @Builder.Default private long seed = 1234;
+    @Builder.Default private int[] inputShape = new int[] {3, 224, 224};
+    private int numLabels;
+    @Builder.Default private IUpdater updater = new Nesterovs(1e-2, 0.9);
+    @Builder.Default private CacheMode cacheMode = CacheMode.DEVICE;
+    @Builder.Default @Deprecated private WorkspaceMode workspaceMode = WorkspaceMode.ENABLED;
```
AlexDBlack: We may want to keep this non-deprecated, for benchmarks and the like?

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/AlexNet.java
```diff
@@ -84,16 +72,20 @@ public MultiLayerConfiguration conf() {
...
+                        .dist(new NormalDistribution(0.0, 0.01))
+                        .activation(Activation.RELU)
+                        .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
+                        .updater(updater)
                         .biasUpdater(new Nesterovs(2e-2, 0.9))
                         .convolutionMode(ConvolutionMode.Same)
                         .gradientNormalization(GradientNormalization.RenormalizeL2PerLayer) // normalize to prevent vanishing or exploding gradients
-                        .trainingWorkspaceMode(WorkspaceMode.ENABLED).inferenceWorkspaceMode(WorkspaceMode.ENABLED)
+                        .trainingWorkspaceMode(WorkspaceMode.ENABLED)
+                        .inferenceWorkspaceMode(WorkspaceMode.ENABLED)
```
AlexDBlack: Config option should be used here, not hard-coded

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/GoogLeNet.java
```diff
@@ -135,10 +126,17 @@ private GraphBuilder inception(GraphBuilder graph, String name, int inputSize, i
...
-                        .activation(Activation.RELU).optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
-                        .updater(new Nesterovs(new StepSchedule(ScheduleType.ITERATION, 1e-2, 0.96, 320000), 0.9))
+                        .activation(Activation.RELU)
+                        .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
+                        .updater(updater)
                         .biasUpdater(new Nesterovs(new StepSchedule(ScheduleType.ITERATION, 2e-2, 0.96, 320000), 0.9))
-                        .weightInit(WeightInit.XAVIER).l2(2e-4).graphBuilder();
+                        .weightInit(WeightInit.XAVIER)
+                        .l2(2e-4)
+                        .cacheMode(CacheMode.DEVICE)
```
AlexDBlack: Hard-coded cache mode

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/FaceNetNN4Small2.java
```diff
@@ -26,27 +29,18 @@
...
-    public FaceNetNN4Small2(int numLabels, long seed, WorkspaceMode workspaceMode) {
-        this.numLabels = numLabels;
-        this.seed = seed;
-        this.workspaceMode = workspaceMode;
-        this.cudnnAlgoMode = workspaceMode == WorkspaceMode.ENABLED ? ConvolutionLayer.AlgoMode.PREFER_FASTEST
-                        : ConvolutionLayer.AlgoMode.NO_WORKSPACE;
-    }
+    @Builder.Default private IUpdater updater = new Adam(0.1, 0.9, 0.999, 0.01);
+    @Builder.Default private Activation transferFunction = Activation.RELU;
+    @Builder.Default CacheMode cacheMode = CacheMode.DEVICE;
```
AlexDBlack: This applies to all nets: I'm not comfortable with this being the default, when it's *not* the default for DL4J in general. I haven't run the benchmarks here, but I do know cache mode can use a *lot* of memory... for each conv layer, the cached im2col array is ```activationsNumElements * kH * kW / (sH * sW)```. For something like the early layers of AlexNet (11x11 / 4x4 =7.56 if memory serves me) that's a lot of memory.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/YOLO2.java
AlexDBlack: @saudet see dev channel. But basically, JSON format change for SpaceToBatch

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/FaceNetNN4Small2.java
crockpotveggies: @raver119 can you comment here? personally, I've seen negligible memory usage when cache mode is enabled on VGG16 (with the exception of that strange OOM on batch size > 24).

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/GoogLeNet.java
crockpotveggies: Good catch.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/AlexNet.java
crockpotveggies: Agreed.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/FaceNetNN4Small2.java
crockpotveggies: Because cache mode in my own experience has consistently led to good performance, I'm not in favour of changing the default. Taking a quick look at nvidia-smi on VGG-16, here's the memory usage of cache mode enabled vs disabled. That's not enough for my concern, but please tell me otherwise:

| Model | Cache Mode | Batch | Memory Usage |
|---|---|---|---|
| VGG16 | Enabled | 16 | 3733MiB / 12194MiB |
| VGG16 | Disabled | 16 | 3065MiB / 12194MiB |
| AlexNet | Enabled | 48 | 2317MiB / 12194MiB |
| AlexNet | Disabled | 48 | 2193MiB / 12194MiB |

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/AlexNet.java
crockpotveggies: Good catch

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/FaceNetNN4Small2.java
saudet: This doesn't apply just to the cache setting though. There's a lot of configuration that we cannot change with this interface, so I've always considered what actually gets set as default to be a moot point. We often end up just calling conf() or something and taking it up from there.

AlexDBlack: @crockpotveggies Was CuDNN used for these tests?

crockpotveggies: Yes cuDNN was used in these tests.

AlexDBlack: CacheMode doesn't apply when using CuDNN - it can't, due to how cudnn works

Note the CuDNN helper call (and return) happens before the caching:
https://github.com/deeplearning4j/deeplearning4j/blob/491507004a571ba0793fac90548a1d8515fd8e63/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionLayer.java#L306-L336

https://github.com/deeplearning4j/deeplearning4j/blob/491507004a571ba0793fac90548a1d8515fd8e63/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/convolution/ConvolutionLayer.java#L372-L376

crockpotveggies: Doesn't make sense here. All of those values I showed you used cuDNN. So what is making the memory difference?

AlexDBlack: Not sure, there shouldn't be any difference. Open an issue if you're confident there is a difference here?
(Though I'd recommend checking/confirming it with the memory benchmarks first)

crockpotveggies:
I'm going to change the cache default to NONE for now so we can get this merged. Re: default constructor, we're going to need people to define the numClasses() otherwise I think we'll get support questions about it ("why is my output shape 0?")

crockpotveggies:
Ah I quickly realized this is Java 7, and this pattern would require the impossible consequence of overriding methods from the ZooModel superclass. I'll add the NoArgs back :)

saudet:
Probably related to the other flaky CNN failures we've been seeing. Let's merge master and see?

crockpotveggies:
Hmm looks like it's still happening

```
14:45:50.219 [main] INFO  o.d.zoo.TestInstantiation - Testing training on zoo model org.deeplearning4j.zoo.model.TinyYOLO

java.lang.IllegalArgumentException: Too many indices for array. Number of indexes must be <= rank()
```

crockpotveggies:
I checked file history, including DarkNetHelper, and I don't see anything obvious here. This includes a recent merge from master.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/InstantiableModel.java
```diff
@@ -12,9 +12,7 @@
 
     public Model init();
 
-    public ModelMetaData metaData();
-
-    public ZooType zooType();
+    @Deprecated public ModelMetaData metaData();
```
AlexDBlack: As a general rule: any time we add a ```@Deprecated``` tag we should have javadoc to explain why and/or what to use instead. Even better: add a date/version (pretty sure we're going with 1.0.0-beta for next release)

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/ModelMetaData.java
```diff
@@ -9,6 +9,7 @@
  */
 @Getter
 @AllArgsConstructor
+@Deprecated
```
AlexDBlack: As per previous deprecated comment

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/AlexNet.java
```diff
@@ -69,66 +61,104 @@ public long pretrainedChecksum(PretrainedType pretrainedType) {
...
     public MultiLayerConfiguration conf() {
         double nonZeroBias = 1;
-        double dropOut = 0.5;
 
         MultiLayerConfiguration conf = new NeuralNetConfiguration.Builder().seed(seed)
-                        .weightInit(WeightInit.DISTRIBUTION).dist(new NormalDistribution(0.0, 0.01))
-                        .activation(Activation.RELU).optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
-                        .updater(new Nesterovs(1e-2, 0.9))
+                        .weightInit(WeightInit.DISTRIBUTION)
+                        .dist(new NormalDistribution(0.0, 0.01))
```
AlexDBlack: Not a big deal - but .weightInit(Distribution) is a thing - equivalent to the 2 .weightInit(DISTRIBUTION).dist(y) calls.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/InceptionResNetV1.java
```diff
@@ -103,8 +90,14 @@ public ComputationGraph init() {
         ComputationGraphConfiguration.GraphBuilder graph = new NeuralNetConfiguration.Builder().seed(seed)
                         .activation(Activation.RELU)
                         .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
-                        .updater(new RmsProp(0.1, 0.96, 0.001)).weightInit(WeightInit.DISTRIBUTION)
-                        .dist(new NormalDistribution(0.0, 0.5)).l2(5e-5).miniBatch(true)
+                        .updater(updater)
+                        .weightInit(WeightInit.DISTRIBUTION)
+                        .dist(new NormalDistribution(0.0, 0.5))
```
AlexDBlack: As per previous weightInit comment.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/ResNet50.java
```diff
@@ -173,9 +161,18 @@ private void convBlock(ComputationGraphConfiguration.GraphBuilder graph, int[] k
         ComputationGraphConfiguration.GraphBuilder graph = new NeuralNetConfiguration.Builder().seed(seed)
                         .activation(Activation.IDENTITY)
                         .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
-                        .updater(new RmsProp(0.1, 0.96, 0.001)).weightInit(WeightInit.DISTRIBUTION)
-                        .dist(new NormalDistribution(0.0, 0.5)).l1(1e-7).l2(5e-5).miniBatch(true)
-                        .convolutionMode(ConvolutionMode.Truncate).graphBuilder();
+                        .updater(updater)
+                        .weightInit(weightInit)
+                        .dist(new NormalDistribution(0.0, 0.5))
```
AlexDBlack: As per previous weightInit comment

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/SqueezeNet.java
```diff
@@ -0,0 +1,170 @@
...
+        return model;
+    }
+
+    public ComputationGraphConfiguration.GraphBuilder graphBuilder() {
+
+        ComputationGraphConfiguration.GraphBuilder graph = new NeuralNetConfiguration.Builder().seed(seed)
+                .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
+                .updater(updater)
+                .weightInit(weightInit)
+                .dist(new NormalDistribution(0.0, 0.5))
```
AlexDBlack: As per previous weightInit comment

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/UNet.java
```diff
@@ -0,0 +1,219 @@
...
+        return model;
+    }
+
+    public ComputationGraphConfiguration.GraphBuilder graphBuilder() {
+
+        ComputationGraphConfiguration.GraphBuilder graph = new NeuralNetConfiguration.Builder().seed(seed)
+                .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
+                .updater(updater)
+                .weightInit(weightInit)
+                .dist(new NormalDistribution(0.0, 0.5))
```
AlexDBlack: As per previous weightInit comment

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/Xception.java
```diff
@@ -0,0 +1,238 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class Xception extends ZooModel {
+
+    @Builder.Default private long seed = 1234;
+    @Builder.Default private int[] inputShape = new int[] {3, 299, 299};
+    private int numClasses;
+    @Builder.Default private WeightInit weightInit = WeightInit.RELU;
+    @Builder.Default private IUpdater updater = new AdaGrad(1e-4);
```
AlexDBlack: Is this correct? I'm not familiar with the architecture, but AdaGrad is a curious choice for the default...
(Looks like they use momentum and RMSProp in the paper with a LR decay)

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/InstantiableModel.java
crockpotveggies: See update.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/ModelMetaData.java
crockpotveggies: See updated PR.

PR #5025, opened 2018-05-02
Keras import squeezenet as e2e test

maxpumperla:
For Mr. @crockpotveggies :)

### DIFF
--- a/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasModelUtils.java
+++ b/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasModelUtils.java
@@ -238,13 +238,16 @@ public class KerasModelUtils {
                         // next line will throw an HDF5 group error. This is inessential, as the model still
                         // runs, but might lead to confusion to end users. So we try to catch this here.
                         // TODO: find a better way to do this
+                        String emptyWeightsWarning = "No HDF5 group with weights found for layer with name "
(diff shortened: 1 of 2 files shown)


deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/utils/KerasModelUtils.java
```diff
@@ -238,13 +238,16 @@ public static void importWeights(Hdf5Archive weightsArchive, String weightsRoot,
...
                                 || layerName.contains("rnn") || layerName.contains("gru")
                                 || layerName.contains("embedding") || layerName.contains("batch")
                                 || layerName.contains("locally") || layerName.contains("bidirectional")) {
-                            if (!layerName.contains("pad") && !layerName.contains("block")) {
+                            try {
                                 layerParamNames = weightsArchive.getDataSets(rootPrefix + baseAttributes);
-                            } else {
+                            } catch (Exception e){
+                                log.warn(emptyWeightsWarning);
                                 layerParamNames = Collections.emptyList();
```
crockpotveggies: Does this leave any room for error?

maxpumperla: @crockpotveggies sorry, didn't see this in time. The only thing that happens is that you get these pesky HDF5 group warnings thrown in your face. I can' suppress them (thrown through javacpp). Otherwise this works fine.

Issue #4760, opened 2018-03-05
Add support for frozen output layer


AlexDBlack:
Moved from: https://github.com/deeplearning4j/nd4j/issues/2706 - cc @ionspin

#### Issue Description
> As discussed on gitter, currently frozen layers don't support output layers. If a frozen layer wraps an output layer an exception is thrown.
```
        if (insideLayer instanceof OutputLayer) {
            throw new IllegalArgumentException("Output Layers are not allowed to be frozen " + layerId());
        }
```

#### Version Information

* Deeplearning4j 0.9.2-SNAPSHOT

#### Contributing

I'd like to help out with this issue, but would definitely need some guidance

AlexDBlack:
OK, so the starting point here would be to simply remove that exception and add some unit tests - I'm not sure if it'll work fine or not TBH. The exception was more of a design decision - at the time we didn't think of a legitimate use case that involved freezing the output layer. In most cases it still doesn't make sense, but there are exceptions.

raver119:
I guess it's safe to close this issue now?

AlexDBlack:
No. It *might* work currently, but there's definitely no unit tests for this.

ionspin:
Sorry for my delay on this, I have been moving between cities, so I didn't have much free time. I'll work on it this weekend.

ionspin:
I have tried to understand the the flow and I think this is the current state:
* Currently when there is a frozen layer in a multi-layer network model, all of the layers that came before it will not be trained, as FrozenLayer will return default gradient and null as epsilon.
* Also there is this code in calcBackpropGradients method of MultiLayerNetwork.java https://github.com/deeplearning4j/deeplearning4j/blob/355747d1688e664e80fc3f412fc7a1268af2d958/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/multilayer/MultiLayerNetwork.java#L1541-L1544
that also points to backprop being stopped at first frozen layer.
* On the other hand when I was playing around with MNIST GAN, it was suggested that using the Sgd updater with learning rate set to 0.0 would suffice, as I needed to freeze last three and the output layer, and train the layers leading up to those.

So with this intro, would it make more sense to leave the current FrozenLayer implementation as is, and introduce a new BackpropEnabledFrozenLayer (or similar) class that would use Sgd to freeze the parameters but allow the backprop to continue?

ionspin:
[...]

PR #5095, opened 2018-05-08
Frozen layer with backprop implementation

ionspin:

Introduce FrozenLayerWithBackprop which would enable having frozen layers (including output layer) and still train layers that precede the frozen layer. This should be useful for GANs.

This was achieved by letting the backpropGradient() execute, but making sure that the gradients are zeroed out.

This should solve #4760 as well

## How was this patch tested?

Unit test were written making sure that the frozen layers stay frozen, while the unfrozen layers can be trained. SGD updater with learning rate 0.0 was used to confirm that the trained layers are having the correct values as when using FrozenLayerWithBackprop. Test were done both for ComputationGraph and MultiLayerNetwork

The implementation was also used in a (poorly tuned) GAN implementation which provided same results as when using SGD updaters.

## Quick checklist

### DIFF
--- a/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/FrozenLayerWithBackpropTest.java
+++ b/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/FrozenLayerWithBackpropTest.java
@@ -1,0 +1,612 @@
+package org.deeplearning4j.nn.layers;
+
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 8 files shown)


AlexDBlack:
@ionspin Thanks!
This was mostly good, but I made some modifications/cleanup, added a gradient check and merged it here: https://github.com/deeplearning4j/deeplearning4j/pull/5106
(FYI you'll still get credited for the commit on github :))

PR #5113, opened 2018-05-09
Add NASNet and change no arg constructors

crockpotveggies:

Adds Neural Network Architecture Search. Removes no arg constructors until this specific fix from lombok is merged and we update our dependency: https://github.com/rzwitserloot/lombok/pull/1675

Also adds a generic method to get the input shape from `InputType`.

@AlexDBlack can you please help me verify this network? NASNet is quite complex and it will need to be checked.


deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/inputs/InputType.java
```diff
@@ -67,6 +67,9 @@
     @JsonIgnore
     public abstract int arrayElementsPerExample();
 
+    @JsonIgnore
+    public abstract int[] getShape();
```
AlexDBlack: Let's
(a) add javadoc here, and
(b) add a boolean arg: ```getShape(boolean includeMinibatchDim)```

When arg is true, return things like [-1,size], [-1, size, tsLength], [-1, channels, height, width].
When false, return basically what you already have.

Why? With boolean arg the shape actually matches the INDArray shape you would use (and more useful - I won't have to manually create a new int[] every time I use it for tests etc) - and making it explicit is much less error prone for users too.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/NASNet.java
```diff
@@ -0,0 +1,193 @@
...
+
+    public ComputationGraphConfiguration.GraphBuilder graphBuilder() {
+
+        int filters = (int) Math.floor(penultimateFilters / 24);
+
+        ComputationGraphConfiguration.GraphBuilder graph = new NeuralNetConfiguration.Builder().seed(seed)
+                .optimizationAlgo(OptimizationAlgorithm.STOCHASTIC_GRADIENT_DESCENT)
+                .updater(updater)
+                .weightInit(weightInit)
+                .dist(new NormalDistribution(0.0, 0.5))
```
AlexDBlack: Not used unless using distribution weight init - and can't be customized even then.
I'd remove entirely.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/NASNet.java
```diff
@@ -0,0 +1,193 @@
...
+        ComputationGraphConfiguration conf = graph.build();
+        ComputationGraph model = new ComputationGraph(conf);
+        model.init();
+
+        return model;
+    }
+
+    public ComputationGraphConfiguration.GraphBuilder graphBuilder() {
+
+        int filters = (int) Math.floor(penultimateFilters / 24);
```
AlexDBlack: I'd add a check/exception like the keras implementation:
```
if penultimate_filters % 24 != 0:
        raise ValueError(
            'For NASNet-A models, the value of `penultimate_filters` '
            'needs to be divisible by 24. Current value: %d' %
            penultimate_filters)
```

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/NASNet.java
```diff
@@ -0,0 +1,193 @@
...
+                .updater(updater)
+                .weightInit(weightInit)
+                .dist(new NormalDistribution(0.0, 0.5))
+                .l2(5e-5)
+                .miniBatch(true)
+                .cacheMode(cacheMode)
+                .trainingWorkspaceMode(workspaceMode)
+                .inferenceWorkspaceMode(workspaceMode)
+                .cudnnAlgoMode(cudnnAlgoMode)
+                .convolutionMode(ConvolutionMode.Truncate)
```
AlexDBlack: Model uses a mix of same and valid (truncate) - I'd maybe make it explicit on every layer?

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/NASNet.java
```diff
@@ -0,0 +1,193 @@
...
+                .graphBuilder();
+
+        if(!skipReduction) {
+            graph.addLayer("stem_conv1", new ConvolutionLayer.Builder(3, 3).stride(2, 2).nOut(penultimateFilters).hasBias(false)
```
AlexDBlack: Should be stemFilters not penultimateFilters

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/NASNet.java
```diff
@@ -0,0 +1,193 @@
...
+                    .cudnnAlgoMode(cudnnAlgoMode).build(), "input");
+        } else {
+            graph.addLayer("stem_conv1", new ConvolutionLayer.Builder(3, 3).stride(1, 1).nOut(penultimateFilters).hasBias(false)
```
AlexDBlack: Should be stemFilters not penultimateFilters

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
```diff
@@ -0,0 +1,192 @@
...
+
+
+    public static String sepConvBlock(ComputationGraphConfiguration.GraphBuilder graphBuilder, int filters, int kernelSize, int stride, String blockId, String input) {
+        String prefix = "sepConvBlock"+blockId;
+
+        graphBuilder
+                .addLayer(prefix+"_act", new ActivationLayer(Activation.RELU), input)
+                .addLayer(prefix+"_sepconv1", new SeparableConvolution2D.Builder(kernelSize, kernelSize).stride(stride, stride).nOut(filters).hasBias(false)
+                        .convolutionMode(ConvolutionMode.Same).build(), prefix+"_act")
+                .addLayer(prefix+"_conv1_bn", new BatchNormalization.Builder().eps(1e-3).build(), prefix+"_sepconv1")
```
AlexDBlack: Add gamma/momentum

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
```diff
@@ -0,0 +1,192 @@
...
+                .addLayer(prefix+"_act2", new ActivationLayer(Activation.RELU), prefix+"_conv1_bn")
+                .addLayer(prefix+"_sepconv2", new SeparableConvolution2D.Builder(kernelSize, kernelSize).stride(stride, stride).nOut(filters).hasBias(false)
+                        .convolutionMode(ConvolutionMode.Same).build(), prefix+"_act2")
+                .addLayer(prefix+"_conv2_bn", new BatchNormalization.Builder().eps(1e-3).build(), prefix+"_sepconv2");
```
AlexDBlack: Add gamma/momentum

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+        Map<String, InputType> layerActivationTypes = graphBuilder.getLayerActivationTypes();
+        int[] shapeToMatch = layerActivationTypes.get(inputToMatch).getShape();
+        int[] inputShape = layerActivationTypes.get(input).getShape();
+
+        if(shapeToMatch[1] != inputShape[1]) {
+            graphBuilder
+                    .addLayer(prefix+"_relu1", new ActivationLayer(Activation.RELU), input)
+                    // tower 1
+                    .addLayer(prefix+"_avgpool1", new SubsamplingLayer.Builder(PoolingType.AVG).kernelSize(2,2).stride(2,2)
```
AlexDBlack: Is this kernel(1,1) in the original? (Not that 1,1 makes sense) - https://github.com/keras-team/keras/blob/master/keras/applications/nasnet.py#L511
Also I'd make truncate mode explicit.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
```diff
@@ -0,0 +1,192 @@
...
+                            .build(), prefix+"_relu1")
+                    .addLayer(prefix+"_conv1", new ConvolutionLayer.Builder(1,1).nOut((int) Math.floor(filters / 2))
+                            .convolutionMode(ConvolutionMode.Same).build(), prefix+"_avg_pool_1")
```
AlexDBlack: no bias here.
https://github.com/keras-team/keras/blob/master/keras/applications/nasnet.py#L513

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
```diff
@@ -0,0 +1,192 @@
...
+                    // tower 2
+                    .addLayer(prefix+"_zeropad1", new ZeroPaddingLayer(0,1), prefix+"_relu1")
+                    .addLayer(prefix+"_crop1", new Cropping2D(1,0), prefix+"_zeropad_1")
+                    .addLayer(prefix+"_avgpool2", new SubsamplingLayer.Builder(PoolingType.AVG).kernelSize(1,1).stride(2,2)
```
AlexDBlack: Make truncate (valid) mode explicit.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
```diff
@@ -0,0 +1,192 @@
...
+                            .build(), prefix+"_crop1")
+                    .addLayer(prefix+"_conv2", new ConvolutionLayer.Builder(1,1).nOut((int) Math.floor(filters / 2))
+                            .convolutionMode(ConvolutionMode.Same).build(), prefix+"_avgpool2")
```
AlexDBlack: Add strides?
https://github.com/keras-team/keras/blob/master/keras/applications/nasnet.py#L522

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
```diff
@@ -0,0 +1,192 @@
...
+                            .build(), prefix+"_concat1");
+
+            outputName = prefix+"_bn1";
+        }
+
+        if(inputShape[3] != filters) {
+            graphBuilder
+                    .addLayer(prefix+"_projection_relu", new ActivationLayer(Activation.RELU), outputName)
+                    .addLayer(prefix+"_projection_conv", new ConvolutionLayer.Builder(1,1).stride(1,1).nOut(filters)
+                            .convolutionMode(ConvolutionMode.Same).build(), prefix+"_projection_relu")
```
AlexDBlack: No bias
https://github.com/keras-team/keras/blob/master/keras/applications/nasnet.py#L536

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/NASNet.java
crockpotveggies: added explicit references where recommended. TBH we added that API for a reason, to enforce graph-wide config, so would be lots of work with little reward IMO.

deeplearning4j-zoo/src/main/java/org/deeplearning4j/zoo/model/helper/NASNetHelper.java
crockpotveggies: Fixed in next line.

crockpotveggies:
@AlexDBlack should be ready to merge. Also added InputType fixes and updated checksums and U-Net downloads now, too.

PR #5127, opened 2018-05-10
Depthwise conv2d

maxpumperla:


deeplearning4j-cuda/src/test/java/org/deeplearning4j/gradientcheck/CNNGradientCheckTest.java
```diff
@@ -926,25 +926,89 @@ public void testDeconvolution2D() {
...
-        ConvolutionMode[] cModes = new ConvolutionMode[]{ConvolutionMode.Truncate};
 
         Nd4j.getRandom().setSeed(12345);
 
         int[] ks = new int[]{1,3,1,3,1,3,1,3};
         int[] ss = new int[]{1,1,2,2,1,1,2,2};
         int[] ds = new int[]{1,1,1,1,2,2,2,2};
-        ConvolutionMode[] cms = new ConvolutionMode[]{Truncate, Truncate, Truncate, Truncate, Truncate, Truncate, Truncate, Truncate};
+        ConvolutionMode[] cms = new ConvolutionMode[]{
+                Truncate, Truncate, Truncate, Truncate, Truncate, Truncate, Truncate, Truncate};
```
crockpotveggies: What was the change here?

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/convolutional/KerasSeparableConvolution2D.java
```diff
@@ -72,9 +72,11 @@ public KerasSeparableConvolution2D(Map<String, Object> layerConfig, boolean enfo
         super(layerConfig, enforceTrainingConfig);
 
         hasBias = getHasBiasFromConfig(layerConfig, conf);
-        numTrainableParams = hasBias ? 2 : 1;
+        numTrainableParams = hasBias ? 3 : 2;
```
crockpotveggies: For my own understanding, can you explain what happened here?

crockpotveggies: Ah I think it's because we added depth multiplier.

deeplearning4j-cuda/src/test/java/org/deeplearning4j/gradientcheck/CNNGradientCheckTest.java
```diff
@@ -926,25 +926,89 @@ public void testDeconvolution2D() {
...
+                    .updater(new NoOp())
+                    .activation(Activation.TANH)
+                    .convolutionMode(cm)
+                    .list()
+                    .layer(new DepthwiseConvolution2D.Builder().name("depth-wise conv 2D layer")
+                            .cudnnAllowFallback(false)
+                            .kernelSize(k, k)
+                            .stride(s, s)
+                            .depthMultiplier(depthMultiplier)
+                            .nIn(nIn).build()); // nOut = nIn * depthMultiplier
```
AlexDBlack: We need a layer (with parameters) below the depthwise conv - say a standard conv layer, though it doesn't really mattech what - for the test to be sufficiently thorough.
Without a layer below it, the epsilons (activation gradients) out of depthwise conv could be wrong and yet the test would still pass.

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/convolutional/KerasDepthwiseConvolution2D.java
```diff
@@ -0,0 +1,228 @@
...
+
+
+/**
+ * Keras depth-wise convolution 2D layer support
+ *
+ * @author Max Pumperla
+ */
+@Slf4j
+@Data
+@EqualsAndHashCode(callSuper = false)
```
AlexDBlack: Shouldn't this be callSuper = true?

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/DepthwiseConvolution2D.java
```diff
@@ -0,0 +1,178 @@
...
+            throw new IllegalArgumentException("Stride should include stride for rows and columns (a 2d array)");
+        this.stride = builder.stride;
+        if (builder.padding.length != 2)
+            throw new IllegalArgumentException("Padding should include padding for rows and columns (a 2d array)");
+        this.padding = builder.padding;
+        this.cudnnAlgoMode = builder.cudnnAlgoMode;
+        this.cudnnFwdAlgo = builder.cudnnFwdAlgo;
+        this.cudnnBwdFilterAlgo = builder.cudnnBwdFilterAlgo;
+        this.cudnnBwdDataAlgo = builder.cudnnBwdDataAlgo;
+        this.nOut = this.nIn * this.depthMultiplier;
```
AlexDBlack: These 14 or so lines - the various superclass fields - are already set via super(builder) and should be removed here.
https://github.com/deeplearning4j/deeplearning4j/blob/f347e04b388dc16d356f774948bd61658b304f7e/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/ConvolutionLayer.java#L94-L113

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/DepthwiseConvolution2D.java
```diff
@@ -0,0 +1,178 @@
...
+
+        initializeConstraints(builder);
+    }
+
+    public boolean hasBias() {
+        return hasBias;
```
AlexDBlack: Unnecessary method, already defined in ConvolutionLayer

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/DepthwiseConvolution2D.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @Override
+    public DepthwiseConvolution2D clone() {
+        DepthwiseConvolution2D clone = (DepthwiseConvolution2D) super.clone();
+        if (clone.kernelSize != null)
+            clone.kernelSize = clone.kernelSize.clone();
+        if (clone.stride != null)
+            clone.stride = clone.stride.clone();
+        if (clone.padding != null)
+            clone.padding = clone.padding.clone();
```
AlexDBlack: Clone method is wrong. Kernel/stride/padding are already handled in superclass clone.
And it doesn't set depth multiplier.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/DepthwiseConvolution2D.java
```diff
@@ -0,0 +1,178 @@
...
+    @Override
+    public InputType getOutputType(int layerIndex, InputType inputType) {
+        if (inputType == null || inputType.getType() != InputType.Type.CNN) {
+            throw new IllegalStateException("Invalid input for  depth-wise convolution layer (layer name=\""
+                    + getLayerName() + "\"): Expected CNN input, got " + inputType);
+        }
+
+
+        return InputTypeUtil.getOutputTypeCnnLayers(inputType, kernelSize, stride, padding, dilation,
+                convolutionMode, nOut, layerIndex, getLayerName(), DepthwiseConvolution2DLayer.class);
```
AlexDBlack: This might not be correct, unless nOut is set to the correct value elsewhere?

deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/layers/convolutional/KerasSeparableConvolution2D.java
maxpumperla: no, it was just wrong before and I didn't notice. sep conv has two sets of weights (depth-wise and point-wise) plus optionally bias.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/DepthwiseConvolution2D.java
maxpumperla: that's set correctly before, it's arguably the one thing one has to really be careful about with depth conv, so thanks for flagging this.

maxpumperla:
all comments integrated.

PR #5159, opened 2018-05-14
More e2e tests, orthogonal fix

maxpumperla:


deeplearning4j-modelimport/src/test/java/org/deeplearning4j/nn/modelimport/keras/e2e/KerasModelEndToEndTest.java
```diff
@@ -542,7 +583,15 @@ private static void compareINDArrays(String label, INDArray a, INDArray b, doubl
...
         double max = diff.maxNumber().doubleValue();
         log.info(label + ": " + a.equalsWithEps(b, eps) + ", " + min + ", " + max);
-        assert (a.equalsWithEps(b, eps));
+        double threshold = 1e-7;
+        double aAbsMax = Math.max(Math.abs(a.minNumber().doubleValue()), Math.abs(a.maxNumber().doubleValue()));
+        double bAbsMax = Math.max(Math.abs(b.minNumber().doubleValue()), Math.abs(b.maxNumber().doubleValue()));
+
+        // skip too small absolute inputs
+        if (Math.abs(aAbsMax) > threshold && Math.abs(bAbsMax) > threshold) {
+            assert (a.equalsWithEps(b, eps));
```
AlexDBlack: assert is a Java feature, not a JUnit feature.
Furthermore, assertions aren't always enabled (not by default, and it's easy to forget or not realize they aren't)... so don't use this.

deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/weights/WeightInitUtil.java
```diff
@@ -65,7 +66,11 @@ public static INDArray initWeights(double fanIn, double fanOut, int[] shape, Wei
                     Distribution dist, char order, INDArray paramView) {
         switch (initScheme) {
             case DISTRIBUTION:
-                dist.sample(paramView);
+                if (dist instanceof OrthogonalDistribution) {
+                    dist.sample(paramView.reshape(shape));
```
AlexDBlack: This should be a 'reshape with order' like at the end of the method - ```paramView.reshape(order, shape);```

PR #5280, opened 2018-05-21
Update Reshape for long shapes API.

wmeddie:
Updates ScalNet code to compile with the new shape API.  

This isn't a complete update to longs for shapes because that would changes all the ScalNet APIs.


AlexDBlack:
Where are we at with this? Good to review/merge?

wmeddie:
Yes it should be ready to merge, but Jenkins has trouble building at the moment it seems.

PR #5282, opened 2018-05-21
Add Pytest as test suite, to Jumpy.

akshaysharma096:
These are the changes made.

- Add Pytest as the test suite.
- Some code was still not in PEP8 format, eg: `memory_manager.py` and `array_creation.py`, I have renamed `array_creation.py` to `test_array_creation.py` and ported  `memory_manager.py` to PEP8.
- Updated setup.py
- Updated requirements.py
- Add license as doc string in all code.


jumpy/requirements.txt
```diff
@@ -1,4 +1,9 @@
+attrs==18.1.0
 Cython==0.28.2
+more-itertools==4.1.0
```
maxpumperla: Excuse my ignorance, but I don't see (from your PR) why we need most of these new requirements. @akshaysharma096

akshaysharma096: @maxpumperla ,  `more-itertools` is in the dependency of pytest testing suite, we can verify from https://github.com/pytest-dev/pytest/blob/master/setup.py.

Thanks!

akshaysharma096:
@AlexDBlack @maxpumperla , can we merge ?

jumpy/requirements.txt
maxpumperla: exactly, but you get them with pytest. no need for attr etc.

maxpumperla:
remove redundant dependencies, then we can merge!

akshaysharma096:
@maxpumperla sure, will do.

AlexDBlack:
Thanks :)

PR #5284, opened 2018-05-21
Make CPP running accurately.

shugeo:
Modified exec for cpp using.


raver119:
Please resolve tests failures.

Issue #5229, opened 2018-05-17
Windows CUDA compilation time


raver119:
It's time to migrate from nmake and start using anything that will allow us parallel compilation on Windows. JOM, Ninja, whatever. 15 minutes for 1 cc is too much.

raver119:
cc @sshepel @saudet

saudet:
The standard option on Windows for CUDA is MSBuild. It works well enough for OpenCV and TensorFlow, and supports parallel compilation, but not for CUDA apparently... CMake seems to support Ninja for CUDA since version 3.8: https://cmake.org/cmake/help/v3.10/release/3.8.html I wonder if that does any better.

raver119:
CMake supports ninja and jom, and both do support parallel builds. So, anything that works for us here - would be just awesome to get integrated.

saudet:
Ah, JOM is still NMake: https://cmake.org/cmake/help/v3.10/manual/cmake-generators.7.html#makefile-generators  That should work too, but again no guarantees that it builds in parallel for CUDA.

raver119:
Implemented

raver119:
6 mimutes for 1 cc now, 26 minutes for all ccs

PR #5287, opened 2018-05-21
Windows warnings

raver119:
***WIP; DO NOT MERGE;***

This PR removes lots of build warnings on Windows + MSVC.
On top of that, this PR changes Windows build system to Ninja, to support parallel compilation on Windows.


libnd4j/buildnativeoperations.sh
```diff
@@ -265,12 +265,18 @@ case "$OS" in
...
+        export CMAKE_COMMAND="cmake -G \"Ninja\""
+        export MAKE_COMMAND="ninja"
+        export CC="cl.exe"
+        export CXX="cl.exe"
+        PARALLEL="true"
     else
         export CMAKE_COMMAND="cmake -G \"MSYS Makefiles\""
         export MAKE_COMMAND="make"
+
+        # Sam, do we really need this?
```
saudet: Probably not... BTW. why not use ninja for everything, not just Windows CUDA?

raver119: I was solving 1 particular problem: CUDA compilation time on Windows.
For other platforms we have regular make, which works just fine afaik. Only Windows+CUDA had sequential builds.

raver119: p.s. Ninja is available as package for MinGW.

saudet: Meh, I guess we can switch all to ninja gradually :) @sshepel ?

raver119: Well, if build script code will be the same for all platforms - that would be good definitely. We just need to make sure there's no issues with Ninja availability on other platforms: Linux, macOS.

saudet:
Yes, pretty sure, it's in the release notes as I wrote here: https://github.com/deeplearning4j/deeplearning4j/issues/5229#issuecomment-390096228

PR #5291, opened 2018-05-22
Add changes required for CI

sshepel:


pom.xml
```diff
@@ -50,19 +50,6 @@
...
-        <module>nd4j</module>
-        <module>datavec</module>
-        <module>deeplearning4j</module>
-        <module>arbiter</module>
-<!--        <module>nd4s</module>-->
-        <module>gym-java-client</module>
-        <module>rl4j</module>
-        <module>scalnet</module>
-        <module>jumpy</module>
-    </modules>
```
AlexDBlack: My only issue here is moving these to a profile only (though I get why you're doing it).
Maybe it's unavoidable (I don't have a better solution), but with maven's activeByDefault being as it is, as soon as we enable any other profile (for example, DL4J test backend profile) this will no longer be enabled and hence no modules will be set during the build - which could be pretty confusing for devs that don't know about explicitly adding the "all" profile when using other profiles.

libnd4j/pom.xml
```diff
@@ -169,38 +169,26 @@
...
-                    <!--</execution>-->
                     <execution>
                         <id>libnd4j.cpu.test.run</id>
                         <phase>test</phase>
                         <goals>
                             <goal>exec</goal>
                         </goals>
                         <configuration>
-                            <skip>${maven.test.skip}</skip>
+                            <skip>${maven.libnd4j.test.skip}</skip>
```
saudet: If we remove maven.test.skip from here, we should offer some other way to skip all the tests without having to specify 2 or more options on the command line.

pom.xml
saudet: @sshepel Yes, this isn't going to work well. We should keep those there. I thought we had pretty much figured out how to work around all that last week. What are the issues that you are still trying to solve?

libnd4j/pom.xml
sshepel: Got it, let me try to fix it.

pom.xml
sshepel: @AlexDBlack I know that, but didn't find any more elegant solution for that ATM, but I think I have one more thing to try.
@saudet yeap, those test cases that we try were not working in real build...
In a nutshell: for backend (libnd4j, nd4j) we need to have reactor that only contains those to modules and their dependencies, otherwise even thought you can provide `-pl 'libnd4j,nd4j'` explicitly, without profiles implementation and with `-amd` option whole stack will be build, with `-am` option only those to components without `nd4j` dependencies will be build.

saudet: It's always possible to do `mvn -f libnd4j` followed `mvn -f nd4j`...

libnd4j/pom.xml
saudet: Wouldn't we be building only for libnd4j in those cases with something like `mvn -f libnd4j` anyway?

Issue #5309, opened 2018-05-24
DL4J: Failed to load saved network (ModelSerializer)


AlexDBlack:
AFAIK this only started recently.

```
    @Rule
    public TemporaryFolder testDir = new TemporaryFolder();

    @Test
    public void test() throws Exception {

        VGG16 vgg16 = VGG16.builder()
                .seed(12345)
                .build();

        ComputationGraph pretrained = (ComputationGraph) vgg16.initPretrained(PretrainedType.IMAGENET);

        //Transfer learning
        ComputationGraph newGraph = new TransferLearning.GraphBuilder(pretrained)
                .fineTuneConfiguration(new FineTuneConfiguration.Builder()
                        .updater(new Adam(1e-3))
                        .seed(12345)
                        .build())
                .removeVertexKeepConnections("predictions")
                .addLayer("predictions", new OutputLayer.Builder()
                        .nIn(4096)
                        .nOut(200)  //Tiny imagenet
                        .build(), "fc2")
                .build();


        File f = testDir.newFile();
        newGraph.save(f);

        ComputationGraph g = ComputationGraph.load(f, true);
    }
```

```
java.lang.IllegalArgumentException: Invalid parameters: expected length 135079944, got length 135079936

	at org.deeplearning4j.nn.graph.ComputationGraph.init(ComputationGraph.java:494)
	at org.deeplearning4j.util.ModelSerializer.restoreComputationGraph(ModelSerializer.java:627)
	at org.deeplearning4j.nn.graph.ComputationGraph.load(ComputationGraph.java:4053)
	at org.deeplearning4j.integration.TempTest.test(TempTest.java:49)
```

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/5323

PR #5323, opened 2018-05-25
ser/de fix for CPU

raver119:
This PR fixes:

1) Issue with ser/de of long arrays on CPU backend
2) Removes undesired spam of dtype change


raver119:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/5309

Issue #5314, opened 2018-05-24
DevOps/Mono Repo: Move all version properties to parent pom


AlexDBlack:
Now that we have a mono repo, we should align all dependency versions by moving the properties to the outer-most parent POM.
Currently, we have versions defined at the second-highest level (dl4j parent, nd4j parent, etc).

sshepel:
@AlexDBlack right, was going to do that as soon as we get builds working...

PR #5336, opened 2018-05-25
Add distributionManagement for root pom

sshepel:
Add `distributionManagement` tag to be able to deploy snapshot artifacts to oss sonatype nexus.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,4 +63,28 @@
         <module>jumpy</module>
     </modules>
 
+    <profiles>
+        <profile>
+            <id>sonatype-nexus</id>
+            <activation>
+                <property>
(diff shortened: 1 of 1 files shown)


saudet:
Could we instead simply take all the profiles from, for example, the deeplearning4j submodule, and remove them from all the submodules? Like for #5314?

sshepel:
> Could we instead simply take all the profiles from, for example, the deeplearning4j submodule, and remove them from all the submodules? Like for #5314?

Well, eventually we gonna do that, but for now, base minimum is to have snapshots uploaded, and tests for libnd4j.

Don't want to break everything at once...

sshepel:
Thanks @saudet :+1:

PR #5342, opened 2018-05-26
Add support for CUDA 9.2

### LINKED ISSUE #5308: CUDA 9.2 support
We need to introduce CUDA 9.2.

saudet:
Fixes #5308 and also cleans up the `change-*-versions.sh` scripts a bit. 

The build passes, but for some reason CUDA 9.2 itself isn't working on my machine yet.
If it works for you @raver119 though, feel free to merge!


raver119:
Thanks!

PR #5358, opened 2018-05-28
Fix deploy of snapshot artifacts

### LINKED ISSUE #5337: Mono repo maven build: libnd4j test run as part of normal build
Currently, libnd4j tests are being run as part of a normal maven build with -DskipTests

Partial build log:
https://gist.github.com/AlexDBlack/6d89bf55db1d044175ed385be55a7a57

sshepel:
Fixes #5337 and not only...

### DIFF
--- a/libnd4j/pom.xml
+++ b/libnd4j/pom.xml
@@ -665,6 +665,17 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>skipTestCompileAndRun</id>
+            <activation>
+                <property>
+                    <name>skipTests</name>
+                </property>
+            </activation>
+            <properties>
+                <maven.test.skip>true</maven.test.skip>
+            </properties>
+        </profile>
     </profiles>
 
 </project>
--- a/libnd4j/pom.xml
+++ b/libnd4j/pom.xml
@@ -58,6 +58,7 @@
         <libnd4j.cuda></libnd4j.cuda>
         <libnd4j.compute></libnd4j.compute>
         <libnd4j.classifier>${libnd4j.platform}</libnd4j.classifier>
+        <libnd4j.tests.phase>test</libnd4j.tests.phase>
         <libnd4j.test.is.release.build>true</libnd4j.test.is.release.build>
         <javacpp.version>1.4.1</javacpp.version>
     </properties>
@@ -179,12 +180,12 @@
                 <executions>
                     <execution>
                         <id>libnd4j.test.run</id>
-                        <phase>test</phase>
+                        <phase>${libnd4j.tests.phase}</phase>
                         <goals>
                             <goal>exec</goal>
                         </goals>
                         <configuration>
-                            <skip>${maven.test.skip}</skip>
+                            <skip>${libnd4j.test.skip}</skip>
                             <executable>bash</executable>
                             <workingDirectory>${basedir}/tests_cpu</workingDirectory>
                             <arguments>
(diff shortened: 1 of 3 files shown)


libnd4j/pom.xml
```diff
@@ -618,6 +620,46 @@
...
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
+                        <version>1.6.0</version>
+
+                        <executions>
+                            <execution>
+                                <id>libnd4j.test.run</id>
+                                <!--FIXME: To run tests before build process-resources was chosen-->
```
saudet: What happens at process-resources? Why do we need a profile?

sshepel: :smile: It's only workaround for now, to enable tests for libnd4j only and to run them before build, to not wait for a build itself, and get test results ASAP.

saudet: Ok, we can probably use a property to set the phase instead. Let's do that instead of using a profile.

sshepel: Hm, ok let me try, I don't mind.

libnd4j/pom.xml
```diff
@@ -58,6 +58,8 @@
         <libnd4j.cuda></libnd4j.cuda>
         <libnd4j.compute></libnd4j.compute>
         <libnd4j.classifier>${libnd4j.platform}</libnd4j.classifier>
+        <libnd4j.tests.phase>test</libnd4j.tests.phase>
```
sshepel: @saudet is this smth that you want?

saudet: Yes, looks good! Thanks

libnd4j/pom.xml
```diff
@@ -173,16 +175,19 @@
...
-                        <phase>test</phase>
+                        <!--FIXME: workaround to enable libnd4j tests-->
+                        <!--FIXME: To run tests before build process-resources was chosen-->
+                        <phase>${libnd4j.tests.phase}</phase>
                         <goals>
                             <goal>exec</goal>
                         </goals>
                         <configuration>
-                            <skip>${maven.test.skip}</skip>
+                            <!--<skip>${maven.test.skip}</skip>-->
```
saudet: Let's put this back though?

sshepel: Can't do that, because idea is to run test with `mvn install -Dmaven.skip.test`. Which means that for the rest components test will be disabled, but for libnd4j not.

saudet: We can run `mvn install -pl libnd4j` for that no?

sshepel: In general - yes, but i will be hard to track in our case...

saudet: Anyway, it doesn't matter how it's achieved, but we need to have at least one flag like `skipTests` that allows up to skip all tests, and probably through a profile, so we can have a more complex activation conditions when we need. Could you work on that? That would be a fix for issue #5337

sshepel: Yeah, that is way initially I did this via profile... Should I revert it back?

saudet: Sure, but it didn't look like it was doing that..

sshepel: Ok, drop last commit, which bring us to profile impl... This will allows us to have libnd4j tests skipping option, and tests will be run with the whole stack build.
At the same time temporary workaround to run tests during the install phase will also be there with the help of profile...

saudet: Ok, now let's make it run by default unless `skipTests` or `maven.test.skip` is set. I think it would work with a profile like this as @AlexDBlack pointed out: https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/pom.xml#L69-L91

saudet:
Ok, I see you added another profile for `skipTests`, but I think I'm missing the big picture of what this is supposed to do. When we set `libnd4j.test`, we want to run the tests twice? Why?

sshepel:
> Ok, I see you added another profile for skipTests, but I think I'm missing the big picture of what this is supposed to do. When we set libnd4j.test, we want to run the tests twice? Why?

As I said, logic under `libn4j.test` is temp solution, till test for the rest of the components will not be fixed... So, to not enable test stage for all projects and get every build failed, I've decided to add this workaround...

sshepel:
> What about adding a libnd4j.test.skip property that is set in profiles with the maven.test.skip and/or skipTests properties, but that we can also override independently?

Let me try to implement this really quick ;)

PR #5372, opened 2018-05-28
Attempt to fix SameDiff shape inference

farizrahman4u:
`Nd4j.getExecutioner().calculateOutputShape(op)` returns empty list for SameDiff ops because input arrays are not immediately available.

In this fix we pass a dummy array with the same shape as inputs and see if shape can be  inferred.


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/DynamicCustomOp.java
```diff
@@ -58,8 +58,53 @@ public DynamicCustomOp() {
         tArguments = new ArrayList<>();
     }
 
+
+    // Hack to get output shape without data
+    private Map<long[], INDArray> dummyArrays = new HashMap<>();
```
raver119: At this moment DCO isn't a long living object, so each instance of any op will have own map. What's the idea here?

farizrahman4u: The idea is we pass an INDArray (zeros) of same shape as inputs. In case we have already created an INDArray for a given shape, we re use it.

agibsonccc: What is the larger context here? We can already auto infer op shapes now in the op executioner.

farizrahman4u: But it fails for samediff ops which do not have calculateOutputShape() defined. SDVariable.getShape() returns[].

farizrahman4u:
Deprecated. Moved to #5512

Issue #5347, opened 2018-05-26
SameDiff: eval() double spending


raver119:
At this moment, on multi-output nodes SDVariable.eval() will call op multiple time.

Pseudo code to reproduce problem:
```
 SDVariable[] result = graph.unstack(...);

result[0].eval();
result[1].eval();
```

Aha! Link: https://skymindai.aha.io/features/ND4J-174

PR #5373, opened 2018-05-28
Fix SameDiff double spending

farizrahman4u:
See : #5347


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -4893,7 +4902,7 @@ public INDArray execAndEndResult(List<DifferentialFunction> ops) {
      * @return
      */
     public INDArray execAndEndResult() {
-        resolveVariablesWith(Collections.<String, INDArray>emptyMap());
+        //resolveVariablesWith(Collections.<String, INDArray>emptyMap());
```
agibsonccc: Maybe delete rather than comment out?

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -4904,7 +4913,7 @@ public INDArray execAndEndResult() {
     }
 
     public INDArray[] execAndEndResults() {
-        resolveVariablesWith(Collections.<String, INDArray>emptyMap());
+        //resolveVariablesWith(Collections.<String, INDArray>emptyMap());
```
agibsonccc: Maybe delete rather than comment out?

PR #5378, opened 2018-05-29
Add missing profile for deploy

sshepel:


saudet:
Let's consolidate all of those in the parent pom.xml!

sshepel:
> Let's consolidate all of those in the parent pom.xml!

I remember about this @saudet , but first thing first - need to setup snapshots upload...

PR #5428, opened 2018-06-01
deeplearning4j-modelexport-solr: factor out one static method

cpoerschke:


deeplearning4j/deeplearning4j-modelexport-solr/src/main/java/org/deeplearning4j/nn/modelexport/solr/ltr/model/ScoringModel.java
```diff
@@ -106,24 +106,28 @@ public float score(float[] modelFeatureValuesNormalized) {
...
+  }
+
   /**
    * Currently supports {@link MultiLayerNetwork} and {@link ComputationGraph} models.
    * Pull requests to support additional <code>org.deeplearning4j</code> models are welcome.
    */
-  public static float outputScore(Model model, float[] modelFeatureValuesNormalized) {
-
-    final INDArray input = Nd4j.create(modelFeatureValuesNormalized);
+  public static INDArray output(Model model, INDArray input) {
```
cpoerschke: Would like to factor out a public static method with roughly this signature (though not necessarily this name) so that it can be used by the `ScoringModel` class here and also the `ModelTupleStream` class proposed in add-pull-request-number-here-later. Would welcome guidance on a good class (not here) to place it in.

AlexDBlack: Apologies for the delay, I missed this one.
A method like this would be appropriate perhaps in NetworkUtils: https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/NetworkUtils.java

cpoerschke: No worries. Thanks for the `NetworkUtils.java` pointer, I'll relocate the method to there. Also happy to call it something less vague than `output` though not yet sure what a slightly more descriptive name for it might be.

PR #5451, opened 2018-06-04
SameDiff : Auto naming

farizrahman4u:


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -1496,6 +1497,33 @@ public INDArray getArr() {
 
     }
 
+    // auto naming
+
+    private Integer _var_id = 0;
```
AlexDBlack: Switch to ```int```, and use String.valueOf(int) where you currently use toString

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -1496,6 +1497,33 @@ public INDArray getArr() {
...
+    private String getNewVarName(){
+        String varName = "sd_var_" + _var_id.toString();
+        while(variableMap.containsKey(varName)){
+            _var_id++;
+            varName = "sd_var_" + _var_id.toString();
```
AlexDBlack: What about ```var_``` prefix instead of ```sd_var_```? It's shorter, and no less descriptive.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -1496,6 +1497,33 @@ public INDArray getArr() {
...
+        }
+        return varName;
+    }
+    
+    public SDVariable var(int[] shape) {
```
AlexDBlack: Switch all of these array methods to var-args - ```var(int... shape)``` etc
Maybe also a ```Preconditions.checkArgument(shape != null  && shape.length > 0, "Invalid shape: %s", shape)```

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -1496,6 +1497,33 @@ public INDArray getArr() {
...
+        return var(getNewVarName(), shape);
+    }
+    public SDVariable var(long[] shape) {
+        return var(getNewVarName(), shape);
+    }
+    public SDVariable var(long[] shape, WeightInitScheme weightInitScheme) {
```
AlexDBlack: Switch arg order so you can use varargs

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
farizrahman4u: Because `var_` maybe too common a prefix? For e.g:

```
x = sd.var(arr) // named var_0, user doens't know about this

// somewhere else:

x = sd.var("var_0", arrx) // name already taken
```

AlexDBlack: Yeah, good point. OK, sounds good :)

PR #5468, opened 2018-06-05
Fix NativeImageLoader on Android

saudet:
Makes sure to write back changes with `Indexer.release()` when not using direct indexers, as on Android.

I also updated the unit tests to also use non-direct indexers, and they pass only after the changes.

/cc @jrmerwin @johanvos


johanvos:
I tested this on Android, and it works indeed as expected now. Great work!

PR #5476, opened 2018-06-05
Fix unstack output shape

farizrahman4u:


libnd4j/include/ops/declarable/generic/parity_ops/unstack.cpp
```diff
@@ -67,7 +67,7 @@ namespace nd4j {
                 shape[e] = shape::shapeOf(tad.tadOnlyShapeInfo)[e];
 
             // remove leading 1
-            if (shape.size() == 2 && shape[0] == 1)
+            if (inShape[0] == 2 && shape.size() == 2)
```
raver119: Change needed, as discussed in pm.

PR #5497, opened 2018-06-06
Nd4j like indexing for SDVariables

farizrahman4u:


farizrahman4u:
@AlexDBlack done; except for int->long. Will do after its updated in libnd4j.

raver119:
You shouldn't worry about libnd4j in this case. On Java side int args are already longs. So all ops on java side are 100% ok with longs.

farizrahman4u:
Converted to long.

Issue #5486, opened 2018-06-06
DL4J: Add CuDNN Dropout implementation


AlexDBlack:
As per title.

AlexDBlack:
Merged a week or so ago.

PR #5501, opened 2018-06-07
Add CuDNN dropout support

AlexDBlack:
Addresses: https://github.com/deeplearning4j/deeplearning4j/issues/5486

### DIFF
--- a/deeplearning4j/deeplearning4j-cuda/src/main/java/org/deeplearning4j/nn/layers/dropout/CudnnDropoutHelper.java
+++ b/deeplearning4j/deeplearning4j-cuda/src/main/java/org/deeplearning4j/nn/layers/dropout/CudnnDropoutHelper.java
@@ -72,6 +72,7 @@ public class CudnnDropoutHelper extends BaseCudnnHelper implements DropoutHelper
 
     private CudnnDropoutContext cudnnContext = new CudnnDropoutContext();
     private DataCache states;   //"Pointer to user-allocated GPU memory that will hold random number generator states."
+    private SizeTPointer stateSizeBytesPtr;
 
 
     @Override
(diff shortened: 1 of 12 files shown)


deeplearning4j/deeplearning4j-cuda/src/main/java/org/deeplearning4j/nn/layers/dropout/CudnnDropoutHelper.java
```diff
@@ -0,0 +1,147 @@
...
+                ArrayUtil.toInts(input.stride())));
+        checkCudnn(cudnnSetTensorNdDescriptor(cudnnContext.yTensorDesc, dataType, resultArray.rank(),
+                //TODO int cast
+                ArrayUtil.toInts(resultArray.shape()),
+                ArrayUtil.toInts(resultArray.stride())));
+
+        //Reserve space
+        SizeTPointer stateSizeBytesPtr = new SizeTPointer();        //TODO correct???
+        stateSizeBytes = stateSizeBytesPtr.get();
+        checkCudnn(cudnnDropoutGetReserveSpaceSize(cudnnContext.xTensorDesc, stateSizeBytesPtr));
```
saudet: We'd need to call `stateSizeBytesPtr.get()` after `cudnnDropoutGetReserveSpaceSize()` for this to make sense.

deeplearning4j/deeplearning4j-cuda/src/main/java/org/deeplearning4j/nn/layers/dropout/CudnnDropoutHelper.java
```diff
@@ -0,0 +1,147 @@
...
+
+        //Dropout descriptor:
+        if(statesPtr == null || statesPtr.capacity() != stateSizeBytes){
+            //TODO: Is this correct? "Pointer to user-allocated GPU memory that will hold random number generator states."
+            statesPtr = new BytePointer(stateSizeBytes);
+        }
```
saudet: It probably needs memory allocated on the device. Let's try to use `DataCache` available in the parent and as used by `CudnnConvolutionHelper`, `CudnnBatchNormalizationHelper`, and `CudnnLSTMHelper`.

saudet: We should also call `states.deallocate()` just before `new DataCache()` to free up precious device memory right away.

saudet:
Using the code from Caffe2/PyTorch as reference
https://github.com/pytorch/pytorch/blob/master/caffe2/operators/dropout_op_cudnn.cc
it looks like the buffer size returned by cudnnDropoutGetReserveSpaceSize() is for cudnnDropoutForward(), while the buffer size that cudnnSetDropoutDescriptor() requires we can get that from cudnnDropoutGetStatesSize().

PR #5516, opened 2018-06-08
Fix cuDNN helpers when using HALF data type

saudet:
Fixes https://github.com/deeplearning4j/nd4j/issues/1448

Manually tested with LenetMnistExample (plus the following line as the first statement of the main() method) and CUDA 9.2 on Fedora 27 with good results.
```java
org.nd4j.linalg.factory.Nd4j.setDataType(org.nd4j.linalg.api.buffer.DataBuffer.Type.HALF);
```

/cc @wmeddie


saudet:
With cuDNN, you gotta read the fine prints...

PR #5533, opened 2018-06-08
Update 08. RNNs- Sequence Classification of Synthetic Control Data.ze…

GabeChurch:
…pp.ipynb

Made changes (original notebook would not run). Commented out seemingly invalid options. Changed the file read function to create the data in the needed format? SEE line 111, changed the "line" to "newLine", or the original newLine is simply overwritten without taking effect. Added necessary imports. Fixed the listbuffer.add() which is not supported in scala 2.11->2.12 to listbufer +=


AlexDBlack:
Thanks for the PR, looks like this one got missed :(
The tutorials have all been moved to the DL4J examples repo here: https://github.com/deeplearning4j/dl4j-examples/tree/master/tutorials

Issue #5535, opened 2018-06-09
DL4J: Add OOM info dump when OOM occurs during training/fwd


AlexDBlack:
The idea:
1. Add a try/catch around the feed-forward, fit, etc methods
2. When an OOM occurs, write out a dump with memory info to aid with debugging and memory tuning

The info this dump should write would include:

1. Memory config: Xms, Xmx, javacpp, total system
2. Version info - DL4J/ND4J/CUDA/CUDNN etc
3. Hardware info - system/GPU memory
4. Memory allocation
    a. Parameters size, gradient size (+ memory alloc)
    b. Updater state size (+ memory alloc)
    c. Workspaces sizes
5. Network configuration (inc. for example workspaces and CuDNN-related config)
6. Fwd/bwd activation sizes, per layer and total (inferred from current input size + config method)

PR #5551, opened 2018-06-11
DL4J: Add OOM Crash Dump Reporting

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/5535

The basic idea: when DL4J (MLN and CG specifically) OOMs during training or inference, write an info dump to aid in understanding why, debugging and improving memory config and/or network config.

Output (real-world ResNet50 OOM)
https://gist.github.com/AlexDBlack/c605a01dd20300e753dadd7296df6252

Logged on OOM:
(Note this will by default be written to the current working directory (path here is for unit test))
```
o.d.u.CrashReportingUtil - >>> Out of Memory Exception Detected. Memory crash dump written to: C:\Users\Alex\AppData\Local\Temp\junit7109276185443170604\junit1948375717567978708\dl4j-memory-crash-dump-1528686384749_1.txt
o.d.u.CrashReportingUtil - Memory crash dump reporting can be disabled with CrashUtil.crashDumpsEnabled(false) or using system property -Dorg.deeplearning4j.crash.reporting.enabled=false
o.d.u.CrashReportingUtil - Memory crash dump reporting output location CrashUtil.crashDumpOutputDirectory(File) or using system property -Dorg.deeplearning4j.crash.reporting.directory=<path>
```

Remaining improvements I want to make - but I think I will merge it now, and do a second pass at a later date.
1. Where appropriate, add memory use info for async iterator (maybe unnecessary with (3))
2. Add basic information/explanation to crash dump; add link to DL4J memory page
3. Add info about workspaces used in other threads

### DIFF
--- a/deeplearning4j/deeplearning4j-core/pom.xml
+++ b/deeplearning4j/deeplearning4j-core/pom.xml
@@ -190,7 +190,6 @@
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
         </dependency>
-
     </dependencies>
 
     <profiles>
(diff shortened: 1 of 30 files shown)


AlexDBlack:
@raver119 Added ```.memoryInfo(int minibatch, InputType inputType)``` convenience methods to make the same info available without a crash.

This should be good to do.

PR #5569, opened 2018-06-12
Fix bidirectional lasttimestep

maxpumperla:


maxpumperla:
@sshepel build is failing on one instance due to some javacpp issue in nd4j, which I don't at all touch with this PR.

PR #5576, opened 2018-06-12
[WIP] explicitly specify versions of slf4j transitive dependencies

turbanoff:
Old versions are not used after it.
Before and after slf4j* artifacts:
default

There is still mess with dependencies to cleanup, but I think it's better to do it in small steps.

### DIFF
--- a/datavec/datavec-spark/pom.xml
+++ b/datavec/datavec-spark/pom.xml
@@ -103,6 +103,23 @@
             </exclusions>
         </dependency>
 
+        <!--Override old versions of slf4j artifacts from spark-core dependencies-->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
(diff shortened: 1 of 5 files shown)


datavec/datavec-spark/pom.xml
```diff
@@ -103,6 +103,23 @@
             </exclusions>
         </dependency>
 
+        <!--Override old versions of slf4j artifacts from spark-core dependencies-->
+        <dependency>
+            <groupId>org.slf4j</groupId>
```
agibsonccc: I wonder if a <dependencyManagement> entry would be better here?

AlexDBlack:
I have a few questions here: though in general, later versions of dependencies are usually preferred (with some exceptions) - is this causing an issue for you in practice? (If so, what is the issue? And have you confirmed this fixes it?)

Second - I'm also concerned that we can't really force/override dependencies when it comes to spark, due to the way class paths are managed when using Spark submit (i.e., the "standard" spark dependencies are on the classpath first, by default). So if there is an issue, will it resurface when we use spark submit?

AlexDBlack:
Thanks for the PR - but we're doing a full review here: https://github.com/deeplearning4j/deeplearning4j/pull/6057
That PR will make this one obsolete.

PR #5603, opened 2018-06-15
Make rnnTimestep 3d for recurrent layers, closes #5568

### LINKED ISSUE #5568: SimpleRNN doesn't accept 2D input on rnnTimestep while LSTM does
We should either accept it on both or don't accept it on both.

maxpumperla:


deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/GravesLSTM.java
```diff
@@ -121,6 +122,8 @@ public INDArray activate(boolean training, LayerWorkspaceMgr workspaceMgr) {
     private FwdPassReturn activateHelper(final boolean training, final INDArray prevOutputActivations,
                     final INDArray prevMemCellState, boolean forBackprop, LayerWorkspaceMgr workspaceMgr) {
         assertInputSet(false);
+        Preconditions.checkState(input.shape().length == 3,
+                "3D input expected to RNN layer expected, got " + input.shape().length);
```
AlexDBlack: Good check, but this can be optimized (and is not the correct way to use preconditions). This always results in 3 objects being created (2xlong[] shape, and a String) regardless of whether an error is encountered. Instead, use this (no object creation unless an error is encountered)
```
Preconditions.checkState(input.rank() == 3, "3D ... got %s", input.rank())
```

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/LSTM.java
```diff
@@ -144,6 +145,8 @@ public INDArray activate(boolean training, LayerWorkspaceMgr workspaceMgr) {
     private FwdPassReturn activateHelper(final boolean training, final INDArray prevOutputActivations,
                     final INDArray prevMemCellState, boolean forBackprop, LayerWorkspaceMgr workspaceMgr) {
         assertInputSet(false);
+        Preconditions.checkState(input.shape().length == 3,
+                "3D input expected to RNN layer expected, got " + input.shape().length);
```
AlexDBlack: As per last comment

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/SimpleRnn.java
```diff
@@ -161,6 +162,9 @@ public INDArray activate(boolean training, LayerWorkspaceMgr workspaceMgr){
 
     private Pair<INDArray,INDArray> activateHelper(INDArray prevStepOut, boolean training, boolean forBackprop, LayerWorkspaceMgr workspaceMgr){
         assertInputSet(false);
+        Preconditions.checkState(input.shape().length == 3,
+                "3D input expected to RNN layer expected, got " + input.shape().length);
```
AlexDBlack: As per last comment

PR #5613, opened 2018-06-16
SameDiff: next round of tests + fixes

AlexDBlack:
* Many op fixes
* Reductions now supports keepDims properly
* Mapped a number of libnd4j ops to Java
* Improved tests
* Moves a bunch of op-related tests to OpValidation suite
* Improved coverage (adds libnd4j + TF coverage information)
* Reduction backprop now uses new libnd4j ops
* Adds a bunch of INDArray constructors for custom ops
* Fixes a number of failing tests

Note: ALl tests under org.nd4j.autodiff are now either passing or marked as ignored, in preparation for enabling tests as part of CI.
Still a lot of work left to get to 100% passing (0 ignored) however.
image


raver120:
I'm not quite following what you all do with accumulations. We had sum/mean/max etc for ages. Now, all out of sudden adding additional implementations for them. 

What's the idea?

AlexDBlack:
@raver119 We need keepDims option, and the new implementations do (or at least should) behave differently for 2d input (i.e., returning rank 1 matrices).
And AFAIK we can't simply switch the existing Mean etc ops to the nem implementations, due to no CUDA support (not to mention breaking backward compatibility due to shape issues).

raver119:
keepDims apparently is supported, how TF import would work otherwise? It's the question of output shape, rather then op itself.


But output format change is a good point, it's definitely time to stop using [1, X] as vector and [1, 1] as scalar

AlexDBlack:
OK, so it looks like you are correct regarding import (I can't speak to how well it's tested, however):
The option is tucked away in BaseAccumulation, not made available at all in the actual class constructors:
https://github.com/deeplearning4j/deeplearning4j/blob/74bd70162979bace33cf28b160aeb61131e13bcc/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/BaseAccumulation.java
As for the shapes, again we have this 'newFormat' flag in there, that seems to allow switching for that behaviour.

OK, so I think keeping the old reduction classes will work, if we add both keepDims and newFormat args (at least for the INDArray constructors; the SameDiff constructors should hard-code newFormat=true)

TBH I don't want to get into the rank 2/1/0 issue for ND4J generally right now, as that will require a ton of testing across all projects.

AlexDBlack:
@raver119 btw, now might be a good time to remove the ```(INDArray x, INDArray y, INDArray z, long n)``` constructors on the Java accumulations? AFAIK this ```n``` arg isn't really used/useful.
(And if we do need it in a few places, we can just infer it from x/y/z array lengths)

raver119:
Yes, for accumulations n isn't used, all lengths are calculated using shapes within actual implementation. That's legacy signatures.

AlexDBlack:
From what I understand, we have (Mean/Sum/etc) -> BaseAccumulation -> BaseOp
So we're executing the legacy path based on OpNum, not dynamic based on opName.
I've updated the opname in preparation for switching to the (recently implemented) custom op versions in the future, but it shouldn't make any difference to what is actually executed at the libnd4j level at present?

raver119:
Right, my bad. Too much text. LGTM.

AlexDBlack:
@raver119 Thanks, better to check with these things. :)

I'll merge once CI has done it's thing.

PR #5616, opened 2018-06-17
Update and rename RaspberryPi2.md to RaspberryPi3.md

shdwshard:
This updated build procedure has been tested and works with 1.0.0-beta and a Raspberry Pi 3b.

## What changes were proposed in this pull request? ##

This build procedure for Raspberry Pi has updated and tested to work with 1.0.0-beta and a Raspberry Pi 3b. Other versions haven't been tested, but I'm willing to recompile older versions if there is interest.

## How was this patch tested? ##

This procedure was tested using an Ubuntu 18.04 VM and the latest Raspian on a Raspberry Pi 3. As written, libnd4j and nd4j both successfully build platform specific artifacts for linux-armhf, and all non-platform specific artifacts also successfully build. I have not yet tested the resulting artifacts on the device.


saudet:
Thanks for the update! Let's just rename it RaspberryPi.md?

nd4j/RaspberryPi.md
```diff
@@ -0,0 +1,46 @@
...
+
+    ```
+    $ apt-get/yum install git rsync cmake
+    $ mkdir $HOME/raspberrypi
+    $ cd $HOME/raspberrypi
+    $ git clone git://github.com/raspberrypi/tools.git
+    $ export PATH=$PATH:$HOME/raspberrypi/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/bin
+    $ mkdir $HOME/raspberrypi/rootfs
+    (Replace the ip address in the following command with the IP of a raspberry pi device)
+    $ rsync -rl --delete-after --safe-links pi@(192.168.1.PI):/{lib,usr} $HOME/raspberrypi/rootfs
```
saudet: I'm pretty sure https://github.com/raspberrypi/userland provides the same thing...

nd4j/RaspberryPi.md
```diff
@@ -0,0 +1,46 @@
...
+    $ cat > $HOME/raspberrypi/pi.cmake <<END
```
saudet: Could you add that file to the repository instead?

nd4j/RaspberryPi.md
```diff
@@ -0,0 +1,46 @@
...
+    SET(CMAKE_SYSTEM_NAME Linux)
+    SET(CMAKE_SYSTEM_VERSION 1)
+    SET(CMAKE_C_COMPILER $ENV{HOME}/raspberrypi/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/bin/arm-linux-gnueabihf-gcc)
```
saudet: Let's use a variable for the path to the tools.

nd4j/RaspberryPi.md
shdwshard: I'm not sure they are.  When I have a bit of time, I can try to audit this step to see what exactly is copied, but this took the better part of 30 minutes to complete, so I'm not sure I'd advise trying to recompile all the libraries and /bin files for a pi from scratch.

nd4j/RaspberryPi.md
shdwshard: Not a big problem.

nd4j/RaspberryPi.md
saudet: All 3 like for Android here:
https://github.com/deeplearning4j/deeplearning4j/blob/master/libnd4j/cmake/android-arm.cmake
Or just 1 variable that's used for all 3. It doesn't really matter, as long as it's configurable so we can include it in a script file like this:
https://github.com/deeplearning4j/deeplearning4j/blob/master/libnd4j/buildnativeoperations.sh#L150

nd4j/RaspberryPi.md
saudet: I don't think we need all of the distribution! What do we need exactly that's not in tools and userland? Maybe we should fix that first.

shdwshard: While true, this is taken from the general instructions for how to cross-compile for a raspi.
https://stackoverflow.com/questions/19162072/installing-raspberry-pi-cross-compiler
(See the accepted answer on this question)

Not being a dev on the deeplearning4j project, asking me to customize this is a bit of a lift. It's not exactly in my domain of expertise to know what dl4j needs on this level, so I would appreciate it if someone on the team could take a look at this, otherwise it's unlikely to be completed to a satisfactory level, especially considering things like GPU support. (which currently don't work with these instructions)

saudet: Ah, I see.. Those instructions are not very good, hum.

saudet: What was the issue with the original way of doing it without having to download anything from a device?

shdwshard: I'll squash the rootfs folder and see if it compiles that way, TBH, I haven't tried, this was simply reorganizing the procedure previously listed in the RaspberryPi2.md file.

nd4j/RaspberryPi.md
shdwshard: Ahh, got it.  That makes a lot of sense for how to organize things, so I'll check out the repo locally and properly edit this branch.

nd4j/RaspberryPi.md
shdwshard: Hallelulia, it compiles.  Squashed that part of the instructions and rpi.cmake file.

nd4j/RaspberryPi.md
shdwshard: Modified and modeled after Android.

nd4j/RaspberryPi.md
shdwshard: Modified and modeled after Android.

shdwshard:
@saudet Take a look, it's a lot closer to what you're imagining now I bet.

libnd4j/buildnativeoperations.sh
```diff
@@ -137,7 +137,8 @@ fi
 
 case "$OS" in
     linux-armhf)
-    export CMAKE_COMMAND="$CMAKE_COMMAND -D CMAKE_TOOLCHAIN_FILE=$HOME/raspberrypi/pi.cmake"
+    export RPI_BIN=$HOME/raspberrypi/tools/arm-bcm2708/arm-rpi-4.9.3-linux-gnueabihf/bin/arm-linux-gnueabihf
```
saudet: Let's have users set say RPI_HOME here instead of HOME, and the same in the text below.

nd4j/RaspberryPi.md
```diff
@@ -0,0 +1,33 @@
+Please follow following instructions to build nd4j for raspberry PI:
+
+1. download cross compilation tools for Raspberry PI
+
+    ```
+    $ apt-get/yum install git rsync cmake
```
saudet: No need of rsync anymore.

shdwshard:
@saudet Modified to reflect review comments.  Please re-check.

saudet:
@sshepel Let's plan to add this build to the CI pipeline!

PR #5622, opened 2018-06-18
Add dependency on MKL-DNN and use its version of MKL by default

### LINKED ISSUE #5508: ND4J: 1x1 result of mmul is always zero
The following code snippet always produces `0.0` (instead of `6.0 = 1*0.5+2*2+3*0.5`) as a result. In fact the `mmul` seems to always return 0.0 as long as the result is a scalar. On the other hand, `mmul` seems to work fine, when the result is a matrix. For example `a.transpose().mmul(b)` returns a 3x3 matrix with the correct results.

```
val a = Nd4j.create(doubleArrayOf(1.0, 2.0, 3.0))
val b = Nd4j.create(doubleArrayOf(0.5, 2.0, 0.5))
val actual = a.mmul(b.transpose())
assertEquals(1, actual.shape()[0])
assertEquals(1, actual.shape()[1])
assertEquals(6.0, actual.getDouble(0))
```

bildschirmfoto 2018-06-07 um 14 18 03

I already asked in the Gitter chat if that's a known issue, but oddly it seems that this behavior doesn't appear on other machines. So here is my setup:

**OS:** macOS 10.13.4
**Java:** 10.0.1 (but also tested with 9.0.1 -> same result)
**ND4J-Version:** 1.0.0-beta
**ND4J-Backend:** nd4j-native-platform

And that's the output logged to the console:
```
14:15:49.019 [main] INFO  org.nd4j.linalg.factory.Nd4jBackend - Loaded [CpuBackend] backend
14:15:49.696 [main] INFO  org.nd4j.nativeblas.NativeOpsHolder - Number of threads used for NativeOps: 4
14:15:50.654 [main] INFO  org.nd4j.nativeblas.Nd4jBlas - Number of threads used for BLAS: 4
14:15:50.660 [main] INFO  o.n.l.a.o.e.DefaultOpExecutioner - Backend used: [CPU]; OS: [Mac OS X]
14:15:50.661 [main] INFO  o.n.l.a.o.e.DefaultOpExecutioner - Cores: [8]; Memory: [4,0GB];
14:15:50.661 [main] INFO  o.n.l.a.o.e.DefaultOpExecutioner - Blas vendor: [MKL]
```

saudet:
Fixes #5508 since with this we'll be bundling a free version of MKL sourced from the archives here:
https://github.com/intel/mkl-dnn/releases

It's still possible to use full MKL by setting the "org.bytedeco.javacpp.openblas.load" property to "mkl_rt".

Tested manually on LenetMnistExample and works fine.

### DIFF
--- a/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native-platform/pom.xml
+++ b/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native-platform/pom.xml
@@ -25,6 +25,11 @@
             <version>${mkl.version}-${javacpp-presets.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.bytedeco.javacpp-presets</groupId>
+            <artifactId>mkl-dnn-platform</artifactId>
+            <version>${mkl-dnn.version}-${javacpp-presets.version}</version>
+        </dependency>
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>${nd4j.backend}</artifactId>
             <version>${project.version}</version>
--- a/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
+++ b/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
@@ -52,6 +52,17 @@
 <!--            <version>${mkl.version}-${javacpp-presets.version}</version>-->
 <!--            <classifier>${dependency.platform}</classifier>-->
 <!--        </dependency>-->
+        <dependency>
+            <groupId>org.bytedeco.javacpp-presets</groupId>
+            <artifactId>mkl-dnn</artifactId>
+            <version>${mkl-dnn.version}-${javacpp-presets.version}</version>
+        </dependency>
+<!--        <dependency>-->
+<!--            <groupId>org.bytedeco.javacpp-presets</groupId>-->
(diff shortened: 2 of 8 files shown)


AlexDBlack:
The changes look fine to me - the only real issue I have here is that I'm not seeing MKL being loaded...
Windows 10, IntelliJ. I built the full stack on this branch, deleted javacpp cache, invalidated intellij cache.

Output with ```-Dorg.bytedeco.javacpp.logger=slf4j -Dorg.bytedeco.javacpp.logger.debug=true```: 
https://gist.github.com/AlexDBlack/bd07533892685db427e2d1418f2f72ab

saudet:
You'll need to add either nd4j-native-platform or mkl-dnn-platform as
dependency. The whole profile hacking gets too hairy...

AlexDBlack:
OK, there we go... after adding mkl-dnn-platform:
```
o.n.n.NativeOpsHolder - Number of threads used for NativeOps: 6
o.n.n.Nd4jBlas - Number of threads used for BLAS: 6
o.n.l.a.o.e.DefaultOpExecutioner - Backend used: [CPU]; OS: [Windows 10]
o.n.l.a.o.e.DefaultOpExecutioner - Cores: [12]; Memory: [7.1GB];
o.n.l.a.o.e.DefaultOpExecutioner - Blas vendor: [MKL]
```

(Note that I've removed my existing MKL from the path before this)

PR #5626, opened 2018-06-18
Update run_tests.sh script

sshepel:
Add logic to fix LeakSanitizer error during libnd4j tests

### DIFF
--- a/libnd4j/tests_cpu/run_tests.sh
+++ b/libnd4j/tests_cpu/run_tests.sh
@@ -1,8 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -exuo pipefail
 
 IS_RELEASE='true'
+OSARCH=$(arch)
 
 if [ -f /etc/redhat-release ]; then
     source /opt/rh/devtoolset-7/enable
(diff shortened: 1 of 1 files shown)


libnd4j/tests_cpu/run_tests.sh
```diff
@@ -34,4 +35,12 @@ parse_commandline ()
 
 parse_commandline "$@"
 
-cmake -G "Unix Makefiles" -D_RELEASE=${IS_RELEASE} && make -j4 && layers_tests/runtests --gtest_output="xml:../target/surefire-reports/TEST-results.xml"
+cmake -G "Unix Makefiles" -D_RELEASE=${IS_RELEASE} && make -j4
+
+if [[ "$OSTYPE" == "linux-gnu" && "$OSARCH" == "x86_64" ]]; then
+    # sudo is used as workaround for LeakSanitizer that requires root permissions
+    sudo bash -c './layers_tests/runtests --gtest_output="xml:../target/surefire-reports/TEST-results.xml"'
+    sudo chown -R "${USER:-jenkins}":"${USER:-jenkins}" ../target
```
saudet: I don't think it's a good idea to hard code values for the CI server here. That won't work outside Jenkins.

sshepel: If you mean `jenkins`, then it's not hard coded, because first we check `USER` variable.
But, agree that it's not the best way, at the same time this scripts mostly used by CI, as far as I know.

saudet: Well, as long as we can easily run it outside Jenkins, I'm OK with it... Did you test it outside Jenkins?

sshepel: @saudet right, tested it locally, works fine.

PR #5635, opened 2018-06-18
Reshape : Use negative ints for order

farizrahman4u:
Use `-99`, `-102` instead of `99`, `102` to represent `'c'` and `'f'` respectively.

### DIFF
--- a/libnd4j/include/ops/declarable/generic/shape/reshape.cpp
+++ b/libnd4j/include/ops/declarable/generic/shape/reshape.cpp
@@ -18,14 +18,14 @@ namespace nd4j {
                 auto argumets = block.getIArguments();
                 int argsSize = argumets->size();
 
-                REQUIRE_TRUE(argsSize >= 2, 0, "Reshape arguments should have order and at least 1 dimensions");
-
                 int e = 1;
                 char order = (char) -(*argumets)[0];
                 if (order != 'c' && order != 'f') {
                     order = x->ordering();
                     e = 0;
                 }
+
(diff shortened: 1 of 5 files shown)


AlexDBlack:
What's the idea here?
Looks like order arg is required (line 21) - so is there a benefit to making it negative?
I'm just hestiant to introduce op-specific special cases like this unless they are really required.

AlexDBlack:
> we've discussed this with fariz earlier: idea is to avoid misinterpretation of order and actual reshape dimension value.

If order is a required arg (as it seems to be, by my reading of the code), then I don't think it's needed.

Maybe it doesn't matter in practice as users won't need to use dynamiccustomop (instead will use the java class to call it), but it seems like one of those special cases that could waste dev time, by not following existing conventions.
Anyway, I won't push the issue further. Just my thoughts.

farizrahman4u:
@AlexDBlack  It is not required but optional.. which means an actual dimension of 99 can be misinterpreted as ordering. (Right now we cant reshape an array to shape (99, 10) on both ends - java and cpp)

AlexDBlack:
> It is not required but optional

Ok, so looking at the code again: it *is* required for 1 input array arg, but not 2.
Negatives don't help with the second case.

If it should be optional for both, then we also need to fix this check, as we won't be able to reshape to rank 1 matrix without always specifying an order arg otherwise:
https://github.com/deeplearning4j/deeplearning4j/blob/f1d57a86885aedc50717a4b071e9f5d43227d47c/libnd4j/include/ops/declarable/generic/shape/reshape.cpp#L21

PR #5637, opened 2018-06-18
Special PR

raver119:
another leak for serhii to test CI


sshepel:
@raver119 if you have no any important changes in this PR and it's was just for test, may I close this?

PR #5658, opened 2018-06-20
add java/tensorflow to .gitignore too

cpoerschke:
Something generates a `nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/tensorflow/OpGenOverridesOuterClass.java` and that is not yet in the repo and so helpfully (command line) git flags it up as untracked.

Should it be ignored (e.g. via this pull request) or added?


raver119:
Thanks!

PR #5673, opened 2018-06-21
Fix stack

farizrahman4u:


libnd4j/include/ops/declarable/generic/parity_ops/stack.cpp
```diff
@@ -17,15 +17,15 @@ CUSTOM_OP_IMPL(stack, -1, 1, false, 0, 0) {
 	NDArray<T>* output = OUTPUT_VARIABLE(0);
 	int dim  = block.getIArguments()->size() > 0 ? INT_ARG(0) : 0;	 
 	if(dim < 0)
- 		dim += input->rankOf() + 1;
+ 		dim += input->rankOf();
```
raver119: Why you're changing this?

farizrahman4u: For input of rank R, `dim=-1` would correspond to `dim = R-1`, not `dim = R`.

For e.g, if input is of rank 3, the dims are 0, 1, 2 and  `dim=-1` means `dim=2`.

PR #5714, opened 2018-06-26
Shyrma concat

shyrma:
refine/correct concat ff


libnd4j/include/ops/declarable/helpers/cpu/transforms.cpp
```diff
@@ -758,6 +758,34 @@ void mirrorPad(const NDArray<T>& input, const NDArray<T>& paddings, NDArray<T>&
...
+    // loop through the rest of input arrays
+    for(int i = 1; i < numOfArrs; ++i) {
+        indices[i * rank2 + 2 * axis]     = indices[2 * axis + 1 + (i-1) * rank2];                                // index start from
+        indices[i * rank2 + 2 * axis + 1] = indices[2 * axis + 1 + (i-1) * rank2] + inArrs[i]->sizeAt(axis);      // index end with (excluding)
+    }    
+
+// #pragma omp parallel for if(numOfArrs > Environment::getInstance()->elementwiseThreshold()) schedule(guided)    
+#pragma omp parallel for schedule(guided)    
+    for(int i = 0; i < numOfArrs; ++i) 
+        output((indices + i * rank2), true).assign(inArrs[i]);
```
raver119: If this will work fine for all possible cases from java, it will mean that we did some really good work with NDArray.

libnd4j/include/ops/declarable/headers/transforms.h
```diff
@@ -41,7 +41,7 @@ namespace nd4j {
         DECLARE_CONFIGURABLE_OP(invert_permutation, 1, 1, false, 0, 0);
         #endif
 
-        DECLARE_CUSTOM_OP(concat, -1, 1, false, 0, -2);
+        DECLARE_CUSTOM_OP(concat, -1, 1, false, 0, 1);
```
raver119: Why enforce it? We can just assume default axis as 0

libnd4j/include/ops/declarable/helpers/cpu/transforms.cpp
shyrma: Frankly speaking I'm not sure it will pass all java tests,
anyway we will see...

libnd4j/include/ops/declarable/headers/transforms.h
shyrma: this is due to  compatibility with tf
there is no default value for axis input argument in corresponding tf operation

libnd4j/include/ops/declarable/helpers/cpu/transforms.cpp
```diff
@@ -828,6 +856,9 @@ template void mirrorPad<float>(const NDArray<float>& input, const NDArray<float>
 template void mirrorPad<float16>(const NDArray<float16>& input, const NDArray<float16>& paddings, NDArray<float16>& output, const int mode);
 template void mirrorPad<double>(const NDArray<double>& input, const NDArray<double>& paddings, NDArray<double>& output, const int mode);
 
+template void concat<float>(const std::vector<NDArray<float>*>& inArrs, NDArray<float>& output, const int axis);
+template void concat<float16>(const std::vector<NDArray<float16>*>& inArrs, NDArray<float16>& output, const int axis);
+template void concat<double>(const std::vector<NDArray<double>*>& inArrs, NDArray<double>& output, const int axis);
```
emillynge: Would you consider adding template for IntBuffer and LongBuffer backed arrays?
With this work, it seems like I could simply drop #5686

shyrma: Hi @emillynge 
I am afraid it is too hard to provide such functionality on the present stage of libnd4j development, since this would require a lot off additional templates implementations - corresponding array templates, types conversions, assignment/copy operators and so on.

Issue #5718, opened 2018-06-27
5668 regression


raver119:
Changes to DynamicCustomOp introduced in #5668 introduced few different failures in TF imports, isolated around ops with uncertain number of outputs

Here are **different** stack traces:
https://gist.github.com/raver119/82a56374553291c942a8d8f94774bd34
https://gist.github.com/raver119/66d98a684a23f05fbc9c37f33cb92bc4
https://gist.github.com/raver119/db56c1eaf002b1978821c0ba8d5a2312

AlexDBlack:
Most of these are fixed here: https://github.com/deeplearning4j/deeplearning4j/pull/5737

Remaining failures I'm seeing are documented in more detail here: https://github.com/deeplearning4j/deeplearning4j/issues/5738

PR #5737, opened 2018-06-28
SameDiff Op/Import Fixes

AlexDBlack:
Fixes many failures in: https://github.com/deeplearning4j/deeplearning4j/issues/5718

Plus fixes some other issues and test failures.

Can be merged if CI is happy - though reviews are welcome :)


raver119:
So, all import tests pass. Time to consider some serious graphs! \o/

PR #5743, opened 2018-06-28
[WIP] serde idea

raver119:
***WIP; DO NOT MERGE;***

This PR provides custom ser/de implementation for CG/MLN using ModelSerializer

### DIFF
--- a/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
+++ b/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
@@ -4540,10 +4540,12 @@ public class ComputationGraph implements Serializable, Model, NeuralNetwork {
     private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IOException {
         val cg = ModelSerializer.restoreComputationGraph(ois, true);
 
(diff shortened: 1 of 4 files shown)


deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
```diff
@@ -4534,4 +4532,20 @@ public boolean equals(Object obj) {
...
+    private void writeObject(ObjectOutputStream oos) throws IOException {
+        ModelSerializer.writeModel(this, oos, true);
+    }
+
+    private void readObject(ObjectInputStream ois) throws ClassNotFoundException, IOException {
+        val cg = ModelSerializer.restoreComputationGraph(ois, true);
+
+        this.defaultConfiguration = cg.defaultConfiguration;
+        this.configuration = cg.configuration;
+        this.flattenedParams = cg.flattenedParams;
```
AlexDBlack: I'm wondering if this is what we want... the concern I have is that the array here is now unrelated to the layers' param tables (which should be views of this). So I think we want one of:
(a) this.flattenedParams.assign(flattenedParams) - if former exists (I think it doesn't at this point)
(b) an init() call to initialize the layers and param tables?

PR #5758, opened 2018-06-30
Added nextRecord to ExcelRecordReader to prevent a bug with hasNext() calls

yquemener:
calling nextRecord on an ExcelRecordReader would call the super() function, which results in returning raw binaries from the file.

### DIFF
--- a/datavec/datavec-excel/src/main/java/org/datavec/poi/excel/ExcelRecordReader.java
+++ b/datavec/datavec-excel/src/main/java/org/datavec/poi/excel/ExcelRecordReader.java
@@ -113,7 +113,7 @@ public class ExcelRecordReader extends FileRecordReader {
             Sheet sheet = sheetIterator.next();
             rows = sheet.rowIterator();
             Row currRow = rows.next();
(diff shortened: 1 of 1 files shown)


yquemener:
The tests failed even before that modification because a file named datavec-excel/testsheet.xlsx is nowhere to be found in the repository. That may be another thing to repair.

AlexDBlack:
@yquemener All datavec tests are passing (at least last time I checked - last weekd), and have been for a while...
The test resources you are referring to is here: https://github.com/deeplearning4j/dl4j-test-resources/tree/master/src/main/resources/datavec-excel
Just make sure you have dl4j-test-resources installed, and have the appropriate profile enabled (test-nd4j-native)

yquemener:
Thanks, the tests now run and show an error! Correcting it now.

AlexDBlack:
@yquemener can this be merged now?

yquemener:
Yes, thanks!

yquemener:
This one would be useful to Discover too: https://github.com/deeplearning4j/deeplearning4j/pull/5782

PR #5787, opened 2018-07-03
Update surefire argLine property for nd4j tests

sshepel:


sshepel:
@AlexDBlack fixed, thanks for flagging!

PR #5803, opened 2018-07-04
Fix formatting, add explicit version for surefire plugin, commented out failing test for libnd4j

sshepel:


nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
```diff
@@ -143,6 +143,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
+                    <properties>
+                        <javacpp.compilerOptions>-g -O0 -fsanitize=address</javacpp.compilerOptions>
```
saudet: We need to set an array, just like here:
https://github.com/deeplearning4j/libnd4j/blob/master/pom.xml#L114
Or here:
https://github.com/deeplearning4j/libnd4j/blob/master/pom.xml#L197

PR #5813, opened 2018-07-05
Consolidate build profiles between libnd4j and nd4j modules

saudet:
Reverts to always using the same build directories for libnd4j
to do without symbolic links, which are no longer useful now
that we use the Maven Assembly Plugin for that purpose.

Works fine for Linux with CUDA and Android, should work fine for all other platforms...


nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml
```diff
@@ -112,7 +112,6 @@
                         <includePath>${libnd4jhome}/include/indexing</includePath>
                     </includePaths>
                     <linkPaths>
-                        <linkPath>${libnd4jhome}/blasbuild/cuda-${cuda.version}/blas</linkPath>
```
sshepel: Hm, I would rather keep CUDA version, just in case if someone want to build several CUDA backends on same machine...
But, I may be wrong...

nd4j/pom.xml
```diff
@@ -417,150 +402,6 @@
         <profile>
             <id>testresources</id>
         </profile>
-        <profile>
-            <id>libnd4jhome-defined</id>
```
sshepel: Not sure, that we can remove these two profiles: libnd4jhome-defined, libnd4jhome-undefined.
Libnd4j is not a dependency of nd4j, which means that there could be a chance that developer didn't provide LIBND4J_HOME env variable...

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml
saudet: That was just a build directory created by libnd4j, but it won't use it anymore. It's all in "cuda" now like it was before I tried making this work with symbolic links.

sshepel: Yeap, I remember that, ok then.

PR #5818, opened 2018-07-05
[WIP] multilang support for autogen docs

maxpumperla:
@crockpotveggies jfyi: adding an additional template folder like `template_de` will allow you to output other target languages. of course someone has to write these templates. However, the advantage of doing this programmatic is that we can check if templates diverge etc.

Note that this does not answer the question of _how to integrate multiple languages_ into one or multiple websites. That's a separate concern. jekyll / github pages do not have really good answers out of the box, afaict. But I'm sure there's a way to do this properly.


crockpotveggies:
There happens to be a Jekyll localization plugin which I am examining for
this. Depends if Github has whitelisted it. We may be eventually forced to
migrate to an independent server since Github pages has too many "safe
mode" guards.

maxpumperla:
@crockpotveggies that's cool! please let me know how it goes. autogen can regenerate output in any shape and fashion we need.

Can we hijack this feature to support multiple _DL4J versions_ as well?

PR #5830, opened 2018-07-06
Add consolidated perform-release.sh script

saudet:
Tested on Linux and works fine. Versions get changed properly and tags are added.

I've also changed SKIP_BUILD to default to 1 because we're now using the CI servers to perform the builds themselves. Also, I haven't polished all the Maven commands needed to get it working well with builds for Android and iOS, for example, again because we're relying on the CI servers.


sshepel:
I would need to update logic for changing the version of the module according with these changes for Jenkins CI scripts.

saudet:
It should be the same as the original scripts. I'm pretty sure there's nothing to adjust?

sshepel:
Well, in general - yes, should be the same, but will see :smiley:

PR #5834, opened 2018-07-06
Shyrma factory remove

shyrma:
- get rid of NDArrayFactory class, move it's methods to NDArray and MmulHelper classes
- add concat tests for different orders of input arrays


libnd4j/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.6)
 project(libnd4j)
-set( CMAKE_VERBOSE_MAKEFILE on)
+set(CMAKE_VERBOSE_MAKEFILE OFF)
```
raver119: Please ensure @sshepel ok with this change, they've used verbose mode to avoid jenkins timeouts

libnd4j/blas/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-set( CMAKE_VERBOSE_MAKEFILE on)
+set(CMAKE_VERBOSE_MAKEFILE OFF)
```
raver119: same as above

shyrma: @sshepel is agreed, I asked him

sshepel: Oh, sorry, I though @shyrma was talking about merge... Well it would be better for us to keep build logs smaller, unless you really need those debug messages...

Issue #5823, opened 2018-07-05
tests build issue


raver119:
How CLion (or any other IDE) should be set up for tests environment these days?
I get this error on Windows now:
```
C:\msys64\mingw64\bin\g++.exe  -g -fPIC -std=c++11 -fassociative-math -funsafe-math-optimizations -fmax-errors=2 -DLIBND4J_ALL_OPS=true -fopenmp  -g -fPIC -std=c++11 -fassociative-math -funsafe-math-optimizations -fmax-errors=2   -Wl,--whole-archive CMakeFiles\runtests.dir/objects.a -Wl,--no-whole-archive  -o runtests.exe -Wl,--out-implib,libruntests.dll.a -Wl,--major-image-version,0,--minor-image-version,0 @CMakeFiles\runtests.dir\linklibs.rsp
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/7.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lstatic
collect2.exe: error: ld returned 1 exit status
```

crockpotveggies:
Thanks for submitting this issue. We've migrated our docs to its own separate repository at [deeplearning4j-docs](https://github.com/deeplearning4j/deeplearning4j-docs). I've already migrated your issue and you can find it at https://github.com/deeplearning4j/deeplearning4j-docs/issues/90. 

I'll be closing this, but feel free to comment and re-open if this is not what you intended.

*This was an automatically generated response.*

saudet:
That's strange, I'm not seeing `-lstatic` anywhere in your command or in the files of libnd4j. Where is it coming from...

saudet:
I've just tried `run_tests.sh` on Windows and there's nothing weird happening. It builds fine and the tests run OK. Are you having issues only in CLion? Or on the command line with MSYS2 as well?

raver119:
CLion only. It tries to work with CMakeList and fails with that message at
last step.

shyrma:
Hi guys!
I have a questions concerning script /deeplearning4j/libnd4j/tests_cpu/run_tests.sh:
1) if I don't want to run tests, and I want just to build, what option/flag should I use ?
2) if I don't want to build, and I want just to run all tests, what option/flag should I use ?
3) if I don't want to build, and I want just to run certain test/tests, what filter/option/flag should I use ?

Yes I'm aware about separate commands:

[...]

PR #5845, opened 2018-07-07
Few tweaks

raver119:
- few cmake fixes for #5823
- range step validation


raver119:
Please note, this PR does ***NOT *** improves linking time. It just makes possible to build tests from CLion or any similar IDE.

PR #5886, opened 2018-07-11
Update logic for libnd4j tests

sshepel:
- Remove sudo;

- Remove source /opt/rh/devtoolset-7/enable, because of new logic for tests.


sshepel:
> I feel way more happy now

Believe me, same for me :smiley:

PR #5887, opened 2018-07-12
Update version of OpenBLAS to 0.3.2.dev

saudet:
I've tested that a bit here and it seems to work fine for me.

@AlexDBlack @maxpumperla @raver119 If you guys see any weird failures over the next few days though, check this first.


maxpumperla:
cool, keras import seems unaffected this time around

sshepel:
@saudet just to clarify, we are merging this and then guys, will flag is smth wrong?

PR #5903, opened 2018-07-14
Update versions of OpenBLAS, MKL-DNN, OpenCV, and TensorFlow

saudet:
Note: Here we are reverting to OpenBLAS 0.3.0 as it's the last known working version for now.


sshepel:
LGTM!

PR #5922, opened 2018-07-18
Nd4j squeeze

farizrahman4u:


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
```diff
@@ -406,9 +406,34 @@ public static INDArray expandDims(INDArray input, int dimension) {
         long[] indexes = new long[input.rank() + 1];
         for (int i = 0; i < indexes.length; i++)
             indexes[i] = i < dimension ? shape[i] : i == dimension ? 1 : shape[i - 1];
-        return input.reshape(input.ordering(), indexes);
+        return input.reshape(indexes);
```
AlexDBlack: The order arg here is quite important given the way ND4J works with reshaped arrays being a view of the original.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
```diff
@@ -406,9 +406,34 @@ public static INDArray expandDims(INDArray input, int dimension) {
...
+        if (shape[dimension] != 1){
+            throw new ND4JIllegalStateException("Can squeeze only dimensions of size 1.");
+        }
+        long[] newShape = new long[shape.length - 1];
+        for(int i=0; i < dimension; i++){
+            newShape[i] = shape[i];
+        }
+        for(int i = dimension + 1; i < shape.length; i++){
+            newShape[i - 1] = shape[i];
+        }
```
AlexDBlack: For the preceeding 7 lines: we should be able to do this in 1 line with something like ArrayUtil.removeIndex, or similar method in something like guava or an apache commons library.
https://github.com/deeplearning4j/deeplearning4j/blob/451dd76b50355358dc176f2b704e98c43423c5b8/nd4j/nd4j-common/src/main/java/org/nd4j/linalg/util/ArrayUtil.java#L1063

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
```diff
@@ -406,9 +406,34 @@ public static INDArray expandDims(INDArray input, int dimension) {
...
+     * @param dimension the dimension to remove
+     * @return the array with dimension removed
+     */
+    public static INDArray squeeze(INDArray input, int dimension) {
+        if (dimension < 0){
+            dimension += input.rank();
+        }
+        long[] shape = input.shape();
+        if (shape[dimension] != 1){
+            throw new ND4JIllegalStateException("Can squeeze only dimensions of size 1.");
```
AlexDBlack: Couple of things here:
(a) we can use Preconditions.checkArgument/checkState - less verbose
(b) When throwing exceptions, I think it's good practice to include useful information. What dimension? What's the shape of the array? Without that, I need a debugger to get that information, which adds a lot of time required to fix it...

PR #5957, opened 2018-07-24
Add copyright headers

sshepel:


sshepel:
@AlexDBlack oh, yes you are right, I didn't noticed that I add them in commit, let me fix that really quick.

sshepel:
Thanks, @AlexDBlack !

PR #5993, opened 2018-07-30
Fix k-fold iterator, add test

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/5974


saudet:
Does that make pull #5975 obsolete?

PR #5994, opened 2018-07-30
Handle DataSource/DataProvider in Arbiter UI

eraly:
UI was throwing NPE with getDataProvider. Added a test that I manually ran and verified.


eraly:
Thank you

PR #6042, opened 2018-08-01
Small fixes for copyright headers and libnd4j pom.xml

sshepel:


sshepel:
@saudet but I fixed that formatting with IntelliJ IDEA (it supports Scala files), so I think it's smth else broke that formatting...
So, lets try to fix that this time, but if it will happen again will ignore it...

saudet:
Run the build, the formatter will format it back, and then update your branch!

sshepel:
@saudet have just did that, looks good, there is no any changes in headers.
What formatter are you referring to?

saudet:
This one: https://github.com/deeplearning4j/deeplearning4j/blob/master/scalnet/pom.xml#L190-L205

sshepel:
Ok, got it, fixed (remove scalnet changes from PR).

PR #6057, opened 2018-08-02
Fix version conflicts in dependencies

### LINKED ISSUE #6047: Dependencies review
This issue: tracking conflicting version of dependencies.
Where possible, these should be aligned.

Full dependency tree with ```mvn dependency:tree -Dverbose```
https://gist.github.com/AlexDBlack/56dc761373619f75f3dd5fced7172a29

org.objenesis:objenesis:jar:2.6
- 2.1 pulled in by commons-compress 1.16.
- 2.2 pulled in by de.javakaffee:kryo-serializers / com.esotericsoftware:minlog

com.google.guava:guava:jar:20.0:compile
- Overrides common-compress,  com.github.os72:protobuf-java-util-shaded-351 transitive version 19.0

joda-time:joda-time: 2.7
- 2.2 would be used in nd4j-instrucmentation / com.fasterxml.jackson.datatype:jackson-datatype-joda:jar:2.5.1
- 2.9.9 pulled in by apache arrow 0.9.0

org.apache.commons:commons-lang3: 3.4
- 3.3.2 pulled in by nd4j-common

org.apache.commons:commons-math3 3.6.1
- 3.4.1 direct dependency of nd4j-api
- 3.6.1 test scope dependency of nd4j-tests

org.apache.zookeeper:zookeeper: 3.4.6
- 3.3.1 pulled in by com.101tec:zkclient (nd4j-kafka)

log4j:log4j:jar: 1.2.14
- 1.2.17 pulled in by org.apache.zookeeper:zookeeper:jar:3.4.6 (nd4j-kafka)

com.google.code.gson:gson 2.8
- 2.7 provided - com.github.os72:protobuf-java-util-shaded-351

com.fasterxml.jackson.core:jackson-core: 2.5.1
- 2.7.9 pulled in by apache arrow

org.bytedeco:javacpp:jar: 1.4.2
- 1.4.1 pulled in by org.bytedeco.javacpp-presets:cuda:jar:9.1-7.1-1.4.1 etc

org.apache.httpcomponents:httpclient:jar: 4.5.2
- 4.5.1 in org.apache.httpcomponents:httpasyncclient:jar:4.1.1 (nd4j-parameter-server)

commons-codec:commons-codec:jar
- 1.10 in nd4j-common
- 1.9 in com.mashape.unirest:unirest-java (nd4j-parameter-server)

nd4j-kryo: has a bunch of provided scope dedendencies that conflict

saudet:
Probably also fixes #6047. I've corrected the dependencies highlighted by this Bash command:
```bash
$ mvn -nsu dependency:list | grep "\[INFO\]    " | sort | grep :compile | uniq > /tmp/deps
```
followed by this Python script:
```python
with open("/tmp/deps") as f:
    content = f.read().splitlines()
    newcols = None
    lastcols = content[0].split(":");
    for i in range(1, len(content)):
        cols = content[i].split(":");
        if lastcols[0] == cols[0] and lastcols[1] == cols[1]:
            newcols = cols
        else:
            if newcols != None:
                print(newcols)
            newcols = None
        lastcols = cols
```

The build passes, but I haven't started testing anything. I'm sure we'll encounter something though...

Two remaining issues are 3 set of incompatible Jackson versions, and Maven not wanting to change the version of Guava in arbiter-ui for some reason:
```
['[INFO]    com.fasterxml.jackson.core', 'jackson-annotations', 'jar', '2.8.7', 'compile']
['[INFO]    com.fasterxml.jackson.core', 'jackson-core', 'jar', '2.8.7', 'compile']
['[INFO]    com.fasterxml.jackson.core', 'jackson-databind', 'jar', '2.8.7', 'compile']
['[INFO]    com.fasterxml.jackson.dataformat', 'jackson-dataformat-yaml', 'jar', '2.8.7', 'compile']
['[INFO]    com.fasterxml.jackson.datatype', 'jackson-datatype-jdk8', 'jar', '2.5.4', 'compile']
['[INFO]    com.fasterxml.jackson.datatype', 'jackson-datatype-jsr310', 'jar', '2.5.4', 'compile']
['[INFO]    com.fasterxml.jackson.module', 'jackson-module-scala_2.11', 'jar', '2.8.7', 'compile']
['[INFO]    com.google.guava', 'guava', 'jar', '20.0', 'compile']
```


AlexDBlack:
Hm, so this looks like it'll probably work as a fix.

However, didn't we discuss doing this using a dependency management section in the parent pom? That's the approach I started with when I was looking at this - wouldn't it have been cleaner?

Looking at most of the changes here, I wouldn't expect issues.

A couple of things that I suspect will be issues:
(a) Typesafe config - 1.2.1 - that's what Spark uses, but if that leaks onto the Play classpath anywhere, I'm pretty sure Play won't work
(b) Downgrading nd4j-arrow version of jackson from 2.7.9 to 2.5.1 - jackson is notorious for this sort of thing

saudet:
We can do dependency management as a followup sure, but we first need to figure out the list of those dependencies.. It's basically just copy/pasting everything in the parent pom.xml file, and stripping out the versions in the other pom.xml files. It won't change that much from what I've already done.

So we have to live with Spark not working with Arrow and Play?

saudet:
Then again dependency management won't work for artifacts with conflicting versions such as with Jackon and Typesafe Config. It only supports a single version, so it looks like more trouble than it's worth IMO.

AlexDBlack:
I was thinking we'd do selective overrides (I'm assuming a DepemencyMangament section in child pom would override parent pom for dependencies that appear in both).
I agree the practical effect won't be much different - just cleaner and more maintainable IMO. But sure, let's look at that later.

> So we have to live with Spark not working with Arrow and Play?

Play - yes. I tried some time ago to get them to work together, and gave up.
I could only even get it working locally with a bunch of exclusions - once you add spark submit into the mix, there were just too many issues due to less control of the classpath.

TBH I haven't tested Arrow and Spark, but the Jackson versions there may well cause issues...

saudet:
I've refactored to use dependency management where possible. It is a lot cleaner, yes. I've also put typesafe.config.version back to 1.3.0. Arrow seems to be using Jackson 2.5.1 already? Anyway, we'll sort out Jackson later, but let's make sure this works for now.

AlexDBlack:
Great. I'll run some tests on this later today...

AlexDBlack:
Hm, so initial testing:
DL4J unit tests - so far so good, but still running.

DL4J UI examples:
Typesafe config: for some reason it's not paying attention to the dependency management section...
```
Exception in thread "main" java.lang.NoSuchMethodError: com.typesafe.config.Config.getIsNull(Ljava/lang/String;)Z
```
```
[INFO] |  |  |  |  |  \- (com.typesafe:config:jar:1.3.0:compile - omitted for conflict with 1.2.1)
```
(above is from ```mvn dependency:tree -Dverbose``` on examples - branch ab_snapshots)

saudet:
I was afraid of that, it doesn't look like dependency management works with transitive dependencies:
https://stackoverflow.com/questions/28312975/maven-dependencymanagement-version-ignored-in-transitive-dependencies

AlexDBlack:
Hm, that's annoying. And not very useful... :/
So yes, after doing some more reading, seems we can use DependencyManagement to enforce versions in anything that depends on our parent pom (directly or indirectly). But not for transitive which don't depend on our pom.

Which previously was not my understanding of the way DependencyManagement works, but there you... :(
(I mean even the Maven docs are misleading on this point: "A second, and very important use of the dependency management section is to control the versions of artifacts used in transitive dependencies")
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management

AlexDBlack:
ok, so I finally got to looking at this. A quick review of changes look good to me.

I have tested, without issue:
nd4j-tests (native backend)
datavec modules other than those listed below
deeplearning4j-core (so far - still running; native)
arbiter
dl4j-examples (so far: UIExample, UIStorageExample, UIStorageExample_Java7, BasicHyperparameterOptimizationExample, LenetMnistExample)

I have tested, WITH issues:
datavec-arrow, datavec-local, datavec-spark_2.11, datavec-park-inference-server, datavec-spark-inference-client - bunch of NoClassDefFoundErrors and similar (reproducible via ```mvn clean test -Ptest-nd4j-native```, haven't checked in intellij yet)


I'm currently doing more testing, and will edit this comment once more results are available.

AlexDBlack:
@saudet OK, that's probably it for testing for now.
Looks like the only major issue is Jackson + Arrow.
Did you want to push up those Jackson version changes for the arrow modules?
Then I'll do one last round of tests, and we can get this merged...

saudet:
You mean change change `jackson.version` to 2.6.5? What about `spark.jackson.version`, why do we have that there?

BTW, it looks like `scala-parser-combinators_2.10` isn't available as per the error through by the CI server. Since we're only getting conflicts between 1.0.0 and 1.0.1 for `scala-parser-combinators_2.11` maybe we can just remove that force dependency, what do you think?

AlexDBlack:
I mean keep jackson.version at 2.5.1 - and maybe add a spark2.jackson.version=2.6.5, and use spark2.jackson.version for jackson deps in nd4j-arrow and datavec-arrow.

as for why we have a spark.jackson.version being 2.4.4 - good question, it's legacy at this point. Let's get rid of that and replace where it's used with jackson.version

As for scala-parser-combinators_2.10 1.0.0/1.0.1 - sure, let's get rid of that override.

saudet:
Ok, done. Everything builds fine for Scala 2.10/2.11 and Spark 1/2. Let's see if it runs all well now...

saudet:
Ah, I see that's why we should also have it for datavec-arrow. I've put it back, looks good now.

AlexDBlack:
No, I don't thit that's the problem (and it still fails for me).

```
[INFO] |  \- com.maxmind.geoip2:geoip2:jar:2.8.1:test
[INFO] |     +- com.maxmind.db:maxmind-db:jar:1.2.2:test
[INFO] |     |  \- (com.fasterxml.jackson.core:jackson-databind:jar:2.7.9:test - omitted for conflict with 2.6.5)
[INFO] |     +- org.apache.httpcomponents:httpclient:jar:4.5.3:test
[INFO] |     |  +- org.apache.httpcomponents:httpcore:jar:4.4.6:test
[INFO] |     |  +- commons-logging:commons-logging:jar:1.2:test
[INFO] |     |  \- (commons-codec:commons-codec:jar:1.9:test - omitted for conflict with 1.10)
[INFO] |     \- (com.fasterxml.jackson.core:jackson-databind:jar:2.8.7:test - omitted for conflict with 2.6.5)
```

```
testIPAddressToCoordinatesTransform(org.datavec.local.transforms.transform.TestGeoTransforms)  Time elapsed: 0.193 sec  <<< ERROR!
java.lang.NoSuchMethodError: com.fasterxml.jackson.databind.node.ArrayNode.<init>(Lcom/fasterxml/jackson/databind/node/JsonNodeFactory;Ljava/util/List;)V
        at org.datavec.local.transforms.transform.TestGeoTransforms.testIPAddressToCoordinatesTransform(TestGeoTransforms.java:70)
```

So Geoip library requires higher than Jackson 2.6.5... not sure exactly what we should do about that.

AlexDBlack:
Alright, I see basically 2 options:
1. Set spark2.jackson.version to 2.7.0 or higher
2. Have datavec-geo be another exception - use 2.7.0 or higher

Both options aren't great, but in the absence of a better idea, perhaps the latter is the least ugly option here...

saudet:
Yeah, datavec-geo is another special one, but I thought it did get overrode, guess not. Let's force 2.8.7 there, I know that one works well

AlexDBlack:
OK, I've pushed up some additional changes, I think we're *almost* there now.

AlexDBlack:
ok, so passing: datavec, arbiter, dl4j-integration-tests, dl4j-core, nd4j (native)

It might not be everything, but maybe that's good enough - let's see what CI says about this.
And if there's any other issues, they should be obvious and easy to fix once we start full pre-release QA

PR #6063, opened 2018-08-03
Fix for transposed PCA on CUDA

sebastianarnold:
In [JcublasLapack](https://github.com/deeplearning4j/deeplearning4j/blob/152a425b93700db27deb1f9be45fd78c2b13ef09/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/linalg/jcublas/blas/JcublasLapack.java#L615) implementation, singular value decomposition is being calculated on transposed matrices in some cases:
```
        // we should transpose & adjust outputs if M<N
	// cuda has a limitation, but it's OK we know
	// 	A = U S V'
	// transpose multiply rules give us ...
	// 	A' = V S' U'
```

However, [PCA](https://github.com/deeplearning4j/deeplearning4j/blob/152a425b93700db27deb1f9be45fd78c2b13ef09/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java#L187) calls this function with `U = null`:
```
        // Note - we don't care about U 
        Nd4j.getBlasWrapper().lapack().gesvd(A, s, null, VT);
```

This leads to to a `NullPointerException` in [line 630](https://github.com/deeplearning4j/deeplearning4j/blob/152a425b93700db27deb1f9be45fd78c2b13ef09/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/linalg/jcublas/blas/JcublasLapack.java#L630) if `M<N`:
```
		vt = U.dup('f') ;
```

This PR adds a UnitTest for this case and fix for the SVD implementation in JCublas backend. With CPU backend, everything looks good.

### DIFF
--- a/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/linalg/jcublas/blas/JcublasLapack.java
+++ b/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/linalg/jcublas/blas/JcublasLapack.java
@@ -624,10 +624,13 @@ public class JcublasLapack extends BaseLapack {
 		int tmp1 = N ;
 		N = M ;
 		M = tmp1 ;
+        byte tmp2 = jobu;
+        jobu = jobvt;
+        jobvt = tmp2;
 
 		a = A.transpose().dup('f') ;
-		u = VT.dup('f') ;
-		vt = U.dup('f') ;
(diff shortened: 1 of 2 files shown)


rcorbish:
Sorry for delay, yes - looks like a good fix. Nice cath Sebastian

Issue #6056, opened 2018-08-02
VersionCheck error reading property files


DrChainsaw:
Any program I run on my windows 10 machine has org.nd4j.versioncheck.VersionCheck printing a bunch or "error reading property files". This has been going on since 1.0.0 beta (now using snapshots). From what I can tell everything still works as normal so there does not seem to be any missing files.

Example POM file:
https://github.com/DrChainsaw/AmpControl/blob/dl4j-snapshot/pom.xml

Example program and output:
public static void main(String[] args) {
        System.out.println(Nd4j.create(1));
}

14:58:11.409 [main] INFO  org.nd4j.linalg.factory.Nd4jBackend - Loaded [JCublasBackend] backend
14:58:11.528 [main] DEBUG org.nd4j.versioncheck.VersionCheck - Error reading property files for jar:file:///E:/Software%2520projects/java/maven_repo/org/datavec/datavec-api/1.0.0-SNAPSHOT/datavec-api-1.0.0-20180802.081646-870.jar!/ai/skymind/org.datavec-datavec-api-git.properties
14:58:11.528 [main] DEBUG org.nd4j.versioncheck.VersionCheck - Error reading property files for jar:file:///E:/Software%2520projects/java/maven_repo/org/datavec/datavec-data-audio/1.0.0-SNAPSHOT/datavec-data-audio-1.0.0-20180802.081408-864.jar!/ai/skymind/org.datavec-datavec-data-audio-git.properties
14:58:11.528 [main] DEBUG org.nd4j.versioncheck.VersionCheck - Error reading property files for jar:file:///E:/Software%2520projects/java/maven_repo/org/datavec/datavec-data-image/1.0.0-SNAPSHOT/datavec-data-image-1.0.0-20180802.081649-872.jar!/ai/skymind/org.datavec-datavec-data-image-git.properties
14:58:11.528 [main] DEBUG org.nd4j.versioncheck.VersionCheck - Error reading property files for jar:file:///E:/Software%2520projects/java/maven_repo/org/deeplearning4j/deeplearning4j-common/1.0.0-SNAPSHOT/deeplearning4j-common-1.0.0-20180802.081438-622.jar!/ai/skymind/org.deeplearning4j-deeplearning4j-common-git.properties
[...]

PR #6081, opened 2018-08-04
#6056 Fix spaces in path for version check

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/6056


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/versioncheck/VersionInfo.java
```diff
@@ -94,7 +94,8 @@ public VersionInfo(URI uri) throws IOException {
 
         //Extract values from properties file:
         Properties properties = new Properties();
-        try (InputStream is = uri.toURL().openStream() ) {
+        URL u = new URL(path);  //Can't use URI.toUrl() due to spaces in path
```
agibsonccc: @AlexDBlack can you add a constant to this representing the character, and then clean up the comments? I'd be happy with this after that.

PR #6092, opened 2018-08-06
Fix copyright headers for nd4j and deeplearning4j-dataimport-solrj

sshepel:


AlexDBlack:
What's our policy on copyright headers for test resources?
I did notice those before merging, but wasn't sure that we needed a header there?

saudet:
I don't think we need to worry about tests, but if it's something we actually created ourselves, it can't hurt to have one.

sshepel:
In general all test classes (I think and resources too) should have CH...
@saudet ?

AlexDBlack:
Some test resources physically can't (without breaking test)
but I guess "where possible" as a policy makes sense...

sshepel:
@AlexDBlack yeap, where possible...

PR #6098, opened 2018-08-07
ND4J: Move nd4j-native test dependencies into profiles

saudet:
To fix build issues when building for the CUDA backend only.

`mvn -nsu dependency:tree -Dmaven.test.skip` and `mvn -nsu dependency:tree -Ptestresources` relieves the changes to be effective.


saudet:
Thanks! How's it look now? Also added a couple of unrelated fixes...

PR #6119, opened 2018-08-09
Update GITTER_GUIDELINES.md

TheMatthew:
Fix quickstart guide link


AlexDBlack:
Thanks for the PR - looks like we've since added some redirects, so that http://deeplearning4j.org/quickstart now works again. I think I'll leave it as that as we won't have to update the link for new versions (it'll always redirect to latest)

PR #6139, opened 2018-08-12
Add tests skip for several modules

sshepel:


sshepel:
@AlexDBlack sounds reasonable, let me create an issue for that.

sshepel:
@raver119 @AlexDBlack ok, then just for this case I can note this locally, but for future things like that, I will create an issues, and tag them like `DevOps` or `CI`...

Issue #6154, opened 2018-08-14
Libnd4j: ATan2 op broadcast case issue


AlexDBlack:
Shapes: [3,1,4]+[1,2,4] = [3,2,4]
But atan2 op returns [3,1,4] for its shape function.

```
    @Test
    public void testAtan2BroadcastShape(){
        INDArray arr1 = Nd4j.create(new long[]{3,1,4});
        INDArray arr2 = Nd4j.create(new long[]{1,2,4});

        DynamicCustomOp op = DynamicCustomOp.builder("tf_atan2")
                .addInputs(arr1, arr2)
                .build();

        List<long[]> outShapes = Nd4j.getExecutioner().calculateOutputShape(op);
        assertEquals(1, outShapes.size());

        assertArrayEquals(Arrays.toString(outShapes.get(0)), new long[]{3,2,4}, outShapes.get(0));
    }
```

```
[3, 1, 4]: arrays first differed at element [1]; 
Expected :2
Actual   :1
```

raver119:
But this op never was broadcastable for us 😄 

Should it become one?

raver119:
p.s. it's pairwise for us. And with this shapes it should throw exception during execution.

AlexDBlack:
Yes, it's broadcastable for TF and hence needs to be for TF import :)

raver119:
Okay, easy fix.

raver119:
Fixed long ago.

Issue #6037, opened 2018-08-01
Libnd4j: broadcastable pairwise op gradient failing for edge case


AlexDBlack:
Passing for most cases, but failing for the following case:
Input 1 shape: [1,6]
Input 2 shape: [3,4,5,1]
Output shape: [3,4,5,6]
Failing ops: sub, div, rsub, rdiv

Test case in Java: (note all prior cases are passing)
https://github.com/deeplearning4j/deeplearning4j/blob/0c7d15e2cc3d2167672f6ea96dbe45dec92c1f0e/nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/opvalidation/MiscOpValidation.java#L258

https://gist.github.com/AlexDBlack/774d6a86757d0a246a28789bdd7da471

shugeo:
there are subtract, reversesubtract, divide and reversedivide broadcastable ops.

AlexDBlack:
Looking at the commit history - is this fixed by https://github.com/deeplearning4j/deeplearning4j/pull/6134 ?
Or is there more to do here?

raver119:
Hypothetically should be fixed, at least c++ tests pass now.
Please confirm fix on your end.

PR #6157, opened 2018-08-14
Broadcastable Atan2

raver119:
Fix for #6154


raver119:
Probably won't be merged before #6037 fixed

PR #6190, opened 2018-08-17
PI model update

raver119:
- ParallelInference now able to update model in runtime
- fix for bad import in LSTMHelpers

cc @agibsonccc


raver119:
Model update works, but i think that we should probably check actual nIn, and if it doesn't match new model - we should notify user about that, or, just gracefully reinit PI

PR #6204, opened 2018-08-18
libnd4j: Link with MKL-DNN and OpenBLAS when available from Maven

saudet:
I plan to add support for MKL-DNN to libnd4j by basically using Maven as our repository for native libraries as well. That'll probably eventually include BLAS and LAPACK too, so I've also added OpenBLAS here to test as a fallback for platforms that don't have MKL.

In the case of Windows, Intel doesn't test MKL-DNN with GCC, not sure if it would be a good idea to build it with it, but if we're stuck with MSVC, this means we might have to switch libnd4j to MSVC as well, without good OpenMP support, or switch to the Intel C++ Compiler...

Anyway, for now I've just added a couple of `#include` in `NativeOps.h` to test that it works, and then it's just about using them like any other native dependency.


saudet:
Ok, this is building and running fine for me on Android, Linux, Mac, Windows, so I think it's good to merge!

@sshepel Jenkins is not picking up the updated parent pom.xml. Could you fix that so we can have the CI server test it a bit more?

sshepel:
@saudet sure, let me check that.

saudet:
Actually, the issue seems to be that it's using an old version of nd4j-native to build the rest of the modules...

sshepel:
@saudet oh, I can agree with that, because we had some problems with failing builds for master branch, but now it's ok, so lets fire a new build.

saudet:
@sshepel It's failing here and there in pretty much the same way, but I don't think I'm seeing any failures related to this PR, so I'm merging this.

PR #6254, opened 2018-08-23
VGG offset values are wrong (listed in RGB order)

monktastic:


monktastic:
Not sure why the build is failing. Not familiar with the project build yet.

Issue #6299, opened 2018-08-28
MnistDatasetFetcher shuffles on reset causing unexpected behavior 


GuutBoy:
If the `shuffle` parameter is set on the `MnistDatasetFetcher` it will shuffle the dataset on each call to `reset`. Below is the relevant code from `MnistDatasetFetcher`: 
```java
@Override
public void reset() {
  cursor = 0;
  curr = null;
  if (shuffle)
    MathUtils.shuffleArray(order, rng);
}
```
This appears to be a bug. In particular, this seem to cause an `MnistDatasetIterator` constructed with `numExamples` < 60000 to iterate over a new dataset on each call to `reset()`. The documentation is not super explicit but this does not seem to be the intended behavior. 

An example of this is demonstrated by the code below:
```java
import java.io.IOException;
import org.deeplearning4j.datasets.iterator.impl.MnistDataSetIterator;
import org.nd4j.linalg.dataset.DataSet;

public class MnistDatasetExperiments {

  public static void main(String[] args) throws IOException {
    MnistDataSetIterator it = new MnistDataSetIterator(1, 1, false, false, true, 0);
    DataSet data1 = it.next();
    it.reset();
    DataSet data2 = it.next();
    if (data1.get(0).getFeatures().equals(data2.get(0).getFeatures())) {
      System.out.println("Success 😃");
    } else {
      System.out.println("Failure 😭");
    }
  }
}
```
Here we construct an iterator over a single example with a batch size of 1. I would expect this iterator to return the same dataset on the first and second call to `next` (after the call to `reset`).

Charele:
Why must we get the same dataset?
I think it isn't necessary.

AlexDBlack:
Thanks for reporting.
Fixed here: https://github.com/deeplearning4j/deeplearning4j/pull/6302

GuutBoy:
@Charele It is a design choice and of course you could go with the current behavior. Either way, the actual behavior should probably be clearly documented to avoid confusion.

In my case I wanted to train a classifier over a number of epochs on a limit set of examples. To do this I used a `MnistDatasetIterator` with `numExamples` set to, say 500, and call `fit` on a `MultiLayerNetwork` with the iterator as the argument. Now I would expect this to just train the classifier on a data set of size 500 but, because of this issue, the classifier is trained on a new random subset for each epoch. This was very confusing to me because I would end up with a very good classifier even when training on a very small data set.

Issue #6280, opened 2018-08-27
DL4J: Output Layers - validate incompatible loss/activation function combinations


AlexDBlack:
Common mistakes we should detect + warn users about:
* Softmax + nOut=1
* Softmax + XENT
* Sigmoid + MCXENT
* Non 0-1 activation fns (identity/tanh/relu etc) + XENT or MCXENT (or others that expect probabilities)

A warning would be a good start - an exception (that can be disabled via ```outputValidation(false)``` or something) might be even better - lower risk of users missing the warning.

Charele:
Why dont' throw exception directly?

AlexDBlack:
@Charele 
> Why dont' throw exception directly?

Why no exception at present? Because we just haven't implemented that.
Why no exception in the future? I'm proposing to do exactly that, but allow users to disable the exception manually. There's often edge cases we can't anticipate.

Issue #6128, opened 2018-08-10
StackVertex config class getOutputType is incorrect


AlexDBlack:
StackVertex operates on dimension 0. Hence it should not be adding size or depth:
https://github.com/deeplearning4j/deeplearning4j/blob/451dd76b50355358dc176f2b704e98c43423c5b8/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/graph/StackVertex.java#L121

https://github.com/deeplearning4j/deeplearning4j/blob/451dd76b50355358dc176f2b704e98c43423c5b8/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/graph/StackVertex.java#L171

Issue #6101, opened 2018-08-07
DataVec: improve errors/validation for images with centers outside of grid


AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/blob/451dd76b50355358dc176f2b704e98c43423c5b8/datavec/datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/objdetect/ObjectDetectionRecordReader.java#L233-L236

Currently: if image center is outside of image/grid, user gets an obscure array index exception.

PR #6302, opened 2018-08-29
DL4J: Misc fixes

AlexDBlack:
More javadoc link formatting fixes.

Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/6299
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/6128
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/6101
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/6280

### DIFF
--- a/deeplearning4j/deeplearning4j-data/deeplearning4j-datasets/src/main/java/org/deeplearning4j/datasets/fetchers/MnistDataFetcher.java
+++ b/deeplearning4j/deeplearning4j-data/deeplearning4j-datasets/src/main/java/org/deeplearning4j/datasets/fetchers/MnistDataFetcher.java
@@ -62,6 +62,9 @@ public class MnistDataFetcher extends BaseDataFetcher {
     protected boolean oneIndexed = false;
     protected boolean fOrder = false; //MNIST is C order, EMNIST is F order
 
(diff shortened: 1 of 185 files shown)


deeplearning4j/deeplearning4j-data/deeplearning4j-datasets/src/main/java/org/deeplearning4j/datasets/fetchers/MnistDataFetcher.java
```diff
@@ -232,8 +236,19 @@ public void fetch(int numExamples) {
     public void reset() {
         cursor = 0;
         curr = null;
-        if (shuffle)
-            MathUtils.shuffleArray(order, rng);
+        if (shuffle) {
+            if((train && numExamples < NUM_EXAMPLES) || (!train && numExamples < NUM_EXAMPLES_TEST)){
+                //Shuffle only first N elements
+                if(firstShuffle){
+                    MathUtils.shuffleArray(order, rng);
```
GuutBoy: Maybe just do this in the last line of the constructor rather than call `reset()`. Would save some code complexity and avoid problems if a client attempts to override the public `reset()` method.

PR #6320, opened 2018-08-30
Fix dependency versions for datavec-uberjar

sshepel:


saudet:
That's only going to work with a single version of Scala and Spark, so if that's alright, ok, but what is the goal of this?

sshepel:
@saudet you mean `datavec-spark`? The goal of this PR fix broken master...

saudet:
Right, but whomever created datavec-uberjar should fix it. If nobody cares, just disable it from the CI.

sshepel:
Fixed in #6360 and #6204

PR #6350, opened 2018-09-03
[WIP] Data types

raver119:
***WIP; DO NOT MERGE;***

This PR adds:
- Additional datatypes for ND4J/libnd4j: bool, int8, int16, int32, int64 and their unsigned variants.
- Operations accept mixed dataypes, i.e.: float * bool = float
- initial String support at libnd4j graph level
- Accumulations now use intermediate dtype, to avoid precision loss.
- IndexReduce operations now return Long datatype regardless of operands


emillynge:
I cannot wait to see this feature in master!

:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

raver119:
@emillynge  Yea, we too. It's going to be a big step forward for nd4j, which was floating-point-only, same-precision-only.

PR #6360, opened 2018-09-04
rem datavec uberjar

farizrahman4u:


saudet:
Is this for #6320? @sshepel

sshepel:
@saudet right, it seems like we don't need `datavec-uberjar`...

PR #6387, opened 2018-09-06
Shyrma deconv3d

shyrma:
- implementation and testing of deconv3d op
- correct input validation in strided_slice


AlexDBlack:
If you 're using "-999" as a valid input - it's not, and these values shouldn't be used in c++ at all. I assume you're taking them from these test cases:
https://github.com/deeplearning4j/deeplearning4j/blob/2c32b5ca8b9f6189a07faabd2c7da0f19510675c/nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/opvalidation/ShapeOpValidation.java#L426

The idea with those -999 values is basically to make it obvious in the test that they aren't intended to be used - and hopefully to cause a crash if they *are* used. They don't mean anything as far as the op implementation is concerned.

Issue #6073, opened 2018-08-03
libnd4j: need segment_X backprop ops


AlexDBlack:
No backprop ops for the segment_x ops (segment_max, mean, etc)
I don't believe there's a way to do this from java.

As for the math, it should be straightforward - identical to the math for the max/mean/min/etc reductions, but executed on the segments. We can omit gradient output for the segment indices (it would be all 0s - as they aren't real valued).

https://github.com/deeplearning4j/deeplearning4j/tree/master/libnd4j/include/ops/declarable/generic/parity_ops

shugeo:
Implemented for segment_X_ ops only.
Not fully implement for unsorted_segment_X_ ops.

AlexDBlack:
Looks like unsorted segment ops all have backprop now, looking at the op implementations...
https://github.com/eclipse/deeplearning4j/tree/master/libnd4j/include/ops/declarable/generic/parity_ops

PR #6402, opened 2018-09-08
Shugeo segment bp

shugeo:
Implemented accordingly to #6073.
Note: not full implemented for unsorted_X ops.


AlexDBlack:
segment_x_bp implemented and tested in Java - passing all gradient checks... thanks @shugeo

Issue #6372, opened 2018-09-05
Arrow/Flatbuffers incompatibility


AlexDBlack:
Arrow 0.9.0/0.10.0 (and possibly earlier) uses an old/incompatible/3rd-party version of flatbuffers, and we can't simply exclude this version and use the newer/official dependency.

The issue has been reported upstream here with more details: https://issues.apache.org/jira/browse/ARROW-3175
However, it may take some time before we get a fix and a new release of Arrow.
Consequently, we need a solution in the short term.

So far the only solution (other than downgrading everything to the old/unofficial arrow version - clearly not acceptable) is to shade both Arrow *and* the 3rd party flatbuffers dependency for use in DataVec (it's not possible to shade just the flatbuffers dependency alone)

AlexDBlack:
So it looks like the API change here is pretty minimal on FlatBufferBuilder:
old version: `int createString(String s)`
new version: `int createString(CharSequence s)`

A couple of (fairly hackish) ideas here that might avoid shading the whole library are to modify just the FlatBufferBuilder class (to add createString(String)) or the Arrow Field class (to call createString(CharSequence s) instead). Either via ByteBuddy or excluding the classes from the build while substituting our own.

agibsonccc:
I wouldn't be against this if we want to just avoid a workaround for now. @saudet how hard would custom arrow presets be?

saudet:
A lot harder than ByteBudder to add one method :)

AlexDBlack:
OK, so I've spent a few hours messing around with ByteBuddy, and it was a good idea, but I don't think it'll work.
In principle, we can add the required FlatBufferBuilder method, that's not really an issue.
However, I'm seeing three blockers for this approach:
1. Modifying classes can only be done before the class is first loaded - but we won't control order of classpath loading on the user's classpath. Specifically, if the user (or any of their dependent libraries) happen to load the FlatBufferBuilder class before any/all of our classes can execute the class modification, our modifications will fail
2. The approach does not seem robust to different ways of constructing the classpath - ByteBuddy can't seem to find the class to modify when running in IntelliJ using "Shorten command line: classpath file"
3. I can't get the modified class to load (see code below) - it complains about the class already being loaded, yet I'm sure that the class is first loaded internally in bytebuddy as part of the class modification process. I'm not sure why this is occuring (I'm following the docs/examples for adding a method) and can't see a way around this "already loaded" issue.

```
        ByteBuddyAgent.install();
[...]

PR #6404, opened 2018-09-10
Fix Arrow Flatbuffers version / incompatibility

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/6372
Also upgrades to arrow 0.10.0 - seems no changes on our end were required for this.

How this works:
- Add org.nd4j:flatbuffers:
    - One relocation: com.google.flatbuffers.FlatFlatBuffersBuilder to org.nd4j.shade.FlatBufferBuilder. All other classes are not relocated/changed (but obviously are included in the shaded JAR)
    - Add new class: org.nd4j.flatbuffers.FlatBufferBuilder extends com.google.flatbuffers.FlatFlatBuffersBuilder with 1 new method. Shaded to com.google.flatbuffers.FlatFlatBuffersBuilder during the build
- Arrow's flatbuffer dependency (com.vlkan:flatbuffers) is excluded
- nd4j-api, nd4j-arrow and datavec-arrow now use org.nd4j:flatbuffers


agibsonccc:
@AlexDBlack where are we with this?

AlexDBlack:
@agibsonccc Blocked. The fix works fine, but it fails every time on CI.
The problem is that when we run the build separately from the tests, the test classpath is wrong - it still includes the original flatbuffers dependency (and not the shaded one), even though everything looks good at runtime and according to the dependency tree.
Why is it doing that? No idea. I've looked at the surefire classpath configuration, and nothing I've tried helps.

Thinking about it now, the only idea I have is to move all tests outside of nd4j-arrow module to some other module that depends on nd4j-arrow. I'll try that and see if it works.

raver119:
What about arrow guys? There were going to fix issue...

AlexDBlack:
Yeah, it's fixed on master as of about a week ago, but we can't access it until 0.11 arrow release.

AlexDBlack:
Last commit here moves tests around (out of nd4j-arrow into nd4j-tests), so should hopefully avoid the test dependency not shaded issue with the tests in nd4j-arrow.

AlexDBlack:
Relocating tests to an entirely different module didn't work.
Looks like the problem is that the maven shade plugin only runs in the package phase, which is after tests, and can't be moved to before the package phase.
And the package phase doesn't run as part of "mvn test" by default (for any modules), hence the original (pre-shade) classes are when running tests instead of the post-shade classes.
Hopefully the last commit (which adds a custom lifecycle) should fix this by forcing the package phase to run for nd4j-flatbuffers even as part of the test phase.

AlexDBlack:
Closing this. Never got this working properly for tests, and arrow 0.11.0 is out now with a fix for the underlying issue.

PR #6436, opened 2018-09-13
3D CNN loss layers

HGuillemet:
Added Cnn3DLossLayer. Most parts are copy/pasted from CnnLossLayer. Not sure if common abstract classes would be worth.
Also fixed a small typo in RecordReaderMultiDataSetIteratorTest.


AlexDBlack:
Thanks for the PR, but as noted it needs tests, and also it needs some fixes for masking.
I've pulled in your commits here (you will still be credited as a commiter) and will make the necessary changes so this can get merged: https://github.com/deeplearning4j/deeplearning4j/pull/6546

HGuillemet:
Sorry Alex. The project where I'd have use for this class is in pause, so I was not too hurry to work on the test part.

AlexDBlack:
@HGuillemet No worries, it needed some changes for the masking plus tests, but otherwise it was in pretty good shape. It is available on master/snapshots now. :)

PR #6456, opened 2018-09-18
Fix TestGeoTransforms.testIPAddressToLocationTransform failing test

sshepel:


AlexDBlack:
I'm a little hesitant to simply change the test like this, especially when it's passing with the old value on all of my machines.
Do we have any idea yet why there's a difference on CI?

saudet:
@AlexDBlack Try to remove ~/.datavec-geoip to get the latest database

PR #6478, opened 2018-09-21
Shugeo image resize bilinear

shugeo:


shugeo:
this tested only for 'c' order.

PR #6541, opened 2018-10-06
To avoid double free, call Pointer.setNull() after freeing memory in MemoryManager

saudet:


nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/jita/memory/CudaMemoryManager.java
```diff
@@ -211,8 +211,10 @@ public void memcpy(DataBuffer dstBuffer, DataBuffer srcBuffer) {
     public void release(Pointer pointer, MemoryKind kind) {
         if (kind == MemoryKind.DEVICE) {
             NativeOpsHolder.getInstance().getDeviceNativeOps().freeDevice(pointer, null);
+            pointer.setNull();
```
raver119: Not sure if we really need to modify device pointer here.

PR #6553, opened 2018-10-10
Make layerConf() public

timmolter:
In order to get later configuration parameters like kernelSize, stride, etc for `SubsamplingLayer` and any others that directly extend `AbstractLayer`, this method should be public. For cases like `ConvolutionLayer` that extend `BaseLayer` `layerConf()` is already public.


timmolter:
Cheers, thanks!

PR #6555, opened 2018-10-10
Update README, improve the quality to meet the standards

amirsojoodi:
Update README, improve the quality to meet the standards

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,3 +28,164 @@ mvn clean install -Ptestresources -Ptest-native -Dlibnd4j.cuda=x.x -Dlibnd4j.com
 We have separate repository with various examples available: https://github.com/deeplearning4j/dl4j-examples
 
 In the examples repo, you'll also find a tutorial series in Zeppelin: https://github.com/deeplearning4j/dl4j-examples/tree/master/tutorials
+
+
+
+# Detailed Instructions to build DL4J
+## Environment Setup (This instructions has been tested on Ubuntu 16.04)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,3 +28,161 @@ mvn clean install -Ptestresources -Ptest-native -Dlibnd4j.cuda=x.x -Dlibnd4j.com
 We have separate repository with various examples available: https://github.com/deeplearning4j/dl4j-examples
 
 In the examples repo, you'll also find a tutorial series in Zeppelin: https://github.com/deeplearning4j/dl4j-examples/tree/master/tutorials
+
+## Detailed Instructions to build DL4J
```
cpoerschke: Could the detailed instructions specific to DL4J perhaps live in the [nd4j/README.md](https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/README.md) file and the top-level README here would sign-post the reader to there with a link of some sort?

amirsojoodi: It's a good idea. Whatever apache developers decide :)

saudet:
Thanks! We already have a page here that's pretty much that already but it's quite outdated:
https://deeplearning4j.org/docs/latest/deeplearning4j-build-from-source
https://github.com/deeplearning4j/deeplearning4j/blob/master/docs/deeplearning4j/templates/build-from-source.md
Could you update that page instead and maybe just add a link in the README.md? That would be great.

PR #6557, opened 2018-10-11
Spark docs

AlexDBlack:


crockpotveggies:
So one step is missing here which is you'll need to add the new pages to the generation template JSON which can be found here: https://github.com/deeplearning4j/deeplearning4j/blob/master/docs/deeplearning4j-spark/pages.json

AlexDBlack:
@crockpotveggies thanks for the review. @eraly will make those corrections along with some other changes shortly in another PR.

PR #6587, opened 2018-10-15
Update dependencies to just released JavaCPP and JavaCV 1.4.3

saudet:
Also switch to CUDA 10.0 by default.


AlexDBlack:
I was seeing a number of CI failures, unrelated: https://github.com/deeplearning4j/deeplearning4j/issues/6591
Only one of concern was linux x86 libnd4j build... that crashed for some reason.
I've re-triggered CI, and I've checked local linux build (runs fine) so linux x86 failure is probably an anomaly, but I'd like to confirm that before merging.

AlexDBlack:
nd4j-native seems fine on this 2nd run, I'll merge this.

PR #6665, opened 2018-11-04
Tensorflow output name resolution

agibsonccc:
Improves graph runner output name resolution.


nd4j/nd4j-tensorflow/src/main/java/org/nd4j/tensorflow/conversion/graphrunner/GraphRunner.java
```diff
@@ -56,8 +57,7 @@
     //a status object used
     private tensorflow.TF_Status status;
     @Getter
-    private List<String> inputNames,outputNames;
```
saudet: Removing this isn't going to work for saved models, that's why I had to add this. Either we put that back or disable support for saved models.

PR #6694, opened 2018-11-10
Update invalid links for Quickstart

YuKitAs:
Some old links don't work any more and lead to 404 error, thus updated the links.


agibsonccc:
Thank you!

PR #6704, opened 2018-11-12
Mark Jetbrains annotations as provided

mkgl:
* Jetbrains annotations all have `RetentionPolicy#CLASS` or lower (`SOURCE`); they have no value at runtime.
* Merely found 9 usages throughout the monorepo; `@NotNull` most prominently. Might as well discard entirely.
  * could be coming from auto-generation—or are those _actually_ checked?

### DIFF
--- a/datavec/datavec-api/pom.xml
+++ b/datavec/datavec-api/pom.xml
@@ -32,6 +32,7 @@
           <groupId>org.jetbrains</groupId>
           <artifactId>annotations</artifactId>
           <version>${jetbrains-annotations.version}</version>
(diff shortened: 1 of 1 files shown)


raver119:
I'm afraid that's just a typo. We usually use lombok annotations, including `@NonNull`, which provides in-place check.

cc @AlexDBlack @saudet ?

AlexDBlack:
Removed here: https://github.com/deeplearning4j/deeplearning4j/pull/6706

mkgl:
awesome response time, thank you guys!

PR #6733, opened 2018-11-19
Add performance debugging page

AlexDBlack:


DrChainsaw:
Very good document. Just happened to spot that in step 9, the sentence about layersize for recurrent layers is unfinished: "For RNNs such as LSTMs, layers are typically"

PR #6782, opened 2018-11-28
Update test coordinates for TestGeoTransforms.java

### LINKED ISSUE #6780: datavec-geo test failure
The database of IP locations changes hence the test needs to be updated to reflect this.
We've seen this a couple of times before.
Perhaps we should consider a different way of running tests that doesn't need to be manually updated from time to time?

```
Failed tests: 
  TestGeoTransforms.testIPAddressToCoordinatesTransform:89 expected:<-93.4865> but was:<-93.158>
  TestGeoTransforms.testIPAddressToLocationTransform:138 expected:<[Minneapolis]> but was:<[Saint Paul]>
```

saudet:
Fixes  #6780

### DIFF
--- a/datavec/datavec-geo/src/test/java/org/datavec/api/transform/transform/TestGeoTransforms.java
+++ b/datavec/datavec-geo/src/test/java/org/datavec/api/transform/transform/TestGeoTransforms.java
@@ -79,7 +79,7 @@ public class TestGeoTransforms {
 
         String in = "128.101.101.101";
         double latitude = 44.9733;
-        double longitude = -93.4865;
+        double longitude = -93.158;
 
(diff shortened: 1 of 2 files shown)


AlexDBlack:
Hm, maybe we can use the same process as here: https://github.com/maxmind/GeoIP2-java/blob/master/src/test/java/com/maxmind/geoip2/DatabaseReaderTest.java
i.e., use the test database instead of the 'live' database?

https://github.com/maxmind/MaxMind-DB/tree/f6ed981c23b0eb33d7c07568e2177236252afda6/test-data

Then add another test to ensure the mean/real URL is still valid for downloading.

AlexDBlack:
I'll merge this as is (as the issue it fixes is stopping later tests from being run on other PRs), and will open an issue for using the test database instead.

PR #6789, opened 2018-11-29
fix initialization of parameter space in AdaGradSpace

akhodakivskiy:
This fields overrides and breaks the `setIndex` logic that uses reflection to initialize nested fields (in this case it's `learningRate` and `lrSchedule`.


AlexDBlack:
Thanks for the PR, LGTM. Apologies for the delay, I had to test this manually due instead of relying on CI.

PR #6792, opened 2018-11-29
Minor maintenance to k-fold iterator test.

RobAltena:
uncomment test for expected exception.
removes System.print.

PR #6810, opened 2018-12-06
Fix incorrect batch sizes in KFoldIterator

### LINKED ISSUE #6786: KfoldIterator calculates incorrect batch sizes
KfoldIterator calculates incorrect batch and lastBatch sizes. The later can be non-positive (lastBatch<=0). E.g.: Number of examples in `allData` `N` = 13, number of folds `k` = 6, then the current implementation calculates that `batch` = 3 and `lastBatch` = -2, which is obviously not what we wanted.

Current constructor looks like this:

```java
public KFoldIterator(int k, DataSet allData) {
	this.k = k;
	this.allData = allData.copy();
	if (k <= 1)
		throw new IllegalArgumentException();
	if (allData.numExamples() % k != 0) {
		this.batch = (int)Math.ceil(allData.numExamples() / (double)k);
		this.lastBatch = allData.numExamples() - (k-1) * this.batch;
	} else {
		this.batch = allData.numExamples() / k;
		this.lastBatch = allData.numExamples() / k;
	}
}
```

My suggested implementation is the following:

```java
public KFoldIterator(int k, DataSet allData) {
	this.k = k;
	this.allData = allData.copy();
	if (k <= 1) {
		throw new IllegalArgumentException();
	}
	this.batch = allData.numExamples() / k;
	this.lastBatch = batch + allData.numExamples() % k;
}
```

I have not tried your test (deeplearning4j/deeplearning4j/pull/5993), but I have made a simple test class that passes with my implementation, and fails with the current one. My test tries every possible `k` numbers from 2 to `N`, and sums number of test examples in the folds.

```java
import static org.junit.Assert.*;
import org.junit.Test;
import org.nd4j.linalg.api.ndarray.INDArray;
import org.nd4j.linalg.dataset.DataSet;
import org.nd4j.linalg.factory.Nd4j;

import org.nd4j.linalg.dataset.api.iterator.KFoldIterator;

public class KFoldIteratorTest {

	private static final int NUM_EXAMPLES = 13;
	private static final int NUM_FEATURES = 3;
	
	@Test
	public void testTestFold() {
		INDArray features = Nd4j.rand(new int[] {NUM_EXAMPLES, NUM_FEATURES});
		INDArray labels = Nd4j.ones(NUM_EXAMPLES, 1);
		DataSet dataSet = new DataSet(features, labels);
		
		for (int k = 2; k <= NUM_EXAMPLES; k++) {
			KFoldIterator kFoldIterator = new KFoldIterator(k, dataSet);
			int numTestExamples = 0;
			for (int i = 0; i < k; i++) {
				kFoldIterator.next();
				numTestExamples += kFoldIterator.testFold().numExamples();
			}
			assertEquals(NUM_EXAMPLES, numTestExamples);
		}
	}
	
}
```

printomi:
This PR will fix #6786 
[KFoldIterator](https://github.com/deeplearning4j/deeplearning4j/blob/release/1.0.0-beta3/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/iterator/KFoldIterator.java) has been modified following the [scikit-learn implementation](https://github.com/scikit-learn/scikit-learn/blob/55bf5d93e5674f13a1134d93a11fd0cd11aabcd1/sklearn/model_selection/_split.py#L438).

`N` number of samples are split into `k` batches. The first `(N%k)` batches contain `(N/k)+1` samples, while the remaining batches contain `(N/k)` samples. 
In case the number of samples (`N`) in the dataset is a multiple of `k`, all batches will contain `(N/k)` samples.

Visibility of private fields and methods in `KFoldIterator` has been changed to protected, so that this class can be extended in the future. (Stratified k-fold and leave-k-out cross-validation iterators would be nice.)

An unneccessary copy of the `DataSet` has been eliminated from `KFoldIterator` constructor. However, the `next()` method will maintain a copy in the fields `train` and `test`.

[Unit tests](https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/dataset/KFoldIteratorTest.java) has been modified too: enhancements, improved documentation comments, and an added new test case.

## How was this patch tested?
Manual test run in the directory `deeplearning4j/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native` with the following command:
```
mvn -e test -Dtest=KFoldIteratorTest -P testresources
```


RobAltena:
This needs to be in the source code comments too. 
> I modified KFoldIterator following the scikit-learn implementation.

The 2nd file we should modify in this PR is the KFoldIteratorTest.java.

printomi:
Thanks for the comment, I will include a comment about this major change of implementation in the author field of the `KFoldIterator` JavaDoc.
I will also use your [minor maintenance changes](https://github.com/deeplearning4j/deeplearning4j/pull/6792) in the unit test update.

printomi:
Finally, I updated `KFoldIteratorTest`, and simplified KFoldIterator so that the main logic be more straightforward.
I see that I made a conflicting change: it is nothing breaking, but a function that is placed above an inner class. I will try to resolve it by placing back this function to it's original place in the file.

printomi:
I have just removed an unneccessary copy of the DataSet from KFoldIterator.
I think that I can't improve any more this PR, so a review would be great now. 
@RobAltena , could you check if the tests pass?

printomi:
Manual test is in progress. I have just merged upstream branch. Rebuilding libnd4j, then recompiling and running tests.

printomi:
All tests passed with latest code. I used the following command in the directory `deeplearning4j/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native` to run the tests:
```
mvn -e test -Dtest=KFoldIteratorTest -P testresources
```

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/iterator/KFoldIterator.java
```diff
@@ -24,43 +24,64 @@
...
+    protected int kCursor = 0;
+    protected DataSet test;
+    protected DataSet train;
     protected DataSetPreProcessor preProcessor;
 
+    /**
+     * Create a k-fold cross-validation iterator given the dataset and k=10 train-test splits.
+     * N number of samples are split into k batches. The first (N%k) batches contain (N/k)+1 samples, while the remaining batches contain (N/k) samples. 
+     * In case the number of samples (N) in the dataset is a multiple of k, all batches will contain (N/k) samples.
+     * @param k number of folds (optional, defaults to 10)
```
RobAltena: There is no parameter k in this constructor.

PR #6883, opened 2018-12-17
[WIP] Some memory improvements

raver119:
This PR adds:
- INDArray & DataBuffer AutoCloseable (with couple of obvious restrictions)
~~- GC handling improvements for CUDA backend~~

### DIFF
--- a/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
+++ b/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
@@ -6468,7 +6468,7 @@ public abstract class BaseNDArray implements INDArray, Iterable {
 
     @Override
     public boolean closeable() {
-        if (released)
+        if (released || isAttached())
             return false;
 
         // empty arrays have no buffer at all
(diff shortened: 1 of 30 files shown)


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
```diff
@@ -6488,7 +6488,7 @@ public void close() throws Exception {
             return;
 
         if (!closeable())
-            throw new ND4JIllegalStateException("Can't release view array");
+            throw new ND4JIllegalStateException("Can't release this INDArray");
```
stolsvik: Would be great to know why it is not closeable.

nd4j/nd4j-buffer/src/main/java/org/nd4j/linalg/api/buffer/BaseDataBuffer.java
```diff
@@ -2260,22 +2263,31 @@ public boolean closeable() {
...
+        this.released = true;
+
+        for (val r:references)
+            r.markReleased();
+    }
+
     @Override
     public void close()  {
         if (!closeable())
             throw new IllegalStateException("Can't release this data buffer");
```
stolsvik: Would be great to know why it is not closeable.

Issue #6870, opened 2018-12-17
VocabConstructor.VocabRunnable hangs if ParallelTokenization == false


twhitmorenz:
Word2Vec vocabulary construction hangs if allowParallelTokenization(false) is set on the builder. This is because VocabConstructor.VocabRunnable never sets the 'done' flag, which the 'ParallelTokenization == false' path waits for. 

Example:
```
        Word2Vec.Builder vb = new Word2Vec.Builder()
                .minWordFrequency(5)
                .stopWords( stopWords)
                .layerSize(100)
                .seed(42)
                .windowSize(5)
                .epochs( epochs)
                .iterate(iter)
                .useUnknown( true)
                .tokenizerFactory( tokenizerF)
                .allowParallelTokenization(false)   // causes hang
                .workers(1);
                
        Word2Vec vec = vb.build();   // hangs
```

The easiest fix for this is to set 'done' & notify all in the finally block of VocabConstructor.run().

twhitmorenz:
This issue was also mentioned at the tail of https://github.com/deeplearning4j/deeplearning4j/issues/5784. In comment https://github.com/deeplearning4j/deeplearning4j/issues/5784#issuecomment-431063161.

twhitmorenz:
I might try and create a PR to fix this. Have cloned dl4j & got dl4j-test-resources, working on a unit-test to demonstrate this.

twhitmorenz:
Created a pull request with fix & unit test:  https://github.com/deeplearning4j/deeplearning4j/pull/6888

raver119:
Fixed/merged

PR #6888, opened 2018-12-17
fix VocabConstructor hang when ParallelTokenization == false

twhitmorenz:
resolves issue https://github.com/deeplearning4j/deeplearning4j/issues/6870, also mentioned at the tail end of issue 5784.

- set 'done' flag on completion of VocabRunnable;
- add unit test.


raver119:
Thanks for highlighting this problem and providing test.

Unfortunately mistake is a bit bigger than that, there's unwanted synchronization used, which prevents completion. So it should be fixed as well. Proper fix is coming in upcoming PR.

PR #6896, opened 2018-12-18
libnd4j: Add support for MKL-DNN

saudet:
@raver119 I'll be adding support to more ops before merging this, but do these initial changes look OK?

Reenable support for MKL-DNN that got disabled due to recent refactoring for types, plus add to more ops, including at least: conv2d/conv2d_bp/conv3dnew/conv3dnew_bp, pooling2d/pooling2d_bp/pooling3d/pooling3d_bp (avg, max, etc), batchnorm_new, lrn/lrn_bp (reference in libnd4j seems wrong though)

## How was this patch tested?

Adjusted unit tests pass.

## Quick checklist


libnd4j/include/ops/declarable/generic/helpers/impl/convolutions.cpp
```diff
@@ -415,9 +415,161 @@ else
...
+        mkldnn::memory::dims& conv_strides, mkldnn::memory::dims& conv_padding, mkldnn::memory::dims& conv_padding_r) {
+    mkldnn::memory::dims conv_src_tz = { bS, iC, iH, iW };
+    mkldnn::memory::dims conv_weights_tz = { oC, iC, kH, kW };
+    mkldnn::memory::dims conv_bias_tz = { oC };
+    mkldnn::memory::dims conv_dst_tz = { bS, oC, oH, oW };
+
+    conv_strides = { sH, sW };
+    conv_padding = { pH, pW };
+    conv_padding_r = { (oH - 1) * sH - iH + kH - pH,
+                       (oW - 1) * sW - iW + kW - pW };
```
AlexDBlack: conv_padding is 'truncate' mode padding values, and conv_padding_r is same mode padding, right?
If so, not sure on the -pH and -pW here... Usually with same mode, we ignore the ph and pW args and just calculate what we should actually use.
https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/ConvolutionMode.java#L62-L63

Also present in a bunch of other places.

libnd4j/include/ops/declarable/generic/nn/batchnorm.cpp
```diff
@@ -156,6 +191,58 @@ CUSTOM_OP_IMPL(batchnorm_new, 3, 1, false, 1, 2) {
...
+    if (block.isUseMKLDNN() && nd4j::MKLDNNStream::isSupported({input, mean, variance, gamma, beta, output}) && numOfAxes == 1) {
+        std::vector<nd4j::MKLDNNStream>& streams = block.getMKLDNNStreams();
+        if (streams.empty()) {
+            streams.push_back(MKLDNNStream("batchnorm_new"));
+        }
+
+        std::vector<Nd4jLong> shape({2, mean->lengthOf()});
+        NDArray weights = NDArrayFactory::create<float>('c', shape, block.getWorkspace());
+        weights({0, 1, 0, 0}).assign(1.0f);
+        weights({1, 2, 0, 0}).assign(0.0f);
```
AlexDBlack: What's the intention here with this weights array?

libnd4j/include/ops/declarable/generic/helpers/impl/convolutions.cpp
saudet: No, conv_padding_r is the padding on the "right", but it needs to match that equation exactly or MKL-DNN will complain: https://github.com/intel/mkl-dnn/blob/master/src/common/convolution.cpp#L116

libnd4j/include/ops/declarable/generic/nn/batchnorm.cpp
saudet: MKL-DNN needs to have both gamma and beta in one contiguous region of space that are the "weights", that's all.

saudet: Here's the bit for reference: https://github.com/intel/mkl-dnn/blob/master/include/mkldnn_types.h#L836

AlexDBlack:
@raver119 @saudet So I'm happy to merge this, but I'm seeing some CI failures in libnd4j (linux) that look unrelated. Are these also on master and can be disregarded?

raver119:
i've been running tests on master 20 minutes ago, when was going to merge reshape. All tests passed.

saudet:
I had the same tests fail on master than on this branch, so I don't think they're related. Anyway we can easily disable MKL-DNN by calling `nd4j::Environment::getInstance()->setUseMKLDNN(false)` when we want to make sure that's not the issue.

PR #6909, opened 2018-12-21
ND4J: Add support for arrays of strings to TensorflowConversion

saudet:
@raver119 Ideally, libnd4j should adopt TensorFlow's internal format so we don't need to copy buffers in TensorflowConversion.

Let's users use strings with TensorFlow models and ND4J's GraphRunner.

## How was this patch tested?

New unit test pass.

## Quick checklist

@crockpotveggies Link to CONTRIBUTING.md is broken.
@sshepel `mvn formatter:format` changes the copyright headers. We should find a way to prevent that...

### DIFF
--- a/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/linalg/jcublas/ops/executioner/CudaExecutioner.java
+++ b/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/linalg/jcublas/ops/executioner/CudaExecutioner.java
@@ -2598,7 +2598,7 @@ public class CudaExecutioner extends DefaultOpExecutioner {
         val addr = ((LongIndexer) buffer.indexer()).get(index);
         val ptr = new PagedPointer(addr);
         val str = new Nd4jCuda.utf8string(ptr);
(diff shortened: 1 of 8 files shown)


AlexDBlack:
> Ideally, libnd4j should adopt TensorFlow's internal format so we don't need to copy buffers in TensorflowConversion.

What format specifically is this?
Also TF is not the only thing we need to worry about here. I'd be much more concerned about matching the internal representation used by Arrow (which IIRC is basicaly UTF8 values sequentially in a buffer, with offset/length arrays?).

saudet:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api.h#L213
Arrow might be using something very similar too. It makes sense that some sort of standard would emerge.

AlexDBlack:
For reference, Arrow is basically the same:
"Utf8 and Binary" - https://arrow.apache.org/docs/metadata.html
* Validity bitmap (optional in arrow, would be all 1s for dense array anyway)
* list offsets (int, but not sure if signed/unsigned and 32/64)
* byte data (utf8)

saudet:
In any case, for now this PR just uses what we have in libnd4j to run graphs with TensorFlow. We'll be updating this with the rest when we get to it. So, ok to merge?

PR #6921, opened 2018-12-28
Add explicit dependency on libnd4j in nd4j-backend-impls

treo:
This allows running maven in parallel build mode (with the `-T` parameter). 

I'm setting the provided scope here, so it doesn't become a transitive dependency and should only be required at build time. 

It appears to work fine on windows with this invocation:

```
export PLATFORM=windows-x86_64

mvn -T1C -B -V -U clean install -pl '!deeplearning4j/deeplearning4j-cuda,!nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda,!nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform,!jumpy,!pydatavec,!pydl4j' -Dlibnd4j.platform=${PLATFORM} -Djavacpp.platform=${PLATFORM} -P native-snapshots -Dmaven.test.skip=true -Dmaven.javadoc.skip=true
```


treo:
Finally got around to updating my CUDA build environment and checked if cuda builds still build fine: they do! 

@raver119 when you've got the time, please try building with this and the `-T1C` option and if it works please merge :)

raver119:
Sure, will do at morning. Windows/Linux i suppose?

PR #6928, opened 2018-12-30
allow KerasTokenizer.numWords to be null

dstine:

This PR allow `KerasTokenizer.numWords` to be `null`. Keras allows `num_words=None`, in fact this is the default.

## How was this patch tested?

I ensured that Keras `TokenizerImportTest` continues to pass, and added an additional test.

## Quick checklist

### DIFF
--- a/deeplearning4j/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/preprocessing/text/KerasTokenizer.java
+++ b/deeplearning4j/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/preprocessing/text/KerasTokenizer.java
@@ -114,7 +114,7 @@ public class KerasTokenizer {
             throw new InvalidKerasConfigurationException("No configuration found for Keras tokenizer");
 
 
-        int numWords = (int) tokenizerConfig.get("num_words");
+        Integer numWords = (Integer) tokenizerConfig.get("num_words");
         String filters = (String) tokenizerConfig.get("filters");
         Boolean lower = (Boolean) tokenizerConfig.get("lower");
         String split = (String) tokenizerConfig.get("split");
--- a/deeplearning4j/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/preprocessing/text/KerasTokenizer.java
+++ b/deeplearning4j/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/preprocessing/text/KerasTokenizer.java
@@ -120,7 +120,7 @@ public class KerasTokenizer {
         String split = (String) tokenizerConfig.get("split");
         Boolean charLevel = (Boolean) tokenizerConfig.get("char_level");
         String oovToken = (String) tokenizerConfig.get("oov_token");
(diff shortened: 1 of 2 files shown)


dstine:
For the new test method to pass, a corresponding additional file must be added to the dl4j-test-resources repo. Not sure how you like to handle this kind of inter-dependency. If this change is acceptable, I can make the corresponding pull request over there. For now the new file is a simple copy of the existing `tokenizer.json` with `num_words` set to `null`.

If there are naming standards or patterns to follow (test method, test resource, etc.) please let me know.

deeplearning4j/deeplearning4j-modelimport/src/main/java/org/deeplearning4j/nn/modelimport/keras/preprocessing/text/KerasTokenizer.java
```diff
@@ -114,7 +114,7 @@ public static KerasTokenizer fromJson(String jsonFileName) throws IOException, I
             throw new InvalidKerasConfigurationException("No configuration found for Keras tokenizer");
 
 
-        int numWords = (int) tokenizerConfig.get("num_words");
+        Integer numWords = (Integer) tokenizerConfig.get("num_words");
```
dstine: In addition to fixing the NPE, this avoids an unnecessary unbox-then-immediately-rebox. (Same pattern is in use for `documentCount` and could also be changed.)

maxpumperla:
thanks, lgtm. please submit the PR with your test resource first. naming is ok, just make sure it's in the correct relative path.

maxpumperla:
p.s changing other fields like above might make sense as well

dstine:
@maxpumperla 

Thanks for the review. Please see https://github.com/deeplearning4j/dl4j-test-resources/pull/176 with addition of the test resource.

I have another commit with boxing change for `documentCount`. Unless you have a different preferred approach, I can wait to push it until the dl4j-test-resources change is merged so that the new commit will trigger the CI build against the newly-added test resource.

maxpumperla:
merged the resource pr. go ahead with the other change then, please. Let me know once it's ready from your side.

thanks again for this contribution!

dstine:
The CUDA 9.0 and 10.0 builds passed, but 9.2 failed. Seems unlikely to be related to this proposed change. How would you like to proceed?

maxpumperla:
can't be related. I'm happy to merge this

dstine:
Great, that works for me. Thanks!

A couple of questions if you have a moment:
* Any forecast on when a release with this change might be produced?
* I didn’t find a description of the relationship between dl4j-test-resources and this repo. Is that documented and I just couldn’t find it, or is that a gap?

agibsonccc:
We do not comment on release timelines for open source.. 
Dl4j test resources is only for providing datasets and test utilities for unit testing deeplearning4j.
We document that on the repo itself: https://github.com/deeplearning4j/dl4j-test-resources - upon a recent scan of our docs, we need to update this. Thanks for pointing this out cc @ShamsUlAzeem

maxpumperla:
@dstine thanks again. btw, while we do not comment on stable releases, you will find this fix in 1.0.0-SNAPSHOT likely tomorrow.

PR #6945, opened 2019-01-07
fix: dl4j RepeatVector

farizrahman4u:


raver119:
why you explicitly detach array within layer?

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/RepeatVector.java
```diff
@@ -101,8 +101,9 @@ protected INDArray preOutput(boolean training, boolean forBackprop, LayerWorkspa
         long miniBatch = input.shape()[0];
         long size = input.shape()[1];
         INDArray output = input.reshape(miniBatch, size, 1);
-
-        return output.repeat(2, (long) getN());
+        output = output.repeat(2, (long)getN());
+        output = output.detach();  // .repeat returns an attached array
+        return output;
```
AlexDBlack: The existing ```return output.repeat(2, (long) getN());``` should be replaced by:
```
try(MemoryWorkspace ws = workspaceMgr.notifyScopeBorrowed(ArrayType.ACTIVATIONS)) {
    return output.repeat(2, (long) getN());
}
```

We don't want this to be detached instead.

PR #6947, opened 2019-01-07
Unify CPU extension parameter selection

### LINKED ISSUE #6933: Use different trigger properties for avx2/avx512 extension profiles?
While trying to figure out https://github.com/deeplearning4j/deeplearning4j/pull/6932 I've come across this:

https://github.com/deeplearning4j/deeplearning4j/blob/d5465b015fd4bf9c6c13ac50a0cb51e1e8698f22/nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml#L331-L339

So we are using `javacpp.extension` here to trigger `javacpp.platform.extension`. This resulted in a bit of confusion on my side. Since we have to set `libnd4j.extension` anyway for this case, why not trigger the profile on it and skip using the javacpp namespace for that property?

/cc @saudet

treo:
This drops javacpp.extension in favor of only using libnd4j.extension
to trigger the proper profiles that will set all other required properties.

So far only two possible values are supported: avx2 and avx512.

Fixes #6933


nd4j/ci/build-linux-x86_64.sh
```diff
@@ -72,6 +72,5 @@ docker run -ti -e SONATYPE_USERNAME -e SONATYPE_PASSWORD -v $HOME/.m2:/root/.m2
         EXTRA_OPTIONS='-pl !nd4j-uberjar,!nd4j-backends/nd4j-backend-impls/nd4j-cuda,!nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform,!nd4j-backends/nd4j-tests'; \
     fi; \
     bash change-scala-versions.sh $SCALA; \
-    mvn clean $MAVEN_PHASE -B -U --settings ./ci/settings.xml -Dmaven.test.skip=true -Dlocal.software.repository=sonatype \
-        -Djavacpp.extension=${EXT:-} \$EXTRA_OPTIONS -DprotocCommand=/build/protobuf-$PROTOBUF/src/protoc;"
+    mvn clean $MAVEN_PHASE -B -U --settings ./ci/settings.xml -Dmaven.test.skip=true -Dlocal.software.repository=sonatype $EXTRA_OPTIONS -DprotocCommand=/build/protobuf-$PROTOBUF/src/protoc;"
```
saudet: The \ in front of $EXTRA_OPTIONS is important here.

Issue #6907, opened 2018-12-20
Eclipse: Need more information concerning intellectual property


saudet:
As an Eclipse project, we need to have a clear and thorough understanding of the IP that is found in this repository. Most the files are under Skymind's copyright, but some are not. I am providing below a list of files and directories that contain third-party content, along with some of my guesses about it. Some have been included as is, some have been modified. The list is not necessarily complete, see issue #5878, for example. Please add any suspicious content to the bottom of the list. For each file or group of files, we need to be in a position to provide this information:

1. The name and the exact version of the project the files were copied from
2. A short description of that project
3. Cryptography? (probably "no")
4. The source of the content (the home page and/or the main source code repository, etc)
5. The short name(s) of the license(s)
6. Was it modified from the original? (If "yes", review by the IP team will take more time)

As a first step, I would like to have people who placed any given third-party content in this repository also write the required information below in comments. In time, any content that can't be accounted for will need to be either investigated or removed before we can move this repository to the Eclipse Foundation. Further, any third-party test resources must also be moved out into dl4j-test-resources.

**Note**: We will be able to keep the current content as part of archived branches.

Apart from those, there are still over 100 files missing copyright headers. We need them in **all** files that can accommodate comments. @sshepel?

For more information about the guidelines we need to follow as an Eclipse project:
 * https://www.eclipse.org/legal/committerguidelines.php
 * https://wiki.eclipse.org/Development_Resources/IP/Test_and_Build_Dependencies
 * https://www.eclipse.org/org/documents/Eclipse_Policy_and_Procedure_for_3rd_Party_Dependencies_Final.pdf
 * https://www.eclipse.org/projects/handbook/#ip-copyright-headers

<span data-href="#aha_task_list_start" data-version="v2" data-aha="Do not remove this tag. Used for Aha! syncing."/>

- [ ] modified nd4j/nd4j-serde/nd4j-aeron/src/main/java/org/nd4j/aeron/ipc/AeronUtil.java (Copyright 2014 - 2016 Real Logic Ltd.)

- [x] modified nd4j/nd4j-common/src/main/java/org/nd4j/util/StringUtils.java (Copyright Apache Hadoop)

- [x] modified nd4j/nd4j-common/src/main/java/org/nd4j/util/FingerPrintKeyer.java (Copyright 2010, Google Inc.)

[...]

PR #6977, opened 2019-01-11
[WIP] Update SerializableCSVParser copyright header

AlexDBlack:
See: https://github.com/deeplearning4j/deeplearning4j/issues/6907#issuecomment-452558131


saudet:
Ah, sorry, was just getting to this. 
Looks good, but we'll need to add blurbs like this in the main LICENSE file: 
https://github.com/apache/incubator-mxnet/blob/master/LICENSE#L204

PR #6990, opened 2019-01-13
Builder properties

rnett:

Property getters and setters were added to layer builders.  Primarily for Kotlin integration.

These properties can usually be accessed in the layer classes, but didn't have accessors in the Builder class.  In the interest of maintaining a single "build pipeline" and keeping things consistent, I think it makes more sense to add them here, rather than use the layer class constructors in Kotlin.

Mostly done via Lombok annotations (I used `@Getter`/`@Setter`, would `@Data` be better?).  Some were implemented to respect array size/shape constraints.

## How was this patch tested?

No functionality was changed, and it compiles.  After looking at what other tests there are, I didn't create any for the properties.  I can if requested.

## Quick checklist

- [x] Created tests for any significant new code additions (N/A). 
- [x] Relevant tests for your changes are passing (N/A).

### DIFF
--- a/deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/samediff/testlayers/SameDiffConv.java
+++ b/deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/samediff/testlayers/SameDiffConv.java
@@ -16,9 +16,7 @@
 
 package org.deeplearning4j.nn.layers.samediff.testlayers;
 
-import lombok.Data;
-import lombok.EqualsAndHashCode;
-import lombok.val;
+import lombok.*;
 import org.deeplearning4j.nn.conf.ConvolutionMode;
 import org.deeplearning4j.nn.conf.InputPreProcessor;
 import org.deeplearning4j.nn.conf.NeuralNetConfiguration;
@@ -31,6 +29,7 @@ import org.deeplearning4j.nn.params.ConvolutionParamInitializer;
 import org.deeplearning4j.nn.weights.WeightInitUtil;
 import org.nd4j.autodiff.samediff.SDVariable;
 import org.nd4j.autodiff.samediff.SameDiff;
+import org.nd4j.base.Preconditions;
 import org.nd4j.linalg.activations.Activation;
 import org.nd4j.linalg.api.memory.MemoryWorkspace;
 import org.nd4j.linalg.api.ndarray.INDArray;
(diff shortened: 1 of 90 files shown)


deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/samediff/testlayers/SameDiffConv.java
```diff
@@ -161,14 +160,51 @@ public void applyGlobalConfigToLayer(NeuralNetConfiguration.Builder globalConfig
 
     public static class Builder extends SameDiffLayer.Builder<Builder> {
```
AlexDBlack: There's no real need to modify these "test only" layers, as they aren't part of the public API... though now that it's done maybe there's no harm.

deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/samediff/testlayers/SameDiffConv.java
```diff
@@ -161,14 +160,51 @@ public void applyGlobalConfigToLayer(NeuralNetConfiguration.Builder globalConfig
...
 
+        @Getter
+        @Setter
```
AlexDBlack: This applies everywhere in this PR: I think we should be able to use ```@Getter``` and ```@Setter``` on the builder classes, rather than having these two annotations on every single field? That would be a lot cleaner, with the same result I think.

deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/samediff/testlayers/SameDiffConv.java
```diff
@@ -161,14 +160,51 @@ public void applyGlobalConfigToLayer(NeuralNetConfiguration.Builder globalConfig
...
+
+        @Getter
+        @Setter
         private Activation activation = Activation.TANH;
+
+        @Getter
         private int[] kernel = new int[]{2, 2};
+
+        public void setKernel(int[] kernel) {
+            Preconditions.checkArgument(kernel.length == 2, "Must have 2 kernel values - got %s", kernel);
```
AlexDBlack: Also applies everywhere: for input validation - if we're adding it, let's have it in the one place for both. So, something like:
```
public void setKernel(int[] kernel) {
    kernel(kernel)
}
		
public Builder kernel(int... kernel){
    Preconditions.checkArgument(kernel.length == 2, "Must have 2 kernel values - got %s", kernel);
    this.kernel = kernel;
    return this;
}
```

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/AbstractLSTM.java
```diff
@@ -1,15 +1,12 @@
 /*******************************************************************************
  * Copyright (c) 2015-2018 Skymind, Inc.
  *
- * This program and the accompanying materials are made available under the
- * terms of the Apache License, Version 2.0 which is available at
- * https://www.apache.org/licenses/LICENSE-2.0.
+ * This program and the accompanying materials are made available under the terms of the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
```
AlexDBlack: These licence headers should not be modified at all.
I realize the content is the same, but there's no reason to change it in just a couple of classes. It should stay consistent with the content of all other files.

deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/nn/layers/samediff/testlayers/SameDiffConv.java
rnett: In some cases, the validation is only done in the constructor (e.g. ZeroPadding1DLayer).  I feel like moving the validation in those cases is a bit beyond the scope of this PR.

In the cases where there is a method to call, I'll use it.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/AbstractLSTM.java
saudet: FYI, that's probably the formatter plugin doing this. Until we fix this, we should probably not recommend running it...

PR #6997, opened 2019-01-14
keras autogen docs license

maxpumperla:
As per discussion in https://github.com/deeplearning4j/deeplearning4j/issues/6907


saudet:
Could you add the copyright as well? If we don't know which copyright applies, I guess we'll have to add all of them.

BTW, the format looks good for now, but we'll probably be adding other components that are under the MIT license in here. The LICENSE file from MXNet offers another way we could go about this:
https://github.com/apache/incubator-mxnet/blob/master/LICENSE#L204
No need to worry about that for now though, jut add the copyright(s), thanks :)

saudet:
Thanks! @AlexDBlack Please do something like that, but we don't need to readd the whole license in your case, just the copyright and something to refer to the license..

PR #7015, opened 2019-01-16
[WIP] Update LICENSE

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/6997#issuecomment-454609889
https://github.com/deeplearning4j/deeplearning4j/pull/6977#issuecomment-454234998

@saudet I wasn't clear if we needed to link to the actual file here?
(i.e., https://github.com/deeplearning4j/deeplearning4j/blob/master/datavec/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/SerializableCSVParser.java )
Also wondering if we shold organize this by licence type, like this: https://github.com/apache/incubator-mxnet/blob/master/LICENSE#L212-L214


saudet:
Yeah, we'll need to organize this somehow eventually, but for now just tacking them along should be OK.

AlexDBlack:
@saudet done - good to go now?

PR #7048, opened 2019-01-21
Fixed deadlock in VocabConstructor

alexanderst:

Fixed reading and writing for conditional variable denoting thread completion.

## How was this patch tested?

Unit test.

## Quick checklist


jmiserez:
@alexanderst A comment on the changes in this MR: 

The line saying "as we see in profiler, this lock isn't really happen too often" seems to be platform and/or JVM-dependent. At least on my machine (Win 7, jdk8u191 (64bit), DL4J 1.0.0beta3) this happens every time, it always waits the full 1 second. There does not seem to be an unpark() call from the other thread and also otherwise no synchronization (e.g. using wait/notify instead of park() or a java.util.concurrent.locks.Condition) that would wake up the main thread. The call to AtomicBoolean.set() will not actually wake it up, so on *some* platforms it may actually always sleep the full amount (such as on my machine).

I'll test once the new release is out and file a real issue then, but from the looks of it there are still synchronization issues even after this change.

raver119:
There were significant changes, and now there's consumer/producer setup with guaranteed ordering. Old mechanics was just removed, as well as other possibilities to lock out.

Now parallelism is localized to c++ level only.

However, any feedback is welcome, and you can try current snapshots.

raver119:
@jmiserez ^^^

PR #7151, opened 2019-02-12
[WIP] DL4J MKL-DNN Support

AlexDBlack:
Adds MKL-DNN support for DL4J layers.

MKL-DNN support for:
* Subsampling layer. Implemented and passing tests.
* ConvolutionLayer: Implemented and passing tests.
* Batch norm: forward pass only implemented (passing tests). No backprop on account of this: https://github.com/deeplearning4j/deeplearning4j/issues/7166

Note that conv3d is already using conv3dnew op, hence already supports MKL-DNN by default.

Previous performance issues with MKL-DNN have been resolved.

LRN is implemented but is disabled on account of this: https://github.com/deeplearning4j/deeplearning4j/issues/7272

LSTM is not implemneted here - we don't yet have MKL-DNN support at libnd4j blockLSTM op (or similar) level.

### DIFF
--- a/deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/LayerHelperValidationUtil.java
+++ b/deeplearning4j/deeplearning4j-core/src/test/java/org/deeplearning4j/LayerHelperValidationUtil.java
@@ -1,0 +1,315 @@
+/*******************************************************************************
+ * Copyright (c) 2015-2019 Skymind, Inc.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License, Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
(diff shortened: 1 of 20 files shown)


saudet:
If I'm not mistaken, that's not limited to MKL-DNN, but whatever gets used by ND4J, right?

AlexDBlack:
@saudet so far, yes - the plan is to check if mkl-dnn is actually available and will be used by libnd4j when executing these ops. I should also check the environment to ensure mkl-dnn isn't explicitly disabled too...

saudet:
I mean, your code looks generic, so it would pick up, for example, cuDNN when running for GPU, assuming it's implemented in libnd4j, right?

AlexDBlack:
> so it would pick up, for example, cuDNN when running for GPU

TBH I didn't actually think about that, but yes - if cuDNN is implemented in the same manner as mkl-dnn, then it should do so.

PR #7163, opened 2019-02-13
[WIP] Size of empty INDArray fixed

alexanderst:

Empty array will have 0 size now. 
JvmShapeInfo becomes immutable.

## How was this patch tested?

Unit testing.

## Quick checklist


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/layers/convolution/Conv1D.java
```diff
@@ -59,6 +59,7 @@
 public class Conv1D extends DynamicCustomOp {
 
     protected Conv1DConfig config;
+    private static final String INVALID_CONFIGURATION = "Invalid Conv1D configuration : s = %d p = %d ";
```
AlexDBlack: Minor issue: Should always be ```%s``` - Preconditions doesn't need format specified for different types.
Everything eventually ends up using Object.toString() anyway for these.
Only valid format specifiers other than ```%s``` are these: https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/util/NDArrayPreconditionsFormat.java#L24

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/layers/convolution/Conv2D.java
```diff
@@ -54,6 +54,7 @@
 public class Conv2D extends DynamicCustomOp {
 
     protected Conv2DConfig config;
+    private static final String INVALID_CONFIGURATION = "Invalid Conv2D configuration : sW = %d pH = %d dW = %d ";
```
AlexDBlack: As per previous %s comment

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/layers/convolution/Conv3D.java
```diff
@@ -49,6 +49,7 @@
 public class Conv3D extends DynamicCustomOp {
 
     protected Conv3DConfig config;
+    private static final String INVALID_CONFIGURATION = "Invalid Conv3D configuration : sW = %d pH = %d dW = %d ";
```
AlexDBlack: %s

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/embeddings/loader/WordVectorSerializer.java
```diff
@@ -2672,11 +2672,20 @@ protected static TokenizerFactory getTokenizerFactory(VectorsConfiguration confi
      * @param file File should point to previously saved w2v model
      * @return
      */
+    public static WordVectors loadStaticModel(InputStream inputStream) throws IOException {
+
+        File tmpFile = DL4JFileUtils.createTempFile("word2vec"+System.currentTimeMillis(), "tmp");
+        tmpFile.deleteOnExit();
+        FileUtils.copyInputStreamToFile(inputStream, tmpFile);
+        return loadStaticModel(tmpFile);
```
AlexDBlack: Better IMO to use:
```
try{
    return loadStaticModel(tmpFile);
} finally {
   tmpFile.delete();
}
```
This is better for long-lived JVMs (and those that don't exit properly).

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseNDArrayFactory.java
```diff
@@ -197,7 +197,32 @@ public INDArray linspace(int lower, int upper, int num, DataType dtype) {
         for (int i = 0; i < num; i++) {
             double t = (double) i / (num - 1);
             data[i] = lower * (1 - t) + t * upper;
+        }
+
+        //edge case for scalars
+        INDArray ret = Nd4j.createUninitialized(dtype, new long[]{data.length});
+        if (ret.isScalar())
+            return ret;
```
AlexDBlack: This looks wrong to me... No data is put for scalars - array content (single value) remains uninitialized.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseNDArrayFactory.java
```diff
@@ -197,7 +197,32 @@ public INDArray linspace(int lower, int upper, int num, DataType dtype) {
         for (int i = 0; i < num; i++) {
             double t = (double) i / (num - 1);
             data[i] = lower * (1 - t) + t * upper;
+        }
```
AlexDBlack: Can we replace these BaseNDArrayFactory methods with a native op call?
Doesn't really make sense to populate in java IMO, especially for CUDA.

raver119: It shouldn't be used anywhere now i think? I'd expect ops to be used for linspace everywhere from now on

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/layers/convolution/Conv1D.java
alexanderst: There are primitive data types in context of this message, no Objects. Should I wrap long to Long?

AlexDBlack: No, no need to box integers to Number classes.
That happens automatically - but only if an exception is to be thrown.
https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-common/src/main/java/org/nd4j/base/Preconditions.java#L73
```
public static void throwEx(String message, Object... args) {
```

The idea with all of the (seeminly redundant) primitive overloads in Preconditions class is to avoid creating objects unless an exception is thrown as much as possible, to avoid unnecessary garbage generation.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseNDArrayFactory.java
AlexDBlack: Hm... maybe we remove linspace methods from BaseNDArrayFactory interface entirely then?

alexanderst: That's what I'm doing, no need for Java linspace implementation anymore.

alexanderst: Fixes for all issues were pushed.

PR #7185, opened 2019-02-18
DL4J UI: Add multi-session support

### LINKED ISSUE #7110: DL4J UI: Add multi-session support
Following on from this discussion: https://github.com/deeplearning4j/deeplearning4j/issues/7103#issuecomment-460228925
@printomi

> @AlexDBlack Yes, I want to work on this, and I would be happy to contribute to DL4J with this. My primary goal is not only training multiple models attached to the same UI, but also loading historical training results with given session ID.
> I would do this by:

> 1. Modifying TrainModule so that it holds multiple StatsStorage instances mapped to session IDs. (Session ID can be set with StatsListener.setSessionID() before training, that is how I give unique IDs to my training statistics.)

This is already supported: 
https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/module/train/TrainModule.java#L89

i.e., all attached StatsStorage instances are present in that map. Key is the session ID, value is the StatsStorage instance that contains data for that session ID. Each StatsStorage instance can contain data for more than one session.

> 2. Adding PlayUIServer a custom StatsStorage loader. This would handle session IDs not already loaded.

Currently, the way to make data available to the UI is to attach it (UIServer.getInstance().attach()). Is the idea here to allow loading of stats storage directly from the UI?
i.e., point it to a file, and do the attach internally? If so, sounds good to me :+1:
We'd need some way to detach though. Longer-lived servers may require some sort of auto detach (based on time/usage) to avoid memory leaks (i.e., if user forgets to detach their session).

> 3. Making an option for hiding session selector dropdown from Java code, to separate users.

I think that would require more substantial changes on the backend. It's singleton/single user design pretty much everywhere.
If we can work out a way to do this, I'd be happy to add it, as long as the current single user type design is retained (as that's what most users need - don't want to make that less user friendly).

> 4. If these work properly, I would insert user authentication somehow, to get a truly multi-user UI server.

This should probably be a separate PR, maybe even with a (brief) written proposal posted first to make sure it's OK before you go through the effort of implementing it.
I'm not opposed to having it in principle, but would want to think carefully about the design first if we're going to be responsible for maintaining it :)

### LINKED ISSUE #7274: UIServer.stop() missing class org/slf4j/bridge/SLF4JBridgeHandler
#### Issue Description

I tried to run UIServer.stop() in one of the new tests in `TestPlayUIMultiSession` in PR #7185, but it threw this exception:
```
java.lang.NoClassDefFoundError: org/slf4j/bridge/SLF4JBridgeHandler
	at play.api.Logger$.shutdown(Logger.scala:306)
	at play.core.server.Server$class.stop(Server.scala:69)
	at play.core.server.NettyServer.stop(NettyServer.scala:175)
	at play.server.Server.stop(Server.java:28)
	at org.deeplearning4j.ui.play.PlayUIServer.stop(PlayUIServer.java:422)
```
See [this gist](https://gist.github.com/printomi/c6999cfe57426d77630fb60b00ecf803) for the full exception stack trace.

This behavior can be also reproduced in [this example](https://github.com/deeplearning4j/dl4j-examples/blob/master/dl4j-examples/src/main/java/org/deeplearning4j/examples/userInterface/UIExample.java).

Do we miss a dependency in [pom.xml](https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/pom.xml)?

#### Version Information

* Deeplearning4j version: 1.0.0-SNAPSHOT, 1.0.0-beta3
* platform information (OS, etc): Windows 10, Java 8

#### Contributing

I guess, that this is a dependency issue. What do you think?

printomi:

This closes #7110
Add optional multi-session functionality to UIServer: separate sessions by including session ID in URL-s, like /train/:sessionId/overview instead of /train/overview in single-session UI-server.
- Add --multiSession command-line option to PlayUIServer. Pass this setting is to TrainModule and DefaultModule. The front-end javascript code queries this setting to choose working mode.
- enabling multi-session mode: change PlayUIServer constructor and add command-line option (--multiSession), add UIServer factory method
- logging attached and detached sessions with URL for convenience in multi-session mode
- JavaScript: in multi-session mode, the session ID is determined from the URL. A new function getSessionSettings(callback) defined in train.js is used to get this information, then query URLs for single/multi-session mode are determined in the callback.
- Set separate i18n for sessions in multi-session mode.
- add option for attaching StatsStorage based on session ID passed in the URL, via custom provider, when in multi-session mode

This closes #7274 
Fix missing dependency for UIServer.stop(), handle stopped state, add test.

## How was this patch tested?

New manual tests in `TestPlayUIMultiSession` to check multi-session support, auto-attaching functionality and an auto-attaching-detaching example.
New test in `TestPlayUI` for stopping functionality.

## Quick checklist


printomi:
I can imagine a unit test for the auto-attach functionality (`PlayUIServer.autoAttachStatsStorageBySessionId(Function<String, StatsStorage> statsStorageProvider)`). With this, I could even create a stress-test, to see, whether we should use more synchronization in `TrainModule`. Shall I try it?

printomi:
With the improved tests, I stressed the UI-server with hundreds of parallel training sessions, without any error. I think, that the current synchronization works well enough, so I don't think that method-level synchronization is needed in `TrainModule`.

printomi:
I have reviewed the changes, and made some refactoring and fixes. @AlexDBlack what do you think about this PR?

printomi:
For the sake of completeness, I suggest, multi-session mode should be enabled in `ConvolutionalListenerModule`, too. In multi-session mode, one would access this page at `/activations/:sessionId`, instead of `/activations`. Shall I do it?

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/api/UIServer.java
```diff
@@ -38,9 +39,27 @@
...
+        if (uiServer== null) {
+            PlayUIServer playUIServer = new PlayUIServer(PlayUIServer.DEFAULT_UI_PORT, multiSession);
+            playUIServer.setMultiSession(multiSession);
+            if (statsStorageProvider != null) {
+                playUIServer.autoAttachStatsStorageBySessionId(statsStorageProvider);
+            }
+            playUIServer.runMain(new String[] {});
             uiServer = playUIServer;
         }
         return uiServer;
```
AlexDBlack: Maybe it's unnecessary, but one way I can think of breaking this is doing the following:
```
UIServer.getInstance();
//some time later...
UIServer.getInstance(true, null);    //User expects multi-session, but it's already initialized as single session mode.
```
Perhaps validating the already-initialized UIServer is actually multi-session would be a good idea?

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/module/defaultModule/DefaultModule.java
```diff
@@ -23,19 +23,38 @@
...
 
+import static play.mvc.Results.ok;
 import static play.mvc.Results.redirect;
 
 /**
  * Landing page - i.e., "/" route
  * @author Alex Black
  */
 public class DefaultModule implements UIModule {
+    private boolean multiSession;
```
AlexDBlack: Minor point: we can probably make this final...
```private final boolean multiSession```

File: deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java [hunk 3]
```diff
@@ -102,14 +104,39 @@
     @Parameter(names = {"-f", "--customStatsFile"}, description = "Path to create custom stats file (remote only)", arity = 1)
     private String customStatsFile;
 
+    @Parameter(names = {"-m", "--multiSession"}, description = "Whether to enable multiple separate browser sessions or not", arity = 1)
+    private boolean multiSession;
+
+    StatsStorageLoader statsStorageLoader;
```
AlexDBlack: Intentionally package private? (If so: add comment to that effect). Otherwise private/protected.

AlexDBlack:
> For the sake of completeness, I suggest, multi-session mode should be enabled in `ConvolutionalListenerModule`, too. In multi-session mode, one would access this page at `/activations/:sessionId`, instead of `/activations`. Shall I do it?

ConvolutionalListener needs some other fixes, as you already know: https://github.com/deeplearning4j/deeplearning4j/issues/7217
Perhaps we tackle it separately, and fix all 3 at once? (I was looking at trying to fix that some time this week, though it isn't a very high priority item for me...)

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/api/UIServer.java
printomi: Yes, i this case, user would get a single-session instance. I would add user-side validation, by introducing `public abstract boolean isMultiSession` in `UIServer`, and a warning message in `getInstance(true, null)`, if the instance is not multi-session.

printomi: Instead of a warning message, I would throw a `RuntimeException`. This way, we only return an instance, if the call to the singleton factory method is consistent with the first call. Is `RuntimeException` the right exception to throw in this case?

printomi:
I agree, let's tackle ConvolutionalListener issues separately, outside of this PR.

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/module/defaultModule/DefaultModule.java
printomi: OK, made `DefaultModule.multiSession` final.

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/play/PlayUIServer.java
printomi: It was by mistake, I made it private now.

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/api/UIServer.java
AlexDBlack: Yeah, I think I'd prefer an exception in this case... it's clear, unambiguous and won't be missed my the user. The exception text should tell user what's going on, why, and what they can do to fix it (I assume just stopping the UI server instance, and starting a new one should suffice).
An isMultiSession method in addition (and used in exception) sounds fine too.

printomi: Well, I have coded the validation with an exception. I have made two test cases to demonstrate how it works. I will add your solution suggestion 
> stopping the UI server instance, and starting a new one

to the exception text. 

On the other hand, I have an issue with stopping the server: I tried to run UIServer.stop() in one of the new tests in `TestPlayUIMultiSession`, but it threw this exception:
```
java.lang.NoClassDefFoundError: org/slf4j/bridge/SLF4JBridgeHandler
	at play.api.Logger$.shutdown(Logger.scala:306)
	at play.core.server.Server$class.stop(Server.scala:69)
	at play.core.server.NettyServer.stop(NettyServer.scala:175)
	at play.server.Server.stop(Server.java:28)
	at org.deeplearning4j.ui.play.PlayUIServer.stop(PlayUIServer.java:422)
```
See [this gist](https://gist.github.com/printomi/c6999cfe57426d77630fb60b00ecf803) for the full exception stack trace. 
Do I miss a dependency in pom.xml?

printomi:
I agree that listing the available sessions on the main page would improve usability in multi-session mode. 
I am ready with the list, you can try it. 

Also, I finished the single/multi session check + exception.

printomi:
I have just made a rebase to reduce commit count, after the last commit 
439cc9f

printomi:
One the question remained for me: how can I stop UIServer without getting the above-mentioned exception?

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-play/src/main/java/org/deeplearning4j/ui/api/UIServer.java
printomi: @AlexDBlack I have reported the above mentioned issue with stopping UIServer #7274 . If there is an easy solution, that could be implemented in this PR. What do you think?

printomi: I think, that commits https://github.com/deeplearning4j/deeplearning4j/pull/7185/commits/67afb807c6453b27163e91628d82beee969e15cf and https://github.com/deeplearning4j/deeplearning4j/pull/7185/commits/dad7bb9da0c65be1befe0affd1968f1b66137549 solve the issue. @AlexDBlack please, resolve this conversation, if you accept the solution.

printomi:
@AlexDBlack I have rebased onto master, and regenerated scala templates.

printomi:
Thanks a lot, @AlexDBlack . I can now start to build upon this, using snapshots.

printomi:
> > For the sake of completeness, I suggest, multi-session mode should be enabled in `ConvolutionalListenerModule`, too. In multi-session mode, one would access this page at `/activations/:sessionId`, instead of `/activations`. Shall I do it?
> 
> ConvolutionalListener needs some other fixes, as you already know: #7217
> Perhaps we tackle it separately, and fix all 3 at once? (I was looking at trying to fix that some time this week, though it isn't a very high priority item for me...)

Now that issue #7217 has been closed, and I can confirm that [TestConvolutionalListener.testUI()](https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-ui/src/test/java/org/deeplearning4j/ui/weights/TestConvolutionalListener.java#L45) passes, I would ask again, if multi-session support for `ConvolutionalListenerModule` is welcome in DL4J-UI?
If so, I would file an issue about this, and I might have time to implement it in the next few weeks. 
(It is not a high priority for me, but for the sake of completeness, I think it would worth.)

AlexDBlack:
@printomi Sure, if you want to, go ahead and do that. I don't think it's something we would want to allocate internal resources to though (other than reviewing PRs etc). Feel free to open an issue about it too if you want.

PR #7188, opened 2019-02-18
datavec-python

farizrahman4u:


datavec/datavec-python/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+    </parent>
+    <packaging>jar</packaging>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>datavec-python</artifactId>
+
+
+    <dependencies>
+
+     <dependency>
```
agibsonccc: Mind adding dependency versions under dependency management in the top level pom? We do this to avoid dependency conflicts across modules.
Do this with 
<dependency>
            <groupId>org.bytedeco.javacpp-presets</groupId>
            <artifactId>cpython-platform</artifactId>
            <version>3.6-1.4.4-SNAPSHOT</version>
        </dependency>
        <dependency> 


converting to:
<dependency>
            <groupId>org.bytedeco.javacpp-presets</groupId>
            <artifactId>cpython-platform</artifactId>
        </dependency>
        <dependency>\

putting the previous declaration under a dependency management section.

Also, please version the python and javacpp versions with variables at the top level.

datavec/datavec-python/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+            <version>1.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.bytedeco.javacpp-presets</groupId>
+            <artifactId>cpython-platform</artifactId>
+            <version>3.6-1.4.4-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.nd4j</groupId>
+            <artifactId>nd4j-native</artifactId>
```
agibsonccc: Replace with ${project.version}?

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public long[] getStrides() {
+        return strides;
+    }
+
+    public DataType getDType() {
+        return dtype;
+    }
+
+    public JSONObject toJSON(){
```
agibsonccc: Instead of simpleJson, can  you just use nd4j's json representation?

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
```diff
@@ -0,0 +1,179 @@
...
+            shape[i] = (Long)shapeJson.get(i);
+        }
+        JSONArray stridesJson = (JSONArray)json.get("strides");
+        strides = new long[stridesJson.size()];
+        for (int i=0; i<strides.length; i++){
+            strides[i] = (Long)stridesJson.get(i);
+        }
+        String dtypeStr = (String)json.get("dtype");
+        if (dtypeStr != null){
+            if (dtypeStr.equals("FLOAT")){
```
agibsonccc: Why not just use Dtype.fromString(..)?

farizrahman4u: `DataType.fromString()` does not exist. Want me to add it to Nd4j?

saudet: We're supposed to have string support now, but not sure it's all fully there yet, @raver119?

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
```diff
@@ -0,0 +1,179 @@
...
+        String dtypeStr = dtype.toString();
+        jsonObject.put("data", data);
+        jsonObject.put("shape", shape);
+        jsonObject.put("dtype", dtypeStr);
+        return jsonObject;
+    }
+
+
+    public NumpyArray(JSONObject json) throws Exception{
+        address = (Long)json.get("address");
```
saudet: This looks inefficient, we should be able to call Python functions directly to get these values without serializing everything to JSON, but I'm not sure how much of a slowdown this is in practice.

datavec/datavec-python/src/main/java/org/datavec/python/PythonExecutioner.java
```diff
@@ -0,0 +1,613 @@
...
+        }
+        else{
+            log.info("CPython: Py_NewInterpreter()");
+            PyThreadState threadState = Py_NewInterpreter();
+            interpreters.put(name, threadState);
+            log.info("CPython: PyThreadState_Swap()");
+            PyThreadState_Swap(threadState);
+            init();
+        }
+        currentInterpreter = name;
```
saudet: Does using multiple Python interpreters actually help? IIRC, the GIL applies to the process as a whole, not for each interpreter, or did that change?

pom.xml
```diff
@@ -301,6 +301,8 @@
         <javacpp.version>1.4.4</javacpp.version>
         <javacpp-presets.version>1.4.4</javacpp-presets.version>
         <javacv.version>1.4.4</javacv.version>
+
+        <cpython-platform.version>3.6-1.4.4-SNAPSHOT</cpython-platform.version>
```
saudet: 1.4.4 has been released, so we can remove the SNAPSHOT bit! And actually just use `${javacpp-presets.version}` to have it take the same version as all the other presets.

datavec/datavec-python/src/main/java/org/datavec/python/PythonExecutioner.java
farizrahman4u: But we can have multiple "contexts" in memory. There are tests for this. (See TestSamdbox.java)

pom.xml
```diff
@@ -292,7 +292,7 @@
         <javacpp-presets.version>1.4.4</javacpp-presets.version>
         <javacv.version>1.4.4</javacv.version>
 
-        <cpython-platform.version>3.6-1.4.4-SNAPSHOT</cpython-platform.version>
+        <cpython-platform.version>3.6-${javacpp-presets.version}</cpython-platform.version>
```
agibsonccc: Vary the python version too.

datavec/datavec-python/src/main/java/org/datavec/python/PythonExecutioner.java
```diff
@@ -0,0 +1,613 @@
...
+        }
+        catch (Exception e){
+            log.error(e.toString());
+        }
+
+    }
+    private String outputCode(PythonVariables pyOutputs){
+        return "";
+        /*
+        if (pyOutputs == null){
```
maxpumperla: what happened here?

datavec/datavec-python/src/main/java/org/datavec/python/PythonTransform.java
```diff
@@ -0,0 +1,171 @@
...
+
+public class PythonTransform {
+    private String setupCode;
+    private String execCode;
+    private String code;
+    private String name;
+    private PythonVariables pyInputs;
+    private PythonVariables pyOutputs;
+
+    private void parseSetupAndExecCode() throws Exception{
```
maxpumperla: this is a confusing name (although I see how you got it). It suggests that you parse the setup and exec the code.

A doc string might help. Maybe find an unambiguous convention.

Also, what's the setup tag for? Can we have a test for it?

farizrahman4u: The "setup" part of the code is whatever you need to do only once (imports, loading stuff to memory). We do it once and persist that state in memory so that we dont have to do it per transaction. I havent committed the tests because of the issue i discussed in our call (regarding arrays getting GCed too early sometimes). Will add tests once it works.

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
farizrahman4u: will fix after string support added to nd4j

pom.xml
farizrahman4u: done

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
```diff
@@ -0,0 +1,114 @@
...
+        this.strides = strides;
+        this.dtype = dtype;
+        setND4JArray();
+        if (copy){
+            nd4jArray = nd4jArray.dup();
+            this.address = nd4jArray.data().address();
+        }
+    }
+
+    public long getAddress() {
```
agibsonccc: lombokify this maybe? We can probably get rid of the getters and setters with the annotation.

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
```diff
@@ -0,0 +1,114 @@
...
+        return nd4jArray;
+    }
+
+    private void setND4JArray(){
+        long size = 1;
+        for(long d: shape){
+            size *= d;
+        }
+        Pointer ptr = nativeOps.pointerForAddress(address);
+        ptr = ptr.limit(size);
```
agibsonccc: @saudet mind checking this part?

datavec/datavec-python/src/main/java/org/datavec/python/PythonCondition.java
```diff
@@ -0,0 +1,163 @@
...
+ * and returns a boolean indicating whether or not to filter a row.
+ * The values of all columns in current row is available as gobal variables to f.
+ */
+public class PythonCondition implements Condition {
+
+    private Schema inputSchema;
+    private PythonVariables pyInputs;
+    private PythonTransform pythonTransform;
+    private String code;
+    public PythonCondition(String pythonCode){
```
agibsonccc: Run the intellij indenter please.

datavec/datavec-python/src/main/java/org/datavec/python/PythonCondition.java
```diff
@@ -0,0 +1,163 @@
...
+
+    public boolean condition(List<Writable> list){
+        PythonVariables inputs = getPyInputsFromWritables(list);
+        try{
+            PythonVariables output = PythonExecutioner.getInstance().exec(pythonTransform, inputs);
+            boolean ret = output.getIntValue("out") != 0;
+            return ret;
+        }
+        catch (Exception e){
+            System.out.println(e);
```
agibsonccc: Slf4j logging please.

datavec/datavec-python/src/main/java/org/datavec/python/PythonTransform.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Override
+    public List<Writable> map(List<Writable> writables){
+        PythonVariables pyInputs = getPyInputsFromWritables(writables);
+        try{
+            PythonVariables pyOutputs = PythonExecutioner.getInstance().exec(this, pyInputs);
+            return getWritablesFromPyOutputs(pyOutputs);
+        }
+        catch (Exception e){
+            throw new RuntimeException(e);
```
agibsonccc: @AlexDBlack is this normal for transforms? Have we changed anything in the data vec transforms you can think of?

datavec/datavec-python/src/main/java/org/datavec/python/PythonTransform.java
```diff
@@ -0,0 +1,261 @@
...
+                case INT:
+                    if (w instanceof LongWritable){
+                        ret.addInt(name, ((LongWritable)w).get());
+                    }
+                    else{
+                        ret.addInt(name, ((IntWritable)w).get());
+                    }
+
+                    break;
+                case FLOAT:
```
agibsonccc: There's FloatWritable and DoubleWritable. Make sure to cover both cases.

datavec/datavec-python/src/main/java/org/datavec/python/PythonTransform.java
```diff
@@ -0,0 +1,261 @@
...
+                case NDARRAY:
+                    out.add((Writable) new NDArrayWritable(pyOuts.getNDArrayValue(name).getND4JArray()));
+                    break;
+            }
+        }
+        return out;
+    }
+
+    private void parseSetupAndExecCode() throws Exception{
+        String startTag = "#<SETUP>";
```
agibsonccc: Mind commenting this? It's not intuitive what we're doing here.

datavec/datavec-python/src/main/java/org/datavec/python/PythonVariables.java
```diff
@@ -0,0 +1,316 @@
...
+                break;
+            case LIST:
+                addList(name);
+                break;
+            case FILE:
+                addFile(name);
+        }
+    }
+
+    public void add (String name, Type type, Object value) throws Exception{
```
agibsonccc: Javadoc.

datavec/datavec-python/src/main/java/org/datavec/python/RestrictedPython.java
```diff
@@ -0,0 +1,35 @@
+package org.datavec.python;
+
+import org.nd4j.linalg.io.ClassPathResource;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class RestrictedPython {
+
```
agibsonccc: Javadoc.

datavec/datavec-spark/pom.xml
```diff
@@ -301,6 +301,12 @@
             <version>${logback.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.datavec</groupId>
+            <artifactId>datavec-python</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
agibsonccc: ${project.version} please

datavec/datavec-python/src/test/java/org/datavec/python/TestPythonTransformProcess.java
```diff
@@ -0,0 +1,106 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+
+public class TestPythonTransformProcess {
+
+    @Test
+    public void testStringConcat() throws Exception{
+        Schema.Builder schemaBuilder = new Schema.Builder();
+        schemaBuilder
```
agibsonccc: More tests around float/double/long/int and mixed types would be good.

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
saudet: Now that you mention it, the size needs to take into account the element size somehow because this is a raw buffer, although this might be reset somewhere in Nd4j.createBuffer() already, but it looks like that's missing for the new types.

datavec/datavec-python/src/test/java/org/datavec/python/TestSerde.java
```diff
@@ -0,0 +1,38 @@
...
+import org.datavec.api.transform.schema.Schema;
+import org.datavec.api.transform.serde.JsonSerializer;
+import org.datavec.api.transform.serde.YamlSerializer;
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+
+public class TestSerde {
+
+    public static YamlSerializer y = new YamlSerializer();
+    public static JsonSerializer j = new JsonSerializer();
```
agibsonccc: Can you add local transform executioner as well? (Arrow)

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
farizrahman4u: The tests pass. We have tests for float, double, int, short and long. So it works right now.

saudet: It probably works for most things, but we're going to hit some place where it won't work. Anyway, this is actually something that should be fixed in https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-buffer/src/main/java/org/nd4j/linalg/api/buffer/factory/DefaultDataBufferFactory.java#L566  where those values should be set in the `create(Pointer, ...)` method like they are in `create(DoublePointer ...)`, etc.

datavec/datavec-python/src/main/java/org/datavec/python/PythonTransform.java
```diff
@@ -0,0 +1,242 @@
...
+import org.datavec.api.transform.Transform;
+import org.datavec.api.transform.schema.Schema;
+import org.datavec.api.writable.*;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+import org.nd4j.shade.jackson.annotation.JsonIgnoreProperties;
+
+import javax.annotation.Nullable;
```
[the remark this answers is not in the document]
agibsonccc: Switch to the lombok nullable please.

farizrahman4u:
Seems we need javacpp 1.5-SNAPSHOT, so merge only after #7296.

datavec/datavec-python/src/main/java/org/datavec/python/NumpyArray.java
agibsonccc: @farizrahman4u can you resolve this if this concern is fixed?

farizrahman4u:
@agibsonccc that buffer creation routine works and we have tests for all precisions. Also we use the same thing in jumpy.

agibsonccc:
@maxpumperla are you satisfied with this?

PR #7213, opened 2019-02-21
Standardize names and packages, standardize builder param validation

rnett:

Did a couple of things here, mostly focused on standardization.

1. Moved conf layer classes to packages matching their implementation classes (some matched, some didn't).
2. Renamed layers that didn't have Layer in their name to include it (some had it, some didn't).
3. Added 2D to 2D conventional layers that were missing it.  In **some** cases these layers aren't forced to be 2D, but their builders are (e.g. `Convolution2DLayer` doesn't enforce 2D in the layer, but does in the builder).  Since the access point for most of these layers is through the Builder, where 2D is enforced, naming them this makes enough sense I went ahead and did it.
4. Moved all Builder validation logic to setters, and moved multi-dimensional array validation to a utility class (`ValidationUtils`).

## How was this patch tested?

It wasn't.  Nothing here should change the implementation, although it would be nice to test the validation some.

When trying to compile, I got `cannot find symbol` errors from various files that I didn't edit except for comments and re-names (and some not at all).  Everything resolves in the IDE and when I look at, I think this is something with the compile process.  I can give more detail if necessary.

# Note

My fork is a bit behind master, wasn't sure if it was worth trying to update.

## Quick checklist

- [?] Relevant tests for your changes are passing.
- [ ] Ran mvn formatter:format (see [formatter instructions](http://code.revelc.net/formatter-maven-plugin/examples.html#Setting_Source_Files) for targeting your specific files). (messes with the license comments)


AlexDBlack:
So this is a fairly substantial PR.
It would have been good to discuss this with us before going to the effort of implementing this.

Here's my concerns:
1. This introduces unnnecessary breaking changes for all past DL4J code/examples, by changing the names/paths.
2. If we change any part of the fully qualified class name for the configuration, JSON deserialization of saved networks will no longer function.
3. Some of the changes aren't IMO an improvement: consider for example LSTM -> LSTMLayer name change. The original name was not ambiguous, and the new (longer) name doesn't provide any improvement or benefit. An argument for consistency could be made, but consistency is far from the only consideration here; and for consistency we could have gone the other way (like remove "Layer" from layer names that have it)


With regard to JSON deserialization, it is possible to work around that with some substantial effort - and a ton of new regression tests.

However, my biggest concern here is the fact that these changes are unnecessary. Though breaking changes are sometimes necessary, they should be implemented only when the benefit outweights the cost to users. I'm not convinced that is the case here. And API changes of this magnitude need to be carefully considered and discussed, not made unilaterally.


So where does this leave us?
1. For some of the changes, we need to step back and have a discussion whether we want them at all (like "Layer in their name")
2. For any renamed or reorganized layers, we need to consider the impact on network deserialization and backward compatibility - and work out strategies to deal with that (plus what new tests we need)
3. Some of the changes (like ValidationUtils) we could pull them out into separate PRs. 

I'd be happy to have those discussions - I'd suggest opening new issue(s) for that.
Also for future reference it's better to have more smaller PRs than one monolithic PR, as we can't merge just the acceptable parts of a monolithic PR unless all parts are acceptable.

I appreciate the effort that has gone into this, but unfortunately I don't see how we can merge this PR before those issues have been discussed an resolved.

PR #7218, opened 2019-02-22
Centralized validation via setters

rnett:

Moved array and non-negative validation to `ValidationUtils`
Made setters use those validation methods.
Made all setting methods go through setters.

## How was this patch tested?

No changes to implementations.  It compiles.

## Quick checklist


deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/ValidationUtils.java
```diff
@@ -0,0 +1,332 @@
+package org.deeplearning4j.util;
+
+import com.google.common.base.Preconditions;
```
AlexDBlack: Use ND4J preconditions class instead: https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-common/src/main/java/org/nd4j/base/Preconditions.java

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/ValidationUtils.java
```diff
@@ -0,0 +1,332 @@
...
+import java.util.Arrays;
+
+public class ValidationUtils {
```
AlexDBlack: Add private no-arg constructor to avoid users being able to do ```new ValidationUtils()```

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/ValidationUtils.java
```diff
@@ -0,0 +1,332 @@
...
+     * Checks that the values is >= 0.
+     *
+     * @param data An int
+     * @param paramName The param name, for error reporting
+     */
+    public static void validateNonNegative(int data, String paramName){
+        Preconditions.checkArgument(data >= 0,
+                "Values for " +
+                        paramName + " must be >= 0, got: " +
+                        data);
```
AlexDBlack: This comment applies to all of your uses of Preconditions case: don't create strings like this.
The correct way of using preconditions class doesn't (in most cases) result in any object creation unless the precondition fails:
```
Preconditions.checkArgument(data >= 0, "Values for %s must be >= 0, got %s", paramName, data);
```

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution1DLayer.java
```diff
@@ -178,10 +179,30 @@ public Builder stride(int stride) {
...
          */
         public Builder padding(int padding) {
-            this.padding[0] = padding;
+            this.setPadding(new int[]{padding});
             return this;
         }
 
+        @Override
+        public void setKernelSize(int[] kernelSize) {
+            this.kernelSize[0] = ValidationUtils.validate1NonNegative(kernelSize, "kernelSize")[0];
```
AlexDBlack: Extra validation here is great. :+1:

Looks like this will result in a NPE if the no-arg constructor is used:
https://github.com/deeplearning4j/deeplearning4j/blob/05640512857fba29db7d29b5faeb6828ada3988d/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution1DLayer.java#L125

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/Convolution1DLayer.java
```diff
@@ -178,10 +179,30 @@ public Builder stride(int stride) {
          * @param padding Padding value
          */
         public Builder padding(int padding) {
-            this.padding[0] = padding;
+            this.setPadding(new int[]{padding});
             return this;
         }
 
+        @Override
+        public void setKernelSize(int[] kernelSize) {
```
AlexDBlack: Let's make all of these ```setX(int[] x)``` varargs methods - that allows users (and indeed the builder methods) to do things like ```setKernelSize(kernelSize);``` instead of ```setKernelSize(new int[]{kernelSize});```
That should still be an override. This comment applies to other layers too, not just this one.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/ValidationUtils.java
```diff
@@ -0,0 +1,318 @@
...
+                "Values for %s must be >= 0, got: %s", paramName, data);
+    }
+
+    /**
+     * Checks that all values are >= 0.
+     *
+     * @param data An array
+     * @param paramName The param name, for error reporting
+     */
+    public static void validateNonNegative(int[] data, String paramName){
```
AlexDBlack: Handling of nulls? You have some in the later methods, but not here...
If null is acceptable, we should return null here. If not, let's either add a lombok ```@NonNull``` or a ```Preconditions.checkNotNull(```. At present, nulls will give a non-useful NPE.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/util/ValidationUtils.java
```diff
@@ -0,0 +1,318 @@
+package org.deeplearning4j.util;
+
+import java.util.Arrays;
+import org.nd4j.base.Preconditions;
+
+public class ValidationUtils {
```
AlexDBlack: Brief class javadoc would be good.
Also feel free to add an ```@author``` tag if you want (totally optional).
Also need copyright header at the start of the file (you can copy/paste from elsewhere)

rnett:
The main reason I changed everything to setters is that if someone goes back and adds or changes validation for a parameter, they don't have to change it then.

It also makes it easier when calling those methods to just always use the setter, rather than having to look at the setter and see if there is any validation there or not for each use.

AlexDBlack:
OK, so test results: looks like we have a small nuber of test failures: introduced by this PR
image
(Unfortunately I ran out of time, didn't run the full suite on the PR yet, will finish tomorrow)

rnett:
I will finish fixing this soonish, its just end of quarter crunch time atm and I haven't had much time to look at it.

I'm not sure whats up with the embedding layer, I didn't change anything there are barely anything else in Dense or Feedforeward.

Quite a few of the errors come from 1D CNN, which may have been from passing `null` to a vararg method, or Upsampling1D, where I forgot a return.

The `TestInvalidConfiguration.testSubsamplingInvalidPadding` error comes from expecting the wrong exception (the -1 is now caught in the setter using Preconditions).

`testSubsamplingInvalidKernel2` and `testSubsamplingInvalidPadding2` are now valid.  When setting the padding to a single value that value is automatically duplicated out to 2, as was done in some other layers.  I can remove these, but wanted to check before I start deleting tests.

`test1dForward` was because one of the methods was named `setInputSize` and conflicted with my setter.

This should fix most of the errors.

PR #7240, opened 2019-02-28
Shyrma mmul

shyrma:
mmul helper has been improved


libnd4j/include/cblas.h
```diff
@@ -21,7 +21,7 @@
 #ifndef NATIVEOPERATIONS_CBLAS_H
 #define NATIVEOPERATIONS_CBLAS_H
 
-#include "config.h"
+// #include "config.h"
```
saudet: This will break the build, please don't do that.

libnd4j/include/graph/impl/GraphUtils.cpp
```diff
@@ -18,7 +18,7 @@
 // Created by GS <sgazeos@gmail.com> 3/7/2018
 //
 
-#include "config.h"
+// #include "config.h"
```
saudet: Here too, please don't comment those out.

libnd4j/include/cblas.h
raver119: Don't worry, it's temporary comment.

libnd4j/include/graph/impl/GraphUtils.cpp
raver119: Don't worry, it's temporary comment.

PR #7241, opened 2019-02-28
[WIP] Benchmarking

raver119:
***WIP; DO NOT MERGE;***

This PR adds simple bencmarking framework for @AlexDBlack


libnd4j/include/cblas.h
```diff
@@ -21,7 +21,9 @@
 #ifndef NATIVEOPERATIONS_CBLAS_H
 #define NATIVEOPERATIONS_CBLAS_H
 
+#ifndef __STANDALONE_BUILD__
 #include "config.h"
```
saudet: You're going to need BLAS, MKL-DNN, etc even in "standalone" mode.

raver119: Once we need it - we'll add it immediately.

Issue #7175, opened 2019-02-16
SameDiff and DL4J: Add Layer Normalization layer/ops


AlexDBlack:
Required for transformer models like BERT and GPT.
https://arxiv.org/abs/1607.06450

Mathematically it is fairly straightforward. We should probably implement it as a single c++ op for performance reasons then use that for both SameDiff and DL4J. We'll also want backprop op.

PR #7251, opened 2019-03-06
[WIP] Add Layer Normalization

treo:
See also https://github.com/deeplearning4j/deeplearning4j/issues/7175

This adds 2 new ops in SameDiff: Standardize and LayerNorm. 

Standardize will turn the examples in the given ndarray into zero mean, unit variance distributed values (as calculated along the given dimensions).

LayerNorm will then use Standardize and additionally apply the gain multiplication and optionally add a bias.

As the layer normalization paper (https://arxiv.org/abs/1607.06450) says that it isn't really suitable for CNNs, this will only be available directly on DenseLayer, SimpleRNN and LSTM.


treo:
After discussing it with @AlexDBlack, I've decided to skip layer normalization support for LSTMs in this PR.

Because neither CuDNN nor MKL-DNN support layer normalization for LSTMs the performance hit for using it would be large enough that no one would be using it anyway. Also it would add a lot of complexity to what is already a pretty complex piece of code at the moment.

We can revisit layer normalization support for LSTMs once we start moving the layer implementations over to SameDiff.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
```diff
@@ -97,8 +93,21 @@ public LayerConfT layerConf() {
         INDArray W = getParamWithNoise(DefaultParamInitializer.WEIGHT_KEY, true, workspaceMgr);
 
         INDArray epsilonNext = workspaceMgr.createUninitialized(ArrayType.ACTIVATION_GRAD, new long[]{W.size(0), delta.size(0)}, 'f');
+        if(hasLayerNorm()) {
+            INDArray g = getParamWithNoise(DefaultParamInitializer.GAIN_KEY, true, workspaceMgr);
```
AlexDBlack: Does it make sense to apply weight noise (dropconnect, gaussian noise etc) on the gain parameter?
Probably not IMO, especially given it's already applied to weights etc.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
```diff
@@ -302,6 +316,13 @@ protected INDArray preOutput(boolean training, LayerWorkspaceMgr workspaceMgr) {
 
         INDArray ret = workspaceMgr.createUninitialized(ArrayType.ACTIVATIONS, W.dataType(), input.size(0), W.size(1));
         input.castTo(ret.dataType()).mmuli(W, ret);     //TODO Can we avoid this cast? (It sohuld be a no op if not required, however)
+
+        INDArray preNorm = ret;
+        if(hasLayerNorm()){
+            preNorm = ret.dup();
```
AlexDBlack: Is this dup necessary?
(If it is: ````ret.dup(ret.order())``` in case input is F ordered)

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/SimpleRnn.java
```diff
@@ -114,12 +126,17 @@ public INDArray rnnActivateUsingStoredState(INDArray input, boolean training, bo
...
+            INDArray rCurrent = (hasLayerNorm() ? p.getFourth().get(all(), all(), point(i)) : null);
             INDArray inCurrent = input.get(all(), all(), point(i));
             INDArray epsOutCurrent = epsOut.get(all(), all(), point(i));
 
             if(dldzNext != null){
                 //Backprop the component of dL/da (for current time step) from the recurrent connections
                 Nd4j.gemm(dldzNext, rw, dldaCurrent, false, true, 1.0, 1.0);
+
+                //Recurrent weight gradients:
+                Nd4j.gemm(aCurrent, dldzNext, rwg, true, false, 1.0, 1.0);
```
AlexDBlack: aCurrent will be null if no layer norm - this will give NPE in that case?

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/params/DefaultParamInitializer.java
```diff
@@ -80,9 +82,10 @@ public long numParams(Layer l) {
         }
     }
 
+
     @Override
     public boolean isWeightParam(Layer layer, String key) {
-        return WEIGHT_KEY.equals(key);
+        return weightKeys(layer).contains(key);
```
AlexDBlack: Let's do this without object creation.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/params/DefaultParamInitializer.java
```diff
@@ -162,6 +181,20 @@ protected INDArray createBias(long nOut, double biasInit, INDArray biasParamView
...
+    protected INDArray createGain(NeuralNetConfiguration conf, INDArray gainParamView, boolean initializeParameters) {
+        org.deeplearning4j.nn.conf.layers.FeedForwardLayer layerConf =
+                (org.deeplearning4j.nn.conf.layers.FeedForwardLayer) conf.getLayer();
+        return createGain(layerConf.getNOut(), layerConf.getGainInit(), gainParamView, initializeParameters);
+    }
+
+    protected INDArray createGain(long nOut, double gainInit, INDArray gainParamView, boolean initializeParameters) {
+        if (initializeParameters) {
+            INDArray ret = Nd4j.valueArrayOf(new long[] {1, nOut}, gainInit);
+            gainParamView.assign(ret);
```
AlexDBlack: Don't create this temporary ret array. Use ```gainParamView.assign(gainInit)``` instead.

libnd4j/include/ops/declarable/generic/transforms/layer_norm.cpp
```diff
@@ -0,0 +1,121 @@
...
+#if NOT_EXCLUDED(OP_layer_norm)
+
+#include <ops/declarable/CustomOperations.h>
+#include <ops/declarable/helpers/reverse.h>
+
+
+namespace nd4j {
+namespace ops  {
+
+    CONFIGURABLE_OP_IMPL(layer_norm, 2, 1, false, 0, -1) {
```
AlexDBlack: No shape function for layer_norm forward pass?

libnd4j/include/ops/declarable/generic/transforms/standardize.cpp
```diff
@@ -0,0 +1,133 @@
...
+#if NOT_EXCLUDED(OP_standardize)
+
+#include <ops/declarable/CustomOperations.h>
+#include <ops/declarable/helpers/reverse.h>
+
+
+namespace nd4j {
+namespace ops  {
+
+    CONFIGURABLE_OP_IMPL(standardize, 1, 1, true, 0, -2) {
```
AlexDBlack: No shape function for standardize op?

libnd4j/include/ops/declarable/generic/transforms/standardize.cpp
```diff
@@ -0,0 +1,133 @@
...
+        
+        std::vector<int> axis;        
+
+        if (block.width() > 1)            
+            axis = INPUT_VARIABLE(1)->template asVectorT<int>();
+        else if (block.numI() > 0) 
+            axis = *block.getIArguments();        
+
+        if(axis.empty()) {      // do not perform standardization
+            output->assign(input);
```
AlexDBlack: Is "no axis" a legitimate use case? More likely to be a bug than "I want to implement identity op via standardize".
IMO exception would be better.

libnd4j/include/ops/declarable/generic/transforms/standardize.cpp
```diff
@@ -0,0 +1,133 @@
...
+        auto output = OUTPUT_VARIABLE(0);
+        std::vector<int> axis;
+
+        if (block.width() == 3)             
+            axis = INPUT_VARIABLE(1)->template asVectorT<int>();
+        else if (block.numI() > 0) 
+            axis = *block.getIArguments();        
+
+        if(axis.empty()) {      // nothing to do in this case
+            output->assign(eps);
```
AlexDBlack: As above (forward pass with no axis).

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/ops/SDMath.java
```diff
@@ -2168,6 +2168,28 @@ public SDVariable step(String name, SDVariable in, double cutoff) {
         return updateVariableNameAndReference(ret, name);
     }
 
+    /**
+     * Standardize input variable along given axis
```
AlexDBlack: Better javadoc would be good. Mainly clarify exactly what you mean by "standardize"

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/custom/LayerNorm.java
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @author Paul Dubs
+ */
+@NoArgsConstructor
+public class LayerNorm extends DynamicCustomOp {
+
+    private boolean noBias = false;
+
+    public LayerNorm(SameDiff sameDiff, SDVariable input, SDVariable gain, SDVariable bias, int... dimensions) {
+        super(null, sameDiff, new SDVariable[] {input, gain, bias}, false);
```
AlexDBlack: Bias can be null - create length 2 array instead of adding null entry.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/custom/LayerNorm.java
```diff
@@ -0,0 +1,110 @@
...
+        this.dimensions = dimensions;
+        addIArgument(dimensions);
```
AlexDBlack: Maybe check for non-null and non-empty dimensions? (here and other constructors)

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/custom/LayerNormBp.java
```diff
@@ -0,0 +1,108 @@
...
+ * @author Paul Dubs
+ */
+@NoArgsConstructor
+public class LayerNormBp extends DynamicCustomOp {
+
+    private boolean noBias = false;
+
+
+    public LayerNormBp(SameDiff sameDiff, SDVariable input, SDVariable gain, SDVariable bias, SDVariable gradient, int... dimensions) {
+        super(null, sameDiff, new SDVariable[] {input, gain, bias, gradient}, false);
```
AlexDBlack: As per LayerNorm - don't add null entry to SDVariable array.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/custom/LayerNormBp.java
```diff
@@ -0,0 +1,108 @@
...
+        this.dimensions = dimensions;
+        addIArgument(dimensions);
+    }
+
+    public LayerNormBp(INDArray input, INDArray gain, INDArray bias, INDArray grad, INDArray dLdx, INDArray dLdg, INDArray dLdb, int... dimensions) {
+        super("layer_norm_bp", new INDArray[]{input, gain, bias, grad}, new INDArray[]{dLdx, dLdg, dLdb});
```
AlexDBlack: No null bias entry here (or throw exception if null not allowed).
Also set noBias field?

nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/opvalidation/LayerOpValidation.java
```diff
@@ -1013,6 +1015,115 @@ public void testConv3dBasic() {
         assertNull(err);
     }
 
+    @Test
+    public void testLayerNorm() {
+        final INDArray random = Nd4j.rand(new int[]{10, 4});
+        final INDArray standardized = Nd4j.emptyLike(random);
```
AlexDBlack: Hm... didn't realize we had an emptyLike... not a good method to use - it's just Nd4j.create but without inheriting datatype. You can use ```random.like()``` or ```random.ulike()``` instead.
https://github.com/deeplearning4j/deeplearning4j/issues/7252

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/recurrent/SimpleRnn.java
treo: it is **rCurrent** that will be null. aCurrent is always defined

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
treo: You are right, it shouldn't be applied.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/layers/BaseLayer.java
treo: It is only necessary for backprop. Will change that accordingly.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/params/DefaultParamInitializer.java
treo: good idea, I'll change it accordingly in the createBias method as well then.

libnd4j/include/ops/declarable/generic/transforms/layer_norm.cpp
treo: CONFIGURABLE_OP_IMPL creates it automatically

PR #7266, opened 2019-03-08
libnd4j: Allow input and output formats to differ in MKL-DNN streams for batchnorm_new, lrn, and lrn_bp

### LINKED ISSUE #7261: libnd4j: MKL-DNN batch normalization given incorrect output for strided input
Permuted input arrays (non-standard strides for input) gives incorrect results for mkl-dnn batch norm.
Yet strides clearly are obtained from the source array:
https://github.com/deeplearning4j/deeplearning4j/blob/master/libnd4j/include/ops/declarable/generic/nn/batchnorm.cpp#L48-L66

```
    @Test
    public void testBatchNormStidedInput(){

        for(boolean useMklDnn : new boolean[]{false, true}) {
            Nd4jCpu.Environment.getInstance().setUseMKLDNN(useMklDnn);

            INDArray in1 = Nd4j.rand('c', new int[]{2, 3, 12, 12});
            INDArray m = Nd4j.rand('c', new int[]{3});
            INDArray v = Nd4j.rand('c', new int[]{3});
            INDArray g = Nd4j.rand('c', new int[]{3});
            INDArray b = Nd4j.rand('c', new int[]{3});

            INDArray out1 = in1.ulike();
            DynamicCustomOp op = DynamicCustomOp.builder("batchnorm_new")
                    .addInputs(in1, m, v, g, b)
                    .addOutputs(out1)
                    .addIntegerArguments(1, 1, 1)
                    .addFloatingPointArguments(1e-3)
                    .build();
            Nd4j.exec(op);

            INDArray out2 = in1.ulike();
            INDArray in2 = Nd4j.create(in1.size(1), in1.size(0), in1.size(2), in1.size(3)).permute(1, 0, 2, 3);
            in2.assign(in1);
            assertTrue(in1.equalShapes(in2));
            assertEquals(in1, in2);

            op.setInputArgument(0, in2);
            op.setOutputArgument(0, out2);

            Nd4j.exec(op);

            assertEquals("Use mkldnn: " + useMklDnn, out1, out2);
        }
    }
```

saudet:
Fixes #7261 

Bug fixes in MKL-DNN streams for batchnorm_new, lrn, and lrn_bp.

## How was this patch tested?

All unit tests of libnd4j pass.

### DIFF
--- a/libnd4j/include/ops/declarable/generic/nn/batchnorm.cpp
+++ b/libnd4j/include/ops/declarable/generic/nn/batchnorm.cpp
@@ -31,9 +31,9 @@ namespace ops {
 #ifdef HAVE_MKLDNN
 using namespace mkldnn;
 
-static void getMKLDNNMemoryDescBatchNorm(const NDArray* src, const NDArray* diff_src,
-        mkldnn::memory::desc* batchnorm_src_md, mkldnn::memory::desc* batchnorm_diff_src_md,
-        mkldnn::memory::desc* user_src_md, mkldnn::memory::desc* user_diff_src_md,int axis) {
+static void getMKLDNNMemoryDescBatchNorm(const NDArray* src, const NDArray* diff_src, const NDArray* dst,
+        mkldnn::memory::desc* batchnorm_src_md, mkldnn::memory::desc* batchnorm_diff_src_md, mkldnn::memory::desc* batchnorm_dst_md,
+        mkldnn::memory::desc* user_src_md, mkldnn::memory::desc* user_diff_src_md, mkldnn::memory::desc* user_dst_md, int axis) {
     const Nd4jLong* shape = src->getShapeInfo();
     Nd4jLong rank = shape[0];
     Nd4jLong dim1 = axis; // MKL-DNN supports only 1 axis, which has to be the "channel" one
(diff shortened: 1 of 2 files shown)


AlexDBlack:
So, unfortunately it looks like test in https://github.com/deeplearning4j/deeplearning4j/pull/7266 still fails.
The changes look like they should fix this, but I'm seeing the same failure after clean build. :(
I did have another look through the code, but nothing stands out to me.

saudet:
Check with MKLDNN_VERBOSE=1 to make sure the reorder operation at the end actually changes.

AlexDBlack:
hm... something else must be up here, not sure what yet.
Now I'm seeing my test fail regardless of whether MKL-DNN is enabled/disabled - and setting MKLDNN_VERBOSE=1 results in no debug output when it is enabled (even though previously I've used it). :/
I'll try master and this PR on a different machine to investigate why this is happening.

saudet:
Are you sure it's running with float and not double data type, for example?

AlexDBlack:
OK, I think there were 2 problems here.
1. Intellij pulled in new snapshots between my build and running the test previously.
2. The non-MKL-DNN implementation of batchnorm_new seems to suffer from the same issue including on master (either this changed recently or perhaps I simply didn't notice it previously).

I can confirm the strided input issue is fixed for MKL-DNN. I will open a separate issue for the non-MKL-DNN implementation.

PR #7311, opened 2019-03-19
Dot Product Attention

treo:
Add dot product attention ops to libnd4j and use them in DL4J layers.

- [x] Dot Product Attention Op
- [x] Multi Headed Dot Product Attention Op
- [X] Self-Attention Layer (apply attention at every timestep with q = timestep)
- [X] Learned Self-Attention Layer (apply attention with q = learned)
- [x] Input Mask support
- [x] Recurrent Attention Layer (apply attention at every timestep with q = hidden state)
- [x] AttentionVertex (takes q, k, v as inputs)

Dropout inside the ops isn't supported yet. And will probably added in a different pull request.


libnd4j/include/helpers/impl/AttentionHelper.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    nd4j::NDArray *
+    AttentionHelper::multiHeadProject(const nd4j::NDArray *input, const nd4j::NDArray *projectionMatrix) {
+        auto miniBatchSize = input->sizeAt(0);
+        auto seqLength = input->sizeAt(2);
+        auto numHeads = projectionMatrix->sizeAt(0);
+        auto projectedSize = projectionMatrix->sizeAt(1);
+
+        auto inputPerm = input->permute({1, 0, 2});
+        auto inputPrep = inputPerm->reshape(input->ordering(), {input->sizeAt(1), (miniBatchSize * seqLength)});
```
AlexDBlack: With this permute - what happens if I call the method with an array with weird strides?
Probably best to add a few extra tests (or grad checks) for this case. Not sure if the libnd4j permute+reshape matches the Nd4J behaviour... but if it does: this could be a problem.

libnd4j/include/helpers/impl/AttentionHelper.cpp
```diff
@@ -0,0 +1,95 @@
...
+    AttentionHelper::multiHeadProjectBp(const nd4j::NDArray *input, const nd4j::NDArray *projectionMatrix,
+                                        const nd4j::NDArray *eps, nd4j::NDArray *dLdInput,
+                                        nd4j::NDArray *dLdProjectionMatrix) {
+        auto miniBatchSize = input->sizeAt(0);
+        auto seqLength = input->sizeAt(2);
+        auto numHeads = projectionMatrix->sizeAt(0);
+        auto projectedSize = projectionMatrix->sizeAt(1);
+
+        auto epsPerm = eps->permute({1, 2, 0, 3});
+        auto epsReshaped = epsPerm->reshape(eps->ordering(), {numHeads * projectedSize, miniBatchSize * seqLength});
```
AlexDBlack: Maybe same permute/reshape as mentioned earlier.

libnd4j/include/ops/declarable/generic/nn/dot_product_attention.cpp
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        if(mask != nullptr){
+            NDArray* reshapedMask;
+            if(weights->rankOf() == 4){
+                reshapedMask = mask->reshape(mask->ordering(), {mask->sizeAt(0), 1, mask->sizeAt(1), 1});
+            }else{
+                reshapedMask = mask->reshape(mask->ordering(), {mask->sizeAt(0), mask->sizeAt(1), 1});
+            }
+            *weights += (*reshapedMask - 1) * 1e9;
```
AlexDBlack: Maybe add a comment here - seemed a little strange on first glance.
I gather you want masked steps to have very large negative attention weights as a way of doing "masked softmax"?
How was the 1e9 chosen? (just wondering if we'll even have numerical stability issues...)

nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/opvalidation/ReductionOpValidation.java
```diff
@@ -1058,4 +1060,168 @@ public void testReductionsBackwards() {
             sd.execBackwards(Collections.emptyMap());
         }
     }
+
+    @Test
+    public void testDotProductAttention(){
```
AlexDBlack: As noted in earlier comment - maybe let's improve all of these with loop testing c/f order input arrays, at least; testing weird strides/permuted would be even better.

nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/opvalidation/ReductionOpValidation.java
```diff
@@ -1058,4 +1060,168 @@ public void testReductionsBackwards() {
...
+        SDVariable sdK = sd.var("k", keys);
+        SDVariable sdV = sd.var("v", values);
+
+        SDVariable t = sd.nn.dotProductAttention(sdQ, sdK, sdV, null, true);
+        t.norm1("out");
+
+        String err = OpValidation.validate(new TestCase(sd)
+                    .expectedOutput("out", finalOut)
+                    .gradientCheck(true));
+        assertNull(err, err);
```
AlexDBlack: Really minor point: just assertNull(err) is fine. It actually prints the object if it fails, even without the string set as the message.
(And yes, I know a ton of my tests do have assertNull(err,err) but I'm cleaning those as I see them)

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/RecurrentAttentionLayer.java
```diff
@@ -0,0 +1,326 @@
...
+
+    @Override
+    public void applyGlobalConfigToLayer(NeuralNetConfiguration.Builder globalConfig) {
+        if (activation == null) {
+            activation = SameDiffLayerUtils.fromIActivation(globalConfig.getActivationFn());
+        }
+    }
+
+    @Override
+    public SDVariable defineLayer(SameDiff sameDiff, SDVariable layerInput, Map<String, SDVariable> paramTable, SDVariable mask) {
```
AlexDBlack: This implementation will only work with sequences that are identical length for all minibatches - but we don't have any checks for that anywhere as far as I can see... let's add checks and also test that. Let's also open an issue about it (once merged), and switch this to proper SameDiff loops when they are ready.

Also not sure how safe/reliable the ```final long[] inputShape = layerInput.getShape();  final val timeSteps = inputShape[2];``` is, though I guess if tests are OK...

libnd4j/include/ops/declarable/generic/nn/dot_product_attention.cpp
treo: The 1e9 is used by the tensor2tensor package as well. Bert on the other hand uses just 1e4, while GPT-2 uses 1e10.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/LearnedSelfAttentionLayer.java
```diff
@@ -0,0 +1,261 @@
...
+import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.linalg.primitives.Pair;
+
+import java.util.Map;
+
+/**
+ * Implements Dot Product Self Attention with learned queries
+ *
+ * Takes in RNN style input in the shape of [batchSize, features, timesteps]
+ * and applies dot product attention using learned queries.
```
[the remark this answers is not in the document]
treo: What kinds of format were you looking for? As this has an input preprocessor configured, it should support any of the common shapes of DL4J layer outputs

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/RecurrentAttentionLayer.java
treo: About the same time that you've commented, I noticed that as an issue as well. Not really sure how to check for different sequence length here though, as `defineLayer` is always only called once.

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/LearnedSelfAttentionLayer.java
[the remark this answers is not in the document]
AlexDBlack: Hm, good point, we can just permute in DL4J preprocessor. OK, disregard then :)

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/RecurrentAttentionLayer.java
AlexDBlack: Two ways I can think of doing this...
One is we solve the general problem here: https://github.com/deeplearning4j/deeplearning4j/issues/7527
Though that might take too long.

Another is we add some sort of "validateInput(INDArray[])" method on the DL4J samediff layers, that defaults to no-op (but can be overridden if required). Then call that just before forward pass. That should be straightforward to add I think.

treo: I think for now we can go with the second option. I'll add it to the SameDiffLayer and SameDiffVertex base classes

treo:
Finally got around to addressing all your points. Good catch on the permute/reshape thing. They do behave differently in libnd4j, but there was still an issue that was uncovered by adding the c/f order tests. For weird strides, they should (according to comments) handle the case already.

I've opened an issue at https://github.com/deeplearning4j/deeplearning4j/issues/7552 to add optional forward masking.

PR #7317, opened 2019-03-20
Few corrections for easier parallelization

yquemener:
Extracted some variables from loops and conditions for easier OpenMP parallelization.

### DIFF
--- a/libnd4j/include/graph/RandomGenerator.h
+++ b/libnd4j/include/graph/RandomGenerator.h
@@ -139,8 +139,11 @@ namespace nd4j {
 
         FORCEINLINE RandomGenerator::RandomGenerator(Nd4jLong rootSeed, Nd4jLong nodeSeed) {
             // this seed is used graph-level state
-            if (rootSeed == 0)
-                rootSeed = currentMilliseconds();
+            if (rootSeed == 0){
+                auto s = std::chrono::system_clock::now().time_since_epoch();
+                rootSeed  = std::chrono::duration_cast<std::chrono::milliseconds>(s).count();
+            }
+
 
             // graph-level state is just first seed
             _rootState._long = rootSeed;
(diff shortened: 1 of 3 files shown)


libnd4j/include/graph/RandomGenerator.h
```diff
@@ -139,8 +139,11 @@ namespace nd4j {
 
         FORCEINLINE RandomGenerator::RandomGenerator(Nd4jLong rootSeed, Nd4jLong nodeSeed) {
             // this seed is used graph-level state
-            if (rootSeed == 0)
-                rootSeed = currentMilliseconds();
+            if (rootSeed == 0){
+                auto s = std::chrono::system_clock::now().time_since_epoch();
+                rootSeed  = std::chrono::duration_cast<std::chrono::milliseconds>(s).count();
```
saudet: This is basically what currentMilliseconds() does, how is this different?

yquemener: In cases when inlining is deactivated (happened with ncc at one point), this results in a constructor calling a member function, which is generally fine, except in one case: https://stackoverflow.com/questions/3091833/calling-member-functions-from-a-constructor/31757790#31757790

I can try to retrieve the case where this was happening, one of the tests was failing because of that.

saudet: We can make currentMilliseconds() static, that should be fine, right?

yquemener: Probably. The constructor used to call SetStates, which was problematic, but currentMilliseconds() is probably no problem.

saudet: Ok, let's try to make it static, and revert these lines to call that function again.

saudet:
@shyrma @shugeo @raver119 
If you add more loops like in this PR, please do them the same way! Thanks

Issue #7368, opened 2019-03-27
Add reshape, permute and other common SameDiff operations to SDVariable


rnett:
There are a number of common unary operations that only exist in SameDiff.  In particular, reshape and permute.

You could make a case for moving reverse, repeat, sum, max, etc, but I don't think they are used frequently enough for it to be necessary, where I think reshape and permute are.

I can implement this if approved.

AlexDBlack:
We have reshape(SDVariable) already, we should add a reshape(long...) IMO. I'm in favor of adding permute (that's very common) and dot methods too.

Sum, max etc are staying. They are very common.
Repeat and reverse - there's no SDVariable repeat/reverse methods. We don't need them on SDVariable IMO.

rnett:
Ah, I don't know how I missed that max etc were already there.

I will definitely add reshape and permute for int... and long..., and a permute for SDVariable, and dot.

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/7371

Issue #7367, opened 2019-03-27
Have SDVariable use Kotlin operator method names


rnett:
SDVariable has quite a few methods that Kotlin will provide operators for if named correctly, like the plus, multiply, ect.  However, most of them are named wrong to automatically have operators.

We can either change the names or create overloads.  Imo changing is cleaner, although it will break stuff.

 * `add` to `plus`
 * `sub` to `minus`
 * `mul` to `times`

If the `addi` type methods overwrite the current variable (e.g. `this.addi(other)` stores the result in `this`) they could be changed to `plusAssign` type methods, but those names are less clean and not as used, so they are probably fine.

Can make a PR if/when the best way to do it is decided.

treo:
Another way to deal with this may be to just create a kotlin helper library, that has those overloads as extension functions.

rnett:
Yeah, I've done that and it works, but it a bit clunky and I don't see much of a downside to just changing the names.

treo:
DL4J is used by quite a lot of people, breaking their code sounds like a lot of downside to me

AlexDBlack:
Yeah, I'm opposed to renaming them, as we want SDVariable to match INDArray as much as possible to minimize the learning/switching costs. Plus breaking API changes are something we try to avoid unless there's a very good reason (and I'm not sure this is good enough).

Another option that that might work is adding overloads: so keep SDVariable.add, but introduce an SDVariable.plus method that is an alias - i.e., just calls add internally.

rnett:
Yeah, I brought that up in the OP, for exactly that reason.  Although it still clutters things up a bit.

Would it be worth marking the originals as deprecated and having them call the new ops instead?

AlexDBlack:
> Would it be worth marking the originals as deprecated and having them call the new ops instead?

No, I'm opposed to deprecating add etc, for the same reassons I outlined earlier. Unless we change INDArray to match (and we're not doing that) it just creates additional cognitive load for most users, just so Kotlin users can have a slightly easier time. :)

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/7371

Issue #7364, opened 2019-03-27
Non-SDVariable invertPermutation


rnett:
There is a quite nice `invertPermutation` method in SameDiff, but it only applies for SDVariable permute sequences.  It would be nice to have a utility `invertPermutation` method for int and long arrays, probably not in `SameDiff` (it is also useful for NDArray permutations, etc.).

I can implement and PR this if someone tells me where to put it.

AlexDBlack:
Sure, PR for this would be good.
I think this would be an appropriate place: https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-common/src/main/java/org/nd4j/linalg/util/ArrayUtil.java

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/pull/7371

PR #7371, opened 2019-03-27
SDVariable Improvements

rnett:

Fixing:
- [x] #7364
- [x] #7367
- [x] #7368

## How was this patch tested?

- [x] Unit tests for invertPermutation

Should I add unit tests for the SDVar methods?  They all call already implemented methods.


rnett:
I can add a `permute(long...)` that just casts to int, is that useful enough to be worth it?  I always end up using ints.

PR #7374, opened 2019-03-27
[WIP] Fixed possible NPE in org.nd4j.graph.ui.LogFileWriter.writeHistogramEventDiscrete

dariuszzbyrad:

Fixed possible NPE in org.nd4j.graph.ui.LogFileWriter.writeHistogramEventDiscrete.

When you pass the null value as parameter for 'binLabels' or 'y', you can see the NullPointerException thrown out. Expected behaviour is log IllegalStateException with full details message, e.g. 
'Number of bin labels (if present) must be same as Y array length (...)'

## How was this patch tested?

Writing unit test and check locally. I'm not sure do we prefer many tests for testing preconditions for all methods, so I didn't commit it. If you have a different opinion, let me know, I'll add it.

```
LogFileWriter logFileWriter = new LogFileWriter(new File("test"));
logFileWriter .writeHistogramEventDiscrete(null, 0l, 0, 0, null, new NDArray());
```

## Quick checklist


AlexDBlack:
Thanks for the pull request, but I'm not going to merge this for 2 reasons.
First, the Y argument here is not optional - it must always be present. Adding an additional check for this being null would be worthwhile.
A null check on name would be worthwile too. For both cases, a lombok ```@NonNull``` should be sufficient; a Preconditions.checkArgument would work fine too.

As for the bin labels being null - the problem is actually the ```(long)binLabels.size()```. It should be ```(binLabels = null ? 0 : (long)binLabels.size()```

Second: there's 17 commits in this pull request for a 1 line change.

Finally, a note your test - don't do this: ```new NDArray()``` - it's not backend independent (i.e., hardcoded to CPU) and also won't correctly initialize the array. Instead, always use the Nd4j.create methods and similar.


Thanks for flagging this (it's definitely a legitimate bug) but I'll fix this shortly in a pull request I already have open for other reasons.

AlexDBlack:
A fix was added here, should be merged soon: https://github.com/deeplearning4j/deeplearning4j/pull/7340

dariuszzbyrad:
Thanks for the CR.

PR #7392, opened 2019-03-29
Add a keepDim option to point SDIndexes

rnett:

Adds a keepDim option to point SDIndexes.  If true, the target dimension will be reduced to size 1 instead of removed.

## How was this patch tested?

N/A.  Get has no unit tests and the changes are very simple.


rnett:
Huh, that didn't show up in my usage search, I will add some more.

rnett:
My tests are passing in jenkins, I think we're good.

> Tests run: 115, Failures: 0, Errors: 0, Skipped: 4, Time elapsed: 0.582 sec - in org.nd4j.autodiff.samediff.SameDiffTests

rnett:
Partially fixes #7394 too (better array, could still probably use more tests).

PR #7403, opened 2019-03-30
[GraphBuilder] Use last added layer/vertex if no inputs specified

rnett:

When building a computation graph with `ComputationGraphConfiguration.GraphBuilder`, when you add a layer and don't specify an input, use the last added layer/vertex (if you have previously added at least one).

## How was this patch tested?

Removed test that expected exception that now will not be thrown.
Changed one other test to use this method of adding inputs.


AlexDBlack:
So, this is an interesting usability idea.
My main concern is that more often than not no input names means user forgot to include it, rather than it being intentional. This check has saved me from making mistakes in network structure more than a few times.

Now, what might be better is a new method: something like "appendLayer(String, Layer)" Not sure on the best name for that method. But with that, it's explicit - there's no ambiguity for "no input means user forgot" vs. "no input means user wants last layer/vertex to be used".

rnett:
Yeah, that makes sense.  I can't think of anything better than `appendLayer`.

rnett:
What do you think about also doing it if `null` is passed to addVertex/Layer?  There is some ambiguity in the method call as to whether you pass it as null or [null] but I can check for both.

AlexDBlack:
Not sure if you missed it or haven't got to it, but Javadoc still mentions previous null behaviour, and there's a removed test in testInvalidConfigurations() that we should keep.

rnett:
Should both be fixed.

Issue #7405, opened 2019-03-30
SameDiff mean w/ dimensions gradient is incorrect


rnett:
#### Issue Description

The gradient of the `SameDiff.mean` is always the gradient as if it is calculated over all dimensions, even if dimensions are specified.

See the tensorflow equivalent calculations [here](https://gist.github.com/rnett/179c75740774c807dc0761ec5060188c#file-mean-gradient-test-ipynb).

The DL4J equivilant looks like:
```java
@Test
public void test1() {
    SameDiff SD = SameDiff.create();
    SDVariable a = SD.reshape("a", SD.linspace("at", 1, 15, 15), 3, 5);//.add(1);
    SDVariable b = SD.one("b", 3, 5);//.add(3);

    SDVariable out = a.mul(b).mean(1);

    //System.out.println(Arrays.toString(out.eval().shape()));
    out.eval();

    SD.execBackwards(null, Lists.asList("a", new String[]{}));
    //System.out.println(out.eval());
    System.out.println(SD.grad("a").eval());
}

@Test
public void test2() {
    SameDiff SD = SameDiff.create();
    SDVariable a = SD.reshape("a", SD.linspace("at", 1, 15, 15), 3, 5);//.add(1);
    SDVariable b = SD.one("b", 3, 5);//.add(3);

    SDVariable out = a.mul(b).mean(0, 1);

    //System.out.println(Arrays.toString(out.eval().shape()));
    out.eval();

    SD.execBackwards(null, Lists.asList("a", new String[]{}));
    //System.out.println(out.eval());
    System.out.println(SD.grad("a").eval());
}
```
and prints
```java
[[    0.0667,    0.0667,    0.0667,    0.0667,    0.0667], 
 [    0.0667,    0.0667,    0.0667,    0.0667,    0.0667], 
 [    0.0667,    0.0667,    0.0667,    0.0667,    0.0667]]
```
for both tests.

#### Version Information

Please indicate relevant versions, including, if relevant:

* Deeplearning4j version: 1.0.0-SNAPSHOT
* platform information (OS, etc): Windows 10
* CPU Backend

AlexDBlack:
I'm not sure the tests are equivalent.

https://www.tensorflow.org/versions/r1.9/api_docs/python/tf/gradients?hl=en
> gradients() adds ops to the graph to output the derivatives of ys with respect to xs. It returns a list of Tensor of length len(xs) where each tensor is the sum(dy/dx) for y in ys.

[...]

PR #7406, opened 2019-03-30
Wasserstein loss function

rnett:

Added the Wasserstein loss function.

Forward pass: `mean(labels * output, dims=1)`
Backwards pass: `labels/labels.shape(1)`

## How was this patch tested?

None.  Should I add tests?  I didn't see tests for any other losses except for crossentropy.

Note that the gradient doesn't match SameDiff's b/c of #7405.


rnett:
Note that the improved Wasserstein loss is better, but requires the 2nd degree gradient, which I don't think we can calculate?  Specifically, it adds a `gradient - 1` term to the loss.

rnett:
Tests are in and the gradients pass.

The tests fail the serialization test because of an extra space at the end of the expected serialized string here:
```
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.deeplearning4j.TestUtils.testModelSerialization(TestUtils.java:63)
	at org.deeplearning4j.gradientcheck.LossFunctionGradientCheck.lossFunctionGradientCheckLossLayer(LossFunctionGradientCheck.java:373)
```

AlexDBlack:
> The tests fail the serialization test because of an extra space at the end of the expected serialized string 

I haven't seen that before in any of the tests :/
I'll check out and build locally

AlexDBlack:
OK, so the problem here is the equals method in DifferentialFunction.
When the equality check fails, it does toString which returns network JSON - that extra space isn't relevant here.
Let's just switch to `@EqualsAndHashCode(callSuper = false)` - superclass fields aren't relevant/used, so we can ignore them. Test passes with that.

PR #7410, opened 2019-04-01
[WIP] Weight sharing

rnett:

Delegate parameters to `WeightPool`, allow sharing of weight pools based on keys.  Added a `cloneAndShareWeights()` method to take advantage of this.

Works for layers and graph vertices (_todo_, although the only vertex with weights is `SameDiffGraphVertex`).

Also adds methods to copy a subset of a graph/model configuration with shared weights (_todo_).

## How was this patch tested?

Planed:
- [ ] SameDiff weight sharing
- [ ] SameDiffOutput weight sharing
- [ ] BaseLayer subclass weight sharing
- [ ] SameDiffVertex weight sharing
- [ ] Config serialization w/ shared weights
- [ ] Model serialization w/ shared weights


AlexDBlack:
So far you've basically skipped all the hard parts of parameter sharing as it relates to backprop. :)
This will not work for training, without considerable modifications.

I appreciate the enthusiasm, but mind holding off on further work on this until we settle on a design for the internals that will work? There's some non-trivial issues here that you haven't run into yet.

rnett:
Yeah, sure, I didn't realize it would be more complicated than this.

agibsonccc:
One suggestion would be to just wrap the logic you want in samediff layer.
Weight sharing there would be easier.

AlexDBlack:
@agibsonccc yeah, we talked at length in gitter
This was the first steps for weight sharing between nets... which is easy in SameDiff already (just associate same array with multiple nets)... and requires a ton of re-engineering for DL4J (not worth it) due to some early design decisions there

PR #7420, opened 2019-04-01
Support multiple use of computation graph vertex as input to other vertices

treo:
While building a test case for the attention vertex, I've stumbled across the use case were I needed to use the same vertex output three times as the input. 

In that specific case it was that I was using the output of an lstm for queries, keys and values. 

The reason why it was failing was that `indexOf` will always only find the first appearance of an entry within a list.

### DIFF
--- a/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
+++ b/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
@@ -16,8 +16,11 @@
 
 package org.deeplearning4j.nn.graph;
 
-import lombok.*;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+import lombok.val;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/graph/ComputationGraph.java
```diff
@@ -644,15 +646,17 @@ public void init(INDArray parameters, boolean cloneParametersArray) {
                 continue; //Output vertex
             VertexIndices[] outputIndices = new VertexIndices[thisVertexOutputsTo.size()];
             int j = 0;
-            for (String s : thisVertexOutputsTo) {
+            for (String s : new HashSet<>(thisVertexOutputsTo)) {
```
AlexDBlack: I don't think we need this new hashset construction here... we're not modifying thisVertexOutputsTo in the loop, so it should be fine as-is?

treo: thisVertexOutputsTo contains duplicates that we need in order to calculate the size of outputIndices correctly. But once we start iterating over it we are only interested in the unique values. Creating a HashSet was just the smallest change that I could think of.

treo:
Sure, I'll add gradient checks. It is how I ran into the issue in the first place (i.e. while creating gradient checks for the AttentionVertex)

PR #7422, opened 2019-04-02
SameDiff output/inference methods and many to one evaluation

rnett:

Add `ComputationGraph` style `output` methods to `SameDiff` to do inference.  It isn't particularly hard to do without them, but this keeps the apis similar and makes it a bit easier.

Also adds a many input one output `evaluate` method.

## How was this patch tested?

Added a output test on an untrained SameDiff that simply adds 2 to the input in `SameDiffOutputTest`.


rnett:
For whatever reason, the arrays don't match if I use `equalsWithEps`, but do if I compare strings.

nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/samediff/SameDiffOutputTest.java
```diff
@@ -0,0 +1,69 @@
...
+                .updater(new Sgd(3e-1))
+                .dataSetFeatureMapping("input")
+                .dataSetLabelMapping()
+                .build();
+
+        sd.setTrainingConfig(conf);
+
+        INDArray output = sd.output(data, "out").get("out");
+
+        assertTrue(output.toString().equals(Nd4j.zeros(10, 10).add(2).toString()));
```
AlexDBlack: The reason it isn't equal without the toString is datatypes.
By design we don't cast on equals.
I'd recommend either specifying datatypes on all INDArrays, or on none of them.
Tests default to double; you are comparing a float [10,10] with double [10,10] hence the failure.

Issue #7287, opened 2019-03-12
SameDiff/DL4J: Add iterators for BERT training


AlexDBlack:
Basically: something like CnnSentenceDataSetIterator, but for BERT.
i.e., raw text in, MultiDataSets out (i.e., a MultiDataSetIterator implementation).
Also with utility methods for performing inference.
This should be usable for both TF import models, and later DL4J models built from scratch.

Edit: we probably want 2 variants here, for both:
(a) Unsupervise pretraining
(b) Supervised classification (single doc/sentence input tasks as well as pairs of docs/sentence tasks)

Aha! Link: https://skymindai.aha.io/features/DL4J-4

PR #7430, opened 2019-04-02
[WIP] Add BertIterator (MultiDataSetIterator for BERT training)

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/7287

### DIFF
--- a/deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
+++ b/deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
@@ -100,7 +100,8 @@ import java.util.Map;
  * Only relevant when the task is set to {@link Task#UNSUPERVISED}. Determine the format of the labels:<br>
  * <b>RANK2_IDX</b>: return int32 [minibatch, numTokens] array with entries being class numbers. Example use case: with sparse softmax loss functions.<br>
  * <b>RANK3_NCL</b>: return float32 [minibatch, numClasses, numTokens] array with 1-hot entries along dimension 1. Example use case: RnnOutputLayer, RnnLossLayer<br>
- * <b>RANK3_NLC</b>: return float32 [minibatch, numTokens, numClasses] array with 1-hot entries along dimension 2<br>
+ * <b>RANK3_LNC</b>: return float32 [numTokens, minibatch, numClasses] array with 1-hot entries along dimension 2. This format is occasionally
+ * used for some RNN layers in libraries such as TensorFlow, for example<br>
  * <br>
  */
 public class BertIterator implements MultiDataSetIterator {
@@ -108,7 +109,7 @@ public class BertIterator implements MultiDataSetIterator {
     public enum Task {UNSUPERVISED, SEQ_CLASSIFICATION}
     public enum LengthHandling {FIXED_LENGTH, ANY_LENGTH, CLIP_ONLY}
     public enum FeatureArrays {INDICES_MASK, INDICES_MASK_SEGMENTID}
-    public enum UnsupervisedLabelFormat {RANK2_IDX, RANK3_NCL, RANK3_NLC}
+    public enum UnsupervisedLabelFormat {RANK2_IDX, RANK3_NCL, RANK3_LNC}
 
     protected Task task;
     protected TokenizerFactory tokenizerFactory;
@@ -277,8 +278,8 @@ public class BertIterator implements MultiDataSetIterator {
                 labelArr = Nd4j.create(DataType.INT, mbPadded, outLength);
             } else if(unsupervisedLabelFormat == UnsupervisedLabelFormat.RANK3_NCL){
                 labelArr = Nd4j.create(Nd4j.defaultFloatingPointType(), mbPadded, vocabSize, outLength);
-            } else if(unsupervisedLabelFormat == UnsupervisedLabelFormat.RANK3_NLC){
-                labelArr = Nd4j.create(Nd4j.defaultFloatingPointType(), mbPadded, outLength, vocabSize);
+            } else if(unsupervisedLabelFormat == UnsupervisedLabelFormat.RANK3_LNC){
+                labelArr = Nd4j.create(Nd4j.defaultFloatingPointType(), outLength, mbPadded, vocabSize);
             } else {
                 throw new IllegalStateException("Unknown unsupervised label format: " + unsupervisedLabelFormat);
             }
@@ -297,8 +298,8 @@ public class BertIterator implements MultiDataSetIterator {
                             labelArr.putScalar(i, j, targetTokenIdx);
                         } else if(unsupervisedLabelFormat == UnsupervisedLabelFormat.RANK3_NCL){
                             labelArr.putScalar(i, j, targetTokenIdx, 1.0);
-                        } else if(unsupervisedLabelFormat == UnsupervisedLabelFormat.RANK3_NLC){
-                            labelArr.putScalar(i, targetTokenIdx, j, 1.0);
+                        } else if(unsupervisedLabelFormat == UnsupervisedLabelFormat.RANK3_LNC){
+                            labelArr.putScalar(j, i, targetTokenIdx, 1.0);
                         }
 
                         lMask.putScalar(i, j, 1.0);
--- a/deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
+++ b/deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
@@ -45,31 +45,61 @@ import java.util.Map;
  * (b) Supervised - For sequence classification (i.e., 1 label per sequence, typically used for fine tuning)<br>
  * The task can be specified using {@link Task}.
  * <br>
- * Example for unsupervised training:<br>
+ * <b>Example for unsupervised training:</b><br>
  * <pre>
  * {@code
- *
+ *          BertWordPieceTokenizerFactory t = new BertWordPieceTokenizerFactory(pathToVocab);
+ *          BertIterator b = BertIterator.builder()
+ *              .tokenizer(t)
+ *              .lengthHandling(BertIterator.LengthHandling.FIXED_LENGTH, 16)
+ *              .minibatchSize(2)
+ *              .sentenceProvider(<sentence provider here>)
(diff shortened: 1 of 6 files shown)


deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
```diff
@@ -0,0 +1,479 @@
...
+/**
+ * BertIterator is a MultiDataSetIterator for training BERT (Transformer) models in the following way:<br>
+ * (a) Unsupervised - Masked language model task (no sentence matching task is implemented thus far)<br>
+ * (b) Supervised - For sequence classification (i.e., 1 label per sequence, typically used for fine tuning)<br>
+ * The task can be specified using {@link Task}.
+ * <br>
+ * Example for unsupervised training:<br>
+ * <pre>
+ * {@code
+ *
```
treo: I guess you forgot to actually add an example here?

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
```diff
@@ -0,0 +1,479 @@
...
+ * </pre>
+ *
+ * This iterator supports numerous ways of configuring the behaviour with respect to the sequence lengths and data layout.<br>
+ * <br>
+ * <u><b>{@link LengthHandling} configuration:</b></u><br>
+ * Determines how to handle variable-length sequence situations.<br>
+ * <b>FIXED_LENGTH</b>: Always trim longer sequences to the specified length, and always pad shorter sequences to the specified length.<br>
+ * <b>ANY_LENGTH</b>: Output length is determined by the length of the longest sequence in the minibatch<br>
+ * <b>CLIP_ONLY</b>: For any sequences longer than the specified maximum, clip them. If the maximum sequence length in
+ * a minibatch is shorter than the specified maximum, no padding will occur.<br>
```
treo: What happens in the ANY_LENGTH and CLIP_ONLY for the shorter sequences? Padding?

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
```diff
@@ -0,0 +1,479 @@
...
+ *<br><br>
+ * <u><b>{@link FeatureArrays} configuration:</b></u><br>
+ * Determines what arrays should be included.
+ * <b>INDICES_MASK</b>: Indices array and mask array only, no segment ID array<br>
+ * <b>INDICES_MASK</b>: Indices array, mask array and segment ID array (which is all 0s for single segment tasks)<br>
```
treo: Both cases are INDICES_MASK here

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
```diff
@@ -0,0 +1,479 @@
...
+ * <u><b>{@link UnsupervisedLabelFormat} configuration:</b></u><br>
+ * Only relevant when the task is set to {@link Task#UNSUPERVISED}. Determine the format of the labels:<br>
+ * <b>RANK2_IDX</b>: return int32 [minibatch, numTokens] array with entries being class numbers<br>
+ * <b>RANK3_NCL</b>: return float32 [minibatch, numClasses, numTokens] array with 1-hot entries along dimension 1<br>
+ * <b>RANK3_NLC</b>: return float32 [minibatch, numTokens, numClasses] array with 1-hot entries along dimension 2<br>
+ * <br>
+ */
+public class BertIterator implements MultiDataSetIterator {
+
+    public enum Task {UNSUPERVISED, SEQ_CLASSIFICATION}
```
treo: If I understand correctly, in the paper they are actually pretraining the model with both tasks at the same time. So we should support that case as well.

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
```diff
@@ -0,0 +1,479 @@
...
+ * Determines what arrays should be included.
+ * <b>INDICES_MASK</b>: Indices array and mask array only, no segment ID array<br>
+ * <b>INDICES_MASK</b>: Indices array, mask array and segment ID array (which is all 0s for single segment tasks)<br>
+ * <br>
+ * <u><b>{@link UnsupervisedLabelFormat} configuration:</b></u><br>
+ * Only relevant when the task is set to {@link Task#UNSUPERVISED}. Determine the format of the labels:<br>
+ * <b>RANK2_IDX</b>: return int32 [minibatch, numTokens] array with entries being class numbers<br>
+ * <b>RANK3_NCL</b>: return float32 [minibatch, numClasses, numTokens] array with 1-hot entries along dimension 1<br>
+ * <b>RANK3_NLC</b>: return float32 [minibatch, numTokens, numClasses] array with 1-hot entries along dimension 2<br>
+ * <br>
```
treo: Maybe add a short explanation when someone would use each of those options

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
```diff
@@ -0,0 +1,479 @@
...
+            if(sentenceProvider != null){
+                numClasses = sentenceProvider.numLabelClasses();
+                List<String> labels = sentenceProvider.allLabels();
+                for(int i=0; i<mb; i++ ){
+                    String lbl = tokenizedSentences.get(i).getRight();
+                    classLabels[i] = labels.indexOf(lbl);
+                    Preconditions.checkState(classLabels[i] >= 0, "Provided label \"%s\" for sentence does not exist in set of classes/categories", lbl);
+                }
+            } else {
+                throw new RuntimeException();
```
treo: Can this case even happen? Earlier in the method you are already throwing an exception if sentencProvider is null

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
```diff
@@ -0,0 +1,479 @@
...
+        /**
+         * Specify what arrays should be returned. See {@link BertIterator} for more details.
+         */
+        public Builder featureArrays(FeatureArrays featureArrays){
+            this.featureArrays = featureArrays;
+            return this;
+        }
+
+        /**
+         * Provide the vocabulary as a map. Keys are
```
treo: Looks like an unfinished sentence?

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/bert/BertMaskedLMMasker.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * See: {@link BertMaskedLMMasker} for details.
+     * @param r                 Random number generator
+     * @param maskProb          Probability of masking each token
+     * @param maskTokenProb     Probability of replacing a selected token with the mask token
+     * @param randomTokenProb    Probability of replacing a selected token with a random token
+     */
+    public BertMaskedLMMasker(Random r, double maskProb, double maskTokenProb, double randomTokenProb){
+        Preconditions.checkArgument(maskProb > 0 && maskProb < 1, "Probability must be beteen 0 and 1, got %s", maskProb);
+        this.r = r;
```
treo: maybe also check that 0 <= maskTokenProb + randomTokenProb <= 1?

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
AlexDBlack: Yes, I'll add that.

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
AlexDBlack: We pad only to the longest sequence in the minibatch.

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/iterator/BertIterator.java
AlexDBlack: Not yet, but it's a guard against future incomplete changes.

AlexDBlack:
@treo Thanks for the review

> From my understanding of the paper, the big thing about Bert is that they pre-train using both the MLM / Cloze task and next sentence classification task at the same time. 

Yes, that is my understanding also, and the omission was intentional at this point. I want to get something usable sooner rather than later, even if it's not optimal. My main goal for now was the sequence classification task as that's what we'll be using initially for transfer learning.

As for supporting the "next sentence classification" task, I want that. But it'll be more complex to implement: instead of iterating over sentences, we'll want to iterate over documents, and we'l need some automated way to break text into sentences.
Alternatively (or, additionally), we introduce a "labelled sentence pair iterator" for this.

So, future additions we'll need here:
* Sentence matching task (supervised)
* Next sentence classification task
* Per token classification - https://github.com/deeplearning4j/deeplearning4j/issues/7436

There's some non-trivial challenges with those, and I planned to address them separately. (Javadoc should reflect that, however)

> The SEQ_CLASSIFICATION task also doesn't look like it is not an instance of the next-sentence classification task.

It's not. It's designed for transfer learning, single sequence in, single label out... I'll clarify that.

treo:
Great, with the clarification that the intent wasn't to support all the unsupervised pre-training options yet, and the fixes in the java doc, only the question of when you'd use RANK3_NLC, my guess is that some of the imported models may need it?

Issue #7424, opened 2019-04-02
SameDiff: Add a way to mark something as non-differentiable / frozen


treo:
There are some cases where we may want to be able to mark something as non-differentiable, or frozen, as it and functions that only take such frozen inputs should not be differentiated. 

In my specific example I have to combine two mask arrays in an attention vertex, and backprop fails because the consuming op doesn't provide a mask gradient (which would be somewhat silly to do).

Issue #6876, opened 2018-12-17
SameDiff: Fix "gradient" calculation for non-floating point arrays


AlexDBlack:
Consider the following simple case:

```
    @Test
    public void testNonFpGrad(){
        SameDiff sd = SameDiff.create();
        SDVariable v1 = sd.var("v1", Nd4j.create(DataType.LONG, 3,4));
        SDVariable v2 = sd.var("v2", Nd4j.create(DataType.LONG, 3,4));

        SDVariable add = v1.add(v2);
        SDVariable sum = add.castTo(DataType.DOUBLE).sum();

        sd.execBackwards();
    }
```
Which (on my current branch) gives:
```
o.n.l.c.n.o.NativeOpExecutioner - Failed to execute op add_bp. Attempted to execute with 3 inputs, 2 outputs, 0 targs and 0 iargs. Inputs: [(LONG,[3,4],c), (LONG,[3,4],c), (LONG,[3,4],c)]. Outputs: [(LONG,[3,4],c), (LONG,[3,4],c)] - Please see above message (printed out from c++) for a possible cause of error.
```

Clearly, the problem here is that we are trying to calculate the "gradient" with respect to arrays with LONG datatype. However, by definition gradients are only defined for real-valued arrays.

An obvious hack here is to just return 0 "gradient" for these non-FP variables, but this is inefficient (still requires array allocations, memory, etc) and a little ugly (we still create gradient SDVariables corresponding to the "gradient" of these long arrays.

We need a better solution that:
- Doesn't involve creating/returning 0s as a workaround when gradients aren't defined/don't exist
- Actually validates that we aren't trying to create "gradient" SDVariables for non-FP SDVariables
- Terminates the gradient calculation procedure early, accounting for situations where gradients are not defined even for some floating point variables
    - For example: float -> cast to long type: the float variable can legitimately not have any gradient

raver119:
Hm, we shouldn't be trying to calculate those. It should fail dtypes validation.

AlexDBlack:
> Hm, we shouldn't be trying to calculate those. It should fail dtypes validation.

Sure, we should detect it earlier. But failing earlier doesn't change the fact that we should never have created that particular add_bp (or whatever) op in the first place :)

We'll also need to do a pass on gradient/backprop functions.
For example, we have segment_mean_bp that returns 2 "gradients" (including one for the indices).
It would make a lot more sense for it to just return gradients for the input - i.e., single output function.
[...]

PR #7452, opened 2019-04-05
[WIP] SameDiff losses/differentiation overhaul

AlexDBlack:
Allows users to specify which variables represent the losses. Also differentiates only the subset of the graph that is actually needed (more efficient). Consequently, it's now possible to train only the subset of the graph that the loss function depends on.

Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/6876
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/7424 (though not in the proposed way)

Some other major changes:
* VARIABLE type SDVariables can now only be floating point. Variable type means "trainable" but only FP are trainable by backprop, by definition.


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -1427,6 +1429,52 @@ public long numElements() {
...
 
+    /**
+     * Get the names of variables (if any) that have been marked as loss variables to be minimized.<br>
+     * Variables can be marked as loss variables in a few different ways:<br>
+     * (a) Losses are automatically added when creating loss functions via {@link #sd()}<br>
+     * (b) Via {@link #setLossVariables(String...)}, @link #addLossVariable(String)} or {@link SDVariable#markAsLoss()}<br>
+     * (c) Via {@link TrainingConfig#setLossVariables(List)}<br>
+     */
+    public List<String> getLossVariables(){
+        return this.lossVariables;
```
treo: do we want to return a mutable version of the list here?

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -3122,70 +3356,150 @@ public void createGradFunction() {
...
-                // available before we try to differentiate the op
-                Queue<DifferentialFunction> availableForDiff = new LinkedList<>();
-                Set<String> seenOps = new HashSet<>();
-                //start with scalar backprop
-                INDArray initGradArr;
-                try(MemoryWorkspace ws = Nd4j.getMemoryManager().scopeOutOfWorkspaces()) {
-                    initGradArr = Nd4j.scalar(1.0);
+
+                //----- Step 1: Determine FP variables connected to loss -----
+                // Find all FP variables that are connected to loss by an FP32 path
```
treo: Should this actually be only an fp32 path?

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -3198,48 +3512,120 @@ public void createGradFunction() {
...
+                            //If no gradient exists at this point, 3 possibilities:
+                            // (a) we have a bug
+                            // (b) output of this op isn't used in calculating the loss
+                            // (c) output isn't a FP type
+                            //In the FP case, we should create a zero variable to backprop, because we can't perform backprop
+                            // for this op otherwise...
+                            if(!v.dataType().isFPType()){
+                                grads.add(null);
+                            } else {
+                                SDVariable gTemp = sameDiff.zerosLike(v);
```
treo: Should this even happen? Calculating with a zero gradient, should basically be "don't do any change at all" and therefore we shouldn't be performing backprop at all in that case?

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -3198,48 +3512,120 @@ public void createGradFunction() {
...
+                            }
+                            if(anyInputsRequired){
+                                if(!differentiatedOps.contains(op.getName())){
+                                    isRequiredOp = true;
+                                }
+                            }
+                        }
+
+                        if(!isRequiredOp){
+                            System.out.println("Skipped as not required: " + opName);
```
treo: Should probably be a `log.info` or `log.trace`?

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -3198,48 +3512,120 @@ public void createGradFunction() {
...
+                                }
+                                //If in't not in the minimal subgraph, loss doesn't depend on it, so we don't care about it
+                            }
                         }
 
-                        if(allAvailable && !seenOps.contains(o.getOp().getOwnName())) {
+                        if(allAvailable && !availableForDiff.contains(o.getOp())){
                             availableForDiff.add(o.getOp());
-                            seenOps.add(o.getOp().getOwnName());
+                            System.out.println("Marked available for diff: " + o.getOp().getOwnName());
```
treo: again, use logger?

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -3198,48 +3512,120 @@ public void createGradFunction() {
...
+                        } else {
+                            if(!availableForDiff.contains(o.getOp())) {
+                                System.out.println("Not all inputs available: " + o.getName());
```
treo: again, use logger?

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/TrainingConfig.java
```diff
@@ -348,17 +357,23 @@ public Builder skipBuilderValidation(boolean skip){
...
             if(!skipValidation) {
                 Preconditions.checkState(updater != null, "Updater (optimizer) must not be null. Use updater(IUpdater) to set an updater");
                 Preconditions.checkState(dataSetFeatureMapping != null, "No DataSet feature mapping has been provided. A " +
-                        "mapping between DataSet array positions and variables/placeholders must be provided - use  dateSetFeatureMapping(...) to set this");
-                Preconditions.checkState(dataSetLabelMapping != null, "No DataSet label mapping has been provided. A " +
-                        "mapping between DataSet array positions and variables/placeholders must be provided - use  dateSetLabelMapping(...) to set this");
+                        "mapping between DataSet array positions and variables/placeholders must be provided - use dateSetFeatureMapping(...) to set this");
+                Preconditions.checkState(markLabelsUnused || dataSetLabelMapping != null, "No DataSet label mapping has been provided. A " +
+                        "mapping between DataSet array positions and variables/placeholders must be provided - use dateSetLabelMapping(...) to set this," +
```
treo: typo: dateSetFeatureMapping and dateSetLabelMapping should be dataSetFeatureMapping and dataSetLabelMapping

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
AlexDBlack: No, any FP is fine. Good catch.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
AlexDBlack: Probably not.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
AlexDBlack: This is a little unintuitive, but it's definitely intended (and required).
Consider the following graph:
```
input -> split -> (A,B)
B -> lossFn
```
Here, A is unused (in that it doesn't contribute to the loss function). But to do split op backprop, we need gradient variables/arrays for both outputs (A and B).
We know the shape and type of dL/dA must be exactly the same as A; we also know that the loss function doesn't depend on A. Hence, dL/dA == zerosLike(A)

treo: got it. Maybe add the explanation to the comment on top?

AlexDBlack: Done.

PR #7460, opened 2019-04-06
Conversion tool protobuffer -> flatbuffer

yquemener:

Added a `nd4j-tools` module that contains a small tool able to convert models from the protocol buffer format to the flat buffer format. Also includes a BERT pre-processor that removes the operations that prevents it from running with libnd4j. (As is done in BERTGraphTest)

This is useful in order to execute models with only the libnd4j C++ API, which can't import protocol buffer models.

Note that I did not add `nd4j-tools` in `nd4j/pom.xml`. That can be a totally optional build.

## How was this patch tested?

I converted a densenet model and ran it through a libnd4j program, verified it produces the desired results.


nd4j/nd4j-tools/src/main/java/org/nd4j/cli/TFtoFlatFileConverter.java
```diff
@@ -0,0 +1,148 @@
...
+import org.nd4j.linalg.api.buffer.DataType;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+
+/**
+ * Created by Yves Quemener on 12/7/18.
+ */
+public class TFtoFlatFileConverter {
```
AlexDBlack: Need at least basic javadoc - on class and both methods.

nd4j/nd4j-tools/src/main/java/org/nd4j/cli/TFtoFlatFileConverter.java
```diff
@@ -0,0 +1,148 @@
...
+
+
+        System.out.println("Exporting file " + outFile);
+        sd.asFlatFile(new File(outFile));
+    }
+
+
+    @Test
+    @Ignore
+    public void convertFile() {
```
AlexDBlack: If we're going to have a test, let's put it in nd4j-tests - and not have it ignored.

nd4j/nd4j-tools/src/main/java/org/nd4j/cli/TFtoFlatFileConverter.java
yquemener: Done

nd4j/nd4j-tools/src/main/java/org/nd4j/cli/TFtoFlatFileConverter.java
yquemener: Done

yquemener:
And I clicked the "update branch" too hastily, thinking it would pull my latest commit, not masters' one. Hopefully it does not seem to have broken anything.

nd4j/nd4j-tensorflow/src/main/java/org/nd4j/tensorflow/conversion/ProtoBufToFlatBufConversion.java
```diff
@@ -0,0 +1,160 @@
+package org.nd4j.tensorflow.conversion;
```
saudet: Only only last thing I think, we'd need a copyright header from Skymind at the top here.

yquemener: Done

PR #7468, opened 2019-04-08
Test profiles

raver119:
This PR changes how we treat tests for backends. Instead of shared tests source folder, we'll use profiles. We need this due to:
- Intellij Idea has issues parsing shared tests sources, and they just get worse over time
- We'll finally get back platform specific tests for thin mechanics where required.

PR TL/DR:
- backend profiles added to nd4j-tests
- legacy platform-specific tests removed (new to be added where required)


raver119:
P.s. now, instead of `cd nd4j-cuda && mvn test` to run tests for CUDA  or `cd nd4j-native && mvn test` for CPU, we'll do `cd nd4j-tests && mvn test -P nd4j-tests-cuda`

saudet:
That sounds OK, but a lot of files have been removed, not moved, why?

raver119:
These files were actual in 2015, and never used since shared tests sources were applied (at the same 2015). So instead of trying to chew through legacy tests, we'll add tests that do make sense in 2019

saudet:
Didn't try to build it, so I'd let the CI server give it a try first, but LGTM.

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml
```diff
@@ -33,12 +33,14 @@
...
 
     <build>
+        <!--
         <testSourceDirectory>../../nd4j-tests/src/test/java</testSourceDirectory>
         <testResources>
             <testResource>
                 <directory>${project.basedir}/../../nd4j-tests/src/test/resources</directory>
             </testResource>
         </testResources>
+        -->
```
AlexDBlack: Let's remove this block entirely (not just comment out) before merging.

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
```diff
@@ -138,8 +138,9 @@
     </dependencies>
 
     <build>
+        <!--
         <testSourceDirectory>../../nd4j-tests/src/test/java</testSourceDirectory>
-
+        -->
```
AlexDBlack: Same here, remove not comment.

nd4j/nd4j-backends/nd4j-backend-impls/pom.xml
```diff
@@ -56,7 +56,9 @@
 
 
     <build>
+        <!--
         <testSourceDirectory>../nd4j-tests/src/test/java</testSourceDirectory>
+        -->
```
AlexDBlack: And here

saudet:
@sshepel We'll need to add those profiles when running tests on the CI server, it's not currently able to run the tests, see this for example: https://jenkins.ci.skymind.io/blue/organizations/jenkins/deeplearing4j%2Fdeeplearning4j/detail/PR-7539/1/pipeline

PR #7505, opened 2019-04-10
GraphExecutor standalone tool

yquemener:

This adds an option (deactivated by default) to generate a standalone tool as part of libnd4j. This tool allows to run an image classifier model on an image from the command line. It allows to display several information about the graph and automatically guess its input and output layers. It prints classification results as human readable labels.

The only modification on libnd4j is an additional parameter to deactivate warming up iterations during profiling (with the previous behavior being the default value of the parameter)

## How was this patch tested?

Libnd4j test suite passes 100%.
GraphExecutor was manually tested.


yquemener:
Another tool developed during the latest project, this one is a standalone libnd4j tool. I think it can be useful in cases where we need a quick tool to check the output of models or to benchmark inference.

saudet:
@yquemener To avoid various issues, make it download libpng at build time like I've recently done for Google Test and FlatBuffers, see pull requests #7469 and #7487.

yquemener:
Got it. I was wondering how to make it work under windows too. That's probably the easiest way. Thanks.

yquemener:
Actually the easiest way was to replace the third parties with a `find_package(...)` and make `libpng` a dependency, which seems reasonable. On Unix/Windows there is no need to link the lib statically.

There is only one third party remaining. A single file in `/thirdparty/klib` that provides `getopt` functions to non-POSIX OS (namely, windows). It is part of KLib (https://github.com/attractivechaos/klib
) which is distributed under MIT/X11. If that's still an issue, I can make it be downloaded.

saudet:
We can require users to install libpng by themselves, that's fine, as long as it's not part of the default build. Make sure that they have to enable it somehow manually. Same thing for klib. Let's not include the source in this repository when we can already download the same file from another repository!

yquemener:
Done. KLib is now downloaded during CMake's configuration. The standalone tool is still disabled by default, you need to switch them on in libnd4j/CMakeLists.txt for them to be built, on this line:

`option(BUILD_STANDALONE "Build standalone GraphExecutor" OFF)`

libnd4j/standalone/GraphExecutor.cpp
```diff
@@ -0,0 +1,548 @@
+#include "LabelsImagenet.h"
```
saudet: We'll need a copyright header here.

libnd4j/standalone/LabelsImagenet.h
```diff
@@ -0,0 +1,1014 @@
+#ifndef LABELS_IMAGENET_H
```
saudet: And here as well. Although I'm not sure we can claim copyright on the list of labels. Let's just load it from a plain text file given by the user? It also makes the tool more useful that way.

yquemener: I think this tool would be useless if we did not provide the labels somehow so I dug a bit more in details. When I included it first, I checked quickly and saw that many people included that in various project, as if it was public domain, slapped several permissible licenses on their own reformatted list.

It turns out that it comes originally from ImageNet's Large Scale Visual Recognition Challenge 2014 (ILSVRC2014), who does not state a license explicitly but says in its rules (http://www.image-net.org/challenges/LSVRC/2014/index#cite)
"If you are reporting results of the challenge or using the dataset, please cite:

    Olga Russakovsky*, Jia Deng*, Hao Su, Jonathan Krause, Sanjeev Satheesh, Sean Ma, Zhiheng Huang, Andrej Karpathy, Aditya Khosla, Michael Bernstein, Alexander C. Berg and Li Fei-Fei. (* = equal contribution) ImageNet Large Scale Visual Recognition Challenge. IJCV, 2015."

Which is this paper: https://arxiv.org/abs/1409.0575, which contains a list of all the categories used. I could not find any other licensing information about the ILSVRC2014 dataset. 

But wait, there is more. What the ILSVRC2014 actually gives is not a list of strings. It is a list of references to WordNet synset (a set of synonyms). So the actual strings come from WordNet, which is under this license (https://wordnet.princeton.edu/license-and-commercial-use) that both wikipedia and the FSF classify as a form of BSD 2-clauses licenses (which on cursory glance seems correct)

Technically I would argue that this file is more documentation/reference than code. It should probably be distributed alongside the models.

libnd4j/include/graph/profiling/GraphProfilingHelper.h
```diff
@@ -29,7 +29,7 @@ namespace nd4j {
     namespace graph {
         class GraphProfilingHelper {
         public:
-            static GraphProfile* profile(Graph *graph, int iterations);
+            static GraphProfile* profile(Graph *graph, int iterations,int warmUpIterations);
```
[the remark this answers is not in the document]
yquemener: Ah yes, my bad, I added it in the `.cpp` file. Fixed.

libnd4j/standalone/GraphExecutor.cpp
yquemener: Done

libnd4j/standalone/LabelsImagenet.h
yquemener: Added a header for now. I can't work on it today but I'll write a loader later. Added on the todo list.

saudet: Yes, it's specific to each model, so that's why we should just have the tool load it from a plain text file anyway. It's available, for example, as synset_words.txt inside http://dl.caffe.berkeleyvision.org/caffe_ilsvrc12.tar.gz . I think we have it online on a server from Skymind somewhere too (@AlexDBlack ? @raver119 ?) because that way we don't have to go through Eclipse's IP team.

saudet: @yquemener The list is available here:
https://github.com/deeplearning4j/dl4j-test-resources/blob/master/src/main/resources/tf_graphs/zoo_models/labels/imagenet_labellist.txt
So use that file for your tests.

PR #7567, opened 2019-04-17
[WIP] OpContext and RNG

raver119:
This PR:
- propagates Nd4j random into custom ops executioner, to guarantee rng reproducibility
- changes custom op executioner in order to use OpContext instead of pointers


raver119:
cc @AlexDBlack

Issue #7503, opened 2019-04-10
DL4J/SameDiff: Validation and UX for model import issues


AlexDBlack:
Users will be loading models and data from a variety of formats in DL4J/ND4J:
* DL4J: ComputationGraph, MultiLayerNetwork
* DL4J Keras import: functional and sequential models
* SameDiff: FlatBuffers, Zip format with FB + training config/updaters
* SameDiff: Tensorflow models
* ND4J: INDArrays
* DL4J: NLP models

We want to improve UX for this. Specifically, we want to:
1. Validate models to pre-empt possible failures (bad format, corrupted model)
2. Provide useful exceptions (tells user what happened, why, and when possible some solutions/next steps)
3. Where feasible, allow users to work around some types of problems

Ways model loading can fail:
1. File corruption (bad model file)
2. Loading model from later version into earlier version (create in beta4, load in beta3)
3. Attempting to load via wrong method: i.e., try to load Keras model via DL4J ModelSerializer
4. Missing ops or layers (DL4J Keras import, SameDiff TF import)
5. Newer external library version (Keras/TF etc format changes)

Some initial thoughts on how to improve usability:
1. Add utilities to determine if a model is (probably) valid - instead of relying on a try/fail/exception type procedure
2. Add version info to all model files/formats we create (to detect loading, for example, beta4 model in beta3 and warning about probable failure)
3. In the case of DL4J-Keras and SameDiff-TF, add a way to get model status (we have a start on something like that for SameDiff/TF [here](https://github.com/deeplearning4j/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/imports/tensorflow/TensorFlowImportValidator.java))

wmeddie:
I would like to add that it would be great if all these validations would work without having to load the weights into memory at all.

AlexDBlack:
Merged to current dev branch - should be merged to master in about a week.
https://github.com/deeplearning4j/deeplearning4j/pull/7701

PR #7643, opened 2019-05-01
[WIP] Model file validation

AlexDBlack:
Fixes: https://github.com/deeplearning4j/deeplearning4j/issues/7503

Lots more left to do.
Basic JSON/ZIP validation implemented; DL4J models next.


nd4j/nd4j-common/src/main/java/org/nd4j/validation/Nd4jCommonValidator.java
```diff
@@ -0,0 +1,218 @@
...
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+public class Nd4jCommonValidator {
+
+    private Nd4jCommonValidator(){ }
```
firefox7025: Total personal preference, but just curious, any reason we don't adapt lombok in the code base?

AlexDBlack: We use Lombok extensively. But the equivalent annotation is actually longer in this particular case :)
`@NoArgsConstructor(access = AccessLevel.PRIVATE)`

AlexDBlack:
Merged to dev branch here: https://github.com/deeplearning4j/deeplearning4j/pull/7701

PR #7731, opened 2019-05-14
partially added 16-bit tiff support

BioinfoTongLI:

To be able to load 16-bit multipage tiff file into INDArray

## How was this patch tested?

unit test with data in testresource

## Quick checklist


datavec/datavec-data/datavec-data-image/src/test/java/org/datavec/image/loader/TestNativeImageLoader.java
```diff
@@ -175,6 +176,22 @@ public void testConvertPix() throws Exception {
         assertEquals(depth1, array9.size(2));
         assertEquals(h5, array9.size(3));
         assertEquals(w5, array9.size(4));
+
```
BioinfoTongLI: This test is not passing. Simplely because I don't know how would you describe an image with 5 channels in INDArray.

datavec/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -824,36 +838,37 @@ private INDArray asMatrix(BytePointer bytes, long length) throws IOException {
         INDArrayIndex[] index = null;
         switch (this.multiPageMode) {
             case MINIBATCH:
-                data = Nd4j.create(pixa.n(), 1, pixa.pix(0).h(), pixa.pix(0).w());
+                data = Nd4j.create(pixa.n(), 1, 1, pixa.pix(0).h(), pixa.pix(0).w());
```
saudet: We want to set the depth and the number of channels here instead of just "1, 1"

BioinfoTongLI: Did I miss somethings ? What do you mean depth here ? Depth of image of depth of 3d volume ?

BioinfoTongLI: I believe you mean depth of 3D volume. But in that case, I can not guess the structure of data. I need more metadata information to be able to correctly load the image. For example, what are the numbers of channel/depth/batch?  How are they arranged ? If you need data at this complexity. I recommend you to wrap this project [tifffile](https://pypi.org/project/tifffile/) into Datavec. I saw you began to work with numpy. It won't be much difficult, if you are already able to communicate numpy with INDArray

BioinfoTongLI: Or this Java project ([bioformat](https://www.openmicroscopy.org/bio-formats/)), if you don't want Python.

saudet: It's not about NDArray, it's about returning the data that is in a TIFF file, but you're right, `pixa.n()` is often basically the "depth",  so we'd need another DEPTH mode would be a good name I think:
```
Nd4j.create(1, channels, pixa.n(), pixa.pix(0).h(), pixa.pix(0).w());
```
But sometimes we want to return it as a batch though, so for MINIBATCH mode we'd have:
```
Nd4j.create(pixa.n(), channels, 1, pixa.pix(0).h(), pixa.pix(0).w());
```
For the CHANNELS mode, I guess something like this:
```
Nd4j.create(1, pixa.n(), 1, pixa.pix(0).h(), pixa.pix(0).w());
```
But the channels information goes away and this is strange... What about supporting only the first one above, the one in NCDHW order, and leave it up to the user to manipulate the output further if they need it?

saudet: Sure, we can use other dependencies than Leptonica if something else is easier to work with.

BioinfoTongLI: Sincerely, I personally don't think hard-coding this part is a good idea. It seems that you started to have biomedical image users. In this area, your 5 channels is still not enough. You will need the time dimension. Generally, we (at least me) work with by default 5-channel images (TimeChannelZXY). For the training, you need anther N. That become really complex from my point of view. What do you think @AlexDBlack ?

BioinfoTongLI: @saudet I believe for instance its better to let user permute their data axis to get what they want, because integrating metadata to get exactly what they want is **painful**...

saudet: I don't think TIFF files contain other axis like time, right? If so let's not worry about that. We're only interested in loading data image files here.

BioinfoTongLI: It depends on what tiff you want to manipulate. 

So, yes. In this case, In fact, I would remove the multipage option. Every image will by default be loaded in `batch` channel. It up to user to reorder the image. Is this what you wanted ?

saudet: Sounds reasonable, yes. Let's just make sure images with more than 1 channel can be loaded.

datavec/datavec-data/datavec-data-image/src/test/java/org/datavec/image/loader/TestNativeImageLoader.java
BioinfoTongLI: Channel option removed, this is no longer an issue. Closing this.

BioinfoTongLI:
Sorry for being laggy for this small changes. 
I'm preparing my thesis defense. Kind of busy.
BTW, I saw the tests are not passing. I believe it's due to the PR in [testresource](https://github.com/deeplearning4j/dl4j-test-resources/pull/194).
As for why that is not passing. I have no idea... Any clue?

AlexDBlack:
@BioinfoTongLI I'm seeing all tests pass in datavec-data-image on your branch when tested locally (using latest master for dl4j-test-resources and mono repo as base)... including your new tests.
That's good enough for me... don't worry about any other failures not related to your changes.
So this is good to go then?

BioinfoTongLI:
I saw that too. Wired.... But Yeah, ok for me to go.

datavec/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
saudet: This still doesn't take into account the number of channels in the image. We need 
to put the `channels` variable somewhere here and below as well.

BioinfoTongLI:
I don't get it.
If I take the `channels` into account. It will be 
`data = Nd4j.create(pixa.n()/n_ch, n_ch, 1, pixa.pix(0).h(), pixa.pix(0).w());`, right?
And for images with more than 1 channel. How can I know its`[c1, c1, c1, c1, c2, c2, c2, c2]` 
or `[c1, c2, c1, c2, c1, c2, c1, c2]`? Tell me if I miss something.

saudet:
I'm pretty sure the number of channels and the number of pages are not related to each other... Anyway, if nobody complains about this, let's just merge it as is I suppose.

PR #7733, opened 2019-05-14
Scalnet: fix basic rnn example

maxpumperla:
Removed GravesLSTM and failing rnn time step usage (not needed for checking validity here).


AlexDBlack:
Is this good to merge? Or is there more left to do?

maxpumperla:
@maxpumperla was good to go. misunderstanding, I thought @sshepel would merge. ;)

PR #7742, opened 2019-05-15
Modify regex for scheme detection

jamesoliverh:
Modified regex expression for scheme detection from
`if(baseString.matches(".*:/.*")){`
to
`if(baseString.matches(".{2,}:/.*")){`
i.e. there must be at least 2 characters before the ":/".

This prevents the string
"C:/path/" 
from being matched (contrary to before), whilst matching e.g.
"file:/C:/path" and "file:/path".

Additionally, with the proposed modification the string ":/path" which matched previously no longer matches (which I assume is desired behaviour).


jamesoliverh:
@AlexDBlack Any pointers on why some stuff might be failing in the continuous integration tests? Apologies, I am quite new to this.

PR #7836, opened 2019-06-03
FastText wrapper

alexanderst:

FastText  wrapper for DL4J-NLP module.

## How was this patch tested?

Unit tests.


deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/pom.xml
```diff
@@ -48,6 +48,12 @@
         <version>${threadly.version}</version>
     </dependency>
 
+    <dependency>
```
saudet: Also try to force a dependency on the same version of JavaCPP as used in the rest of Deeplearning4j.

PR #7838, opened 2019-06-03
[WIP] Added implementation of FastText wrapper

alexanderst:

FastText model for DL4J-N LP

## How was this patch tested?

Unit tests


AlexDBlack:
Looks good so far - but one minor point, let's move the test resources to the test resources repo (especially if they are 3rd party content, this matters for Eclipse).

Also don't write directly to test resources directory like this.
`outputFile("src/test/resources/models/fasttext/supervised.model")`
It's too easy for the output to get added to the repo inadvertantly.

Instead, use JUnit's TemporaryFolder functionality - here's an example:
https://github.com/deeplearning4j/deeplearning4j/blob/c85d9629da93fa5b8c54fc632f890066bf742aa4/nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/samediff/listeners/CheckpointListenerTest.java#L41-L42
plus `File dir = testDir.newFolder();`
Use that as the output location instead.

deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/pom.xml
```diff
@@ -86,6 +86,12 @@
           <artifactId>commons-lang3</artifactId>
           <version>${commonslang.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.github.vinhkhuc</groupId>
+        <artifactId>jfasttext</artifactId>
+        <version>0.4</version>
+      </dependency>
```
saudet: Also try to force a dependency on the same version of JavaCPP as used in the rest of Deeplearning4j.

raver119: At this moment that's not our code, and as I see in maven central - 0.4 is the only available version at this moment.

saudet: No, but newer versions of ND4J will not work with an old version of JavaCPP, while an old version of JFastText *might* work with a new version of JavaCPP.

PR #7915, opened 2019-06-18
Add thousand-separator commas to TotalParams

jxtps:

The number of parameters can be quite large, and it would help the reading of the summary printout to have the TotalParams column & values at the bottom have thousand-separator-commas in them. This PR adds that.

## How was this patch tested?

Sorry, no testing, edits are tiny and hopefully clear. Didn't run mvn formatter since all changes were inline.

Hopefully the signoff worked...


saudet:
@AlexDBlack It looks like committers can still squash and merge though...

AlexDBlack:
@jxtps Thanks for the PR
We'll have docs up on Eclipse IP process here soon, we're still working out the process here given we've only just done the transition.

PR #7941, opened 2019-06-24
[WIP] RL4J Cleanup / Refac

aboulang2002:
WORK IN PROGRESS

RL4J could be a fantastic library. In it's current state however, it looks like an elaborate proof of concept. Ruben Fiszel has done a pretty good job trying to reproduce the implementation described in "Playing Atari with Deep Reinforcement Learning". It's always impressive to see what deep learning can achieve.
However, using RL4J or understanding what the code does is very difficult. For example QLearningDiscrete.trainStep() deals with a lot of different concerns (frame skipping, history processor, scaling, experience replay, ...). The mixing of concerns throughout RL4J makes it impossible to understand what and how it does reinforcement learning. It also makes it impossible to change it's behavior to something it wasn't explicitly designed for.

With this PR, i'd like to refac RL4J. Any comment along the way is very welcome.

## Quick checklist

- [ ] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.org/eclipse-contributors) page for details
- [ ] Reviewed the [Contributing Guidelines](https://github.com/eclipse/deeplearning4j/blob/master/CONTRIBUTING.md) and followed the steps within.


aboulang2002:
HistoryProcessor is very GYM oriented and unfortunately very deeply bound to RL4J classes. The frame-skipping and history recording/adding is not required in all use cases and obscures the logic. It is more of an 'observation' concern than a learning or policy concern and so, does not belong in those classes.

My first commit (
d5bcbfa) attempts to reduce the dependency to HistoryProcessor. I temporarily introduce two classes and an interface that does the initialization of the MDP and removes most of the HistoryProcessor dependency from the base Learning classs.

My second commit (
5c959d8) removes the hard dependency to DataManager and replaces it with a listener pattern.

aboulang2002:
Commit 2843259

Ideally, Policies and Learning classes shouldn't have to transform observations; they should get them in their usable form. I added a namespace, observation, that will deal with taking raw MDP observations and change them into what the configured policy and learning want them. To transform the raw observation, one assemble a serie of simple transform that can be easily understood, easily changed and easily tested. Used like this:

```
        transform = PipelineTransform.builder()
            .flowTo(SignalingTransform.builder().listener(new RecorderTransformListener(hp)).build())
            .flowTo(SkippingTransform.builder().skipFrame(4).build())
            .flowTo(PoolingTransform.builder().observablePool(observablePool).build())
            .build();
```

Policy now uses an ObservationTransform but still does the transform. To be fixed later.

Breaking changes in Policy:
 - Policy.play did not historyProcessor.record the last observation. Now every observation is passed to the transform.

aboulang2002:
There is a detail implementation that I disagree with the original implementation. The original Policy.java does something like this:
```
 while (!mdp.isDone()) {

			[...]
 
            StepReply<O> stepReply = mdp.step(action);
            reward += stepReply.getReward();

            if (isHistoryProcessor)
                hp.add(stepReply.getObservation().toNDArray());

            history = isHistoryProcessor
                    ? hp.getHistory()
                    : new INDArray[] { stepReply.getObservation().toNDArray() };
            ++step;
        }	

```
HistoryProcessor.add() is called for every observation which is not my interpretation of the paper:

> "Following previous approaches to playing Atari games, we also use a simple frame-skipping technique [3]. More precisely, the agent sees and selects actions on every k-th frame instead of every frame, and its last action is repeated on skipped frames."

My understanding is that 3 out 4 frames is completely ignored. I.e. the policy/learning just keep doing the previous action and does not use or keep the skipped observation in any way.

I searched other implementations of this paper on internet and found that some add the skipped observation and some don't. Then I found a very interesting article discussing this very point (see https://danieltakeshi.github.io/2016/11/25/frame-skipping-and-preprocessing-for-deep-q-networks-on-atari-2600-games/). 

> "Seeing four consecutive frames without subsampling doesn’t give enough information to discern motion that well, especially after the downsampling process."

Therefore I propose to change this behavior to: "skipped frames are not added to the history". Of course, with an ObservationTransform pipeline, one can use the behavior he wishes.

saudet:
It's probably not the only thing wrong with that function in Policy.java. I'm not entirely sure what it's really doing... @rubenfiszel?

saudet:
BTW, all classes from gym-java-client should be moved under the rl4j-core module, or the rl4j-gym one for anything specific to Python. rl4j-core shouldn't depend on anything related to OpenAI Gym. It makes sense for Python frameworks, but it doesn't make sense for RL4J to depend on Python.

Also, let's try to limit the amount of refactoring to as small (ideally independent) increments as possible, with one PR for each increment. Sounds good? Thanks a lot for your time on this!

rubenfiszel:
Hello,

I will review this evening. Thanks for the contribution.

aboulang2002:
@saudet I will close this PR and submit smaller ones instead.

PR #7945, opened 2019-06-26
Merge master to upstream

AlexDBlack:


aeiplatform:
Why did you make methods getTrainIterator() and getTestIterator() in class DataSetIteratorSplitter deprecated?
Is there any equivalent substitution to have same behavior of pseudo example:

`graph.fit(data.getTrainIterator(), EPOCH);`
`graph.evaluate(data.getTestIterator());`

raver119:
Because now you can have any number of iterators, not only 2.

aeiplatform:
If anybody uses non deprecated constructor DataSetIteratorSplitter(@NonNull DataSetIterator baseIterator, long totalBatches, double ratio) then that method will cause NPE.

aeiplatform:
Same defect in MultiDataSetIteratorSplitter.

PR #7949, opened 2019-06-27
SameDiff variables documentation

KolyaIvankov:
Signed-off-by: Nikolay Ivankov <kolya.ivankov@gmail.com>
Basically the stuff on variables already discussed with @AlexDBlack Alex, added some information on methods `getVariable`, `eval` and `getGradient` at the end of the document. 
Hopefully the references will mostly work.

`overview.md` and `ops.md` are on their way.


docs/samediff/templates/variables.md
```diff
@@ -1,3 +1,214 @@
...
+title: Types of variables in SameDiff
+short_title: Variables
+description: What types of variables are used in SameDiff, their properties and how to switch these types.
+category: SameDiff
+weight: 3
+---
+
+# Variables in `SameDiff`
+
+##What are variables
```
AlexDBlack: Minor issue: Looks like we need a space after `##` for it to render properly in markdown... https://github.com/eclipse/deeplearning4j/blob/871d5897f810953160a10dfdaeb4d0913c16a5f1/docs/samediff/templates/variables.md

docs/samediff/templates/variables.md
```diff
@@ -1,3 +1,214 @@
...
+samediff.convertToConstant(placeHolder);
+```
+For now it is not possible to turn a constant back into a placeholder, we may consider adding this functionality if 
+there is a need for that. For now, if you wish to effectively freeze your placeholder but be able to use it again, 
+consider supplying it with constant values rather than turning it into a constant.
+
+##Variables' names and values
+### Getting variables from `SameDiff`
+Recall that every variable in an instance of `SameDiff` has its unique `String` name. Your `SameDiff` actually tracks your 
+variables by their names, and allows you to retrieve them by using `getVaraible(String name)` method.
```
AlexDBlack: Typo: getVaraible

KolyaIvankov:
Added changes, hopefully this will do.

PR #8011, opened 2019-07-15
Use a semaphore instead of busy-ish wait

jxtps:

Switch from using `LockSupport.parkNanos(1000)` to `semaphore.acquire()`

## How was this patch tested?

Terribly sorry, no testing, don't have code checked out locally, editing on github.com - your CI system runs the tests, no!?

### DIFF
--- a/deeplearning4j/deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/inference/observers/BasicInferenceObserver.java
+++ b/deeplearning4j/deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/inference/observers/BasicInferenceObserver.java
@@ -20,8 +20,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import java.util.Observable;
 import java.util.Observer;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.LockSupport;
+import java.util.concurrent.Semaphore;
 
 /**
  * Simple Observer implementation for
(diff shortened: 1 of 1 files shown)


jxtps:
In reference to https://github.com/eclipse/deeplearning4j/issues/7980

raver119:
Please read: https://accounts.eclipse.org/legal/eca/validation/8993

TLDR: you need to accept Eclipse Contributor Agreement

jxtps:
It's the signed off by footer. I can't get it to work on github.com. Sorry. Please override it. I've accepted the contributor agreement.

deeplearning4j/deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/inference/observers/BasicInferenceObserver.java
```diff
@@ -31,23 +30,24 @@
...
+    private Semaphore semaphore = new Semaphore(0);
+    volatile private boolean finished = false;
 
     @Override
-    public void update(Observable o, Object arg) {
-        finished.set(true);
+    synchronized public void update(Observable o, Object arg) {
+        if (!finished) {
+            finished = true;
+            semaphore.release(Integer.MAX_VALUE);
```
AlexDBlack: This doesn't look safe to me.
If the semaphore available permits is anything other than 0, this will overflow.

Semaphore.release:
```
    public void release(int permits) {
        if (permits < 0) throw new IllegalArgumentException();
        sync.releaseShared(permits);
    }
```

This ultimately ends up calling this method, on Semaphore.Sync:
```
        protected final boolean tryReleaseShared(int releases) {
            for (;;) {
                int current = getState();
                int next = current + releases;
                if (next < current) // overflow
                    throw new Error("Maximum permit count exceeded");
                if (compareAndSetState(current, next))
                    return true;
            }
        }
```

Unless we can guarantee that the semaphore only ever has 0 permits at this point (unlikely), this is a bug.

AlexDBlack (changes requested):
See comment.

deeplearning4j/deeplearning4j-scaleout/deeplearning4j-scaleout-parallelwrapper/src/main/java/org/deeplearning4j/parallelism/inference/observers/BasicInferenceObserver.java
jxtps: You're right, did some reading and this use-case actually calls for a [`CountDownLatch`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CountDownLatch.html), _A synchronization aid that allows one or more threads to wait until a set of operations being performed in other threads completes_.

```java
public class BasicInferenceObserver implements Observer {
    private CountDownLatch latch = new CountDownLatch(1);

    @Override
    public void update(Observable o, Object arg) {
        latch.countDown();
    }

    public void waitTillDone() {
        latch.await();
    }
}
```

Only caveat is that `latch.await()` can throw an `InterruptedException`, which arguably `waitTillDone()` should also do - wrapping it in a `RuntimeException` is poor form ;)

agibsonccc: Updates here?

PR #8036, opened 2019-07-22
libnd4j: Provide a C API

saudet:

libnd4j already exposes an interface that is mostly C in NativeOps.h, but it is still C++, making it less portable than it could be. Here we adjust the function declarations to export them as actual C functions. Tasks to be done:

- [x] Refactor NativeOps.h to export C functions
- [x] Adapt the Java wrappers in ND4J generated with JavaCPP
- [x] Create C wrappers for some of the C++ classes currently used by ND4J

## How was this patch tested?

All builds pass and existing unit tests are not broken.


raver119:
Looks great, but one question: why to eclipse directly and not to fork first?

saudet:
Merged to https://github.com/SkymindIO/deeplearning4j/tree/master

PR #8040, opened 2019-07-24
audio Feature extraction

wangfeng-skymind:
mfcc

### DIFF
--- a/datavec/datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/mfcc/AudioFeatures.java
+++ b/datavec/datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/mfcc/AudioFeatures.java
@@ -2,7 +2,6 @@ package org.datavec.audio.mfcc;
 
 
 import org.datavec.audio.dsp.FastFourierTransform;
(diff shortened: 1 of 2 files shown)


AlexDBlack (changes requested):
Apologies for not getting to review this sooner.

I have checked some of the math vs. the linked paper, earlier steps looked correct to me but I didn't go into depth for the later steps.

There are some issues/questions here:

First: This is all Java, single threaded.
Why not use ND4J for this? All of the math operations can be done in ND4J, with full parallelization, multiple datatype (float16/32/64, in the same implementation) and GPU support. ND4J is designed for exactly these sorts of things.

Second: No Javadoc on any of the methods.
The code comments you have on many methods would be a good start as javadoc. A few others are missing javadoc.

Third: No tests at all here. Given the complexity of the math, the probability of a mistake is high.
Without tests, how do we know your implementation is correct, and how can we make sure nobody else will break it in the future?

wangfeng-skymind:
modifying and adding other code and javadoc

saudet:
@wangfeng-skymind Also please make sure to "sign" Eclipse's ECA. Thanks!

agibsonccc:
Closing due to lack of activity.

PR #8043, opened 2019-07-24
SameDiff ops documentation

KolyaIvankov:
Added ops.md to SameDiff.


docs/samediff/templates/ops.md
```diff
@@ -1,3 +1,243 @@
...
+
+### `loss` - Loss functions
+In this branch we keep common loss functions. Most loss functions may be created quite simply, like that:
+```java
+SDVariable logLoss = sameDiff.loss.logLoss("logLoss", label, predictions);
+```
+where `labels` and `predictions` are `SDVariable`'s. A `String` name is a mandatory parameter in most `loss` methods, 
+yet it may be set to `null` - in this case, the name will be generated automatically. You may also create weighted loss
+functions by adding another `SDVariable` parameters containing weights, as well as specify a reduction method for the 
+loss over the minibatch. Thus, a full-fledged `logLoss` operation may
```
AlexDBlack: "reduction method" - let's basically copy paste the content from here to explain the reduction modes:
https://github.com/eclipse/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/loss/LossReduce.java

docs/samediff/templates/variables.md
```diff
@@ -1,3 +1,214 @@
...
+```
+adds a variable constituting of a 784x10 array of `float` numbers - weights for a single layer MNIST perceptron 
+in this case - to a pre-existing `SameDiff` instance `samediff`.
+However, this way the values within a variable will be set as zeros. If we wish them to be filled, say, with random 
+numbers, we need to add a weight initialization scheme, for instance like that:
+```java
+SDVariable weights = samediff.var("weights", new XavierInitScheme('c', 784, 10), DataType.FLOAT, 784, 10);
+```
+Now, the weights will be randomly initialized using the Xavier scheme. There are other ways to create and 
+fill variables, you may see a full reference in our [javadoc](https://deeplearning4j.org/api/latest/).
```
AlexDBlack: Can we link more directly?
"see the 'known subclasses' section of https://deeplearning4j.org/api/latest/org/nd4j/weightinit/WeightInitScheme.html"

Also add a sd.var(..., INDArray) example, with a manually (Nd4j.someMethod) created array.

PR #8050, opened 2019-07-26
RL4J Added listener pattern to SyncLearning

aboulang2002:
Signed-off-by: Alexandre Boulanger <aboulang2002@yahoo.com>

The goal of this PR is to remove IDataManager from SyncLearning. I added a listener pattern to SyncLearning, and a listener class (DataManagerSyncTrainingListener) that will make the calls to DataManager that SyncLearning used to do.

I removed the try/catch from SyncLearning.train(). Listeners should be well-behaved and not throw. If a listener need to stop the training (because of an IOException for example), it can with SyncTrainingEvent.setCanContinue(false).

Since SyncLearning doesn't use IDataManager anymore, I moved it to ASyncLearning (where a similar refac will be done in an upcoming PR).

I did not add javadoc to IDataManager and IAsyncGlobal.
In the case of IDataManager/DataManager, it may be worthwhile to refac it. For example, isSaveData() will soon not be necessary anymore. Also, it seems strange to me that a DataManager would return the path where to save the videos which are then saved by a HistoryProcessor. I will write the javadoc when I change this class.
In the case of IAsyncGlobal, I don't understand a few things and I'd like to take the time to read and understand "Hogwild!" before writing the javadoc.

Summary of changes:
 - Moved up IDataManager from Learning to ASyncLearning and removed it from SyncLearning
 - Some code in QLearningDiscrete.postEpoch() has been moved to DataManagerSyncTrainingListener.onEpochEnd()
 - Added listener pattern to SyncLearning along with a listener, DataManagerSyncTrainingListener.
 - Adjusted SyncLearningTest for the new listener
 - Added refac tests to QLearningDiscrete to make sure it remains functionally the same as before.

## How was this patch tested?

Unit tests

## Quick checklist

### DIFF
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/SyncLearning.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/SyncLearning.java
@@ -30,36 +30,58 @@ import java.util.ArrayList;
 import java.util.List;
 
 /**
- * @author rubenfiszel (ruben.fiszel@epfl.ch) on 8/3/16.
- * @author Alexandre Boulanger
- *
  * Mother class and useful factorisations for all training methods that
  * are not asynchronous.
  *
+ * @author rubenfiszel (ruben.fiszel@epfl.ch) on 8/3/16.
+ * @author Alexandre Boulanger
  */
 @Slf4j
 public abstract class SyncLearning<O extends Encodable, A, AS extends ActionSpace<A>, NN extends NeuralNet>
-                extends Learning<O, A, AS, NN> {
+        extends Learning<O, A, AS, NN> {
+
+    private List<SyncTrainingListener> listeners = new ArrayList<>();
 
     public SyncLearning(LConfiguration conf) {
         super(conf);
     }
 
-    private List<SyncTrainingListener> listeners = new ArrayList<>();
+    /**
+     * Add a listener at the end of the listener list.
+     *
+     * @param listener
+     */
     public void addListener(SyncTrainingListener listener) {
         listeners.add(listener);
     }
 
+    /**
+     * This method will train the model<p>
+     * The training stop when:<br>
+     * - the number of steps reaches the maximum defined in the configuration (see {@link LConfiguration#getMaxStep() LConfiguration.getMaxStep()})<br>
+     * OR<br>
+     * - a listener explicitly stops it<br>
+     * <p>
+     * Listeners<br>
+     * For a given event, the listeners are called sequentially in same the order as they were added. If one listener
+     * returns {@link SyncTrainingListener.ListenerResponse SyncTrainingListener.ListenerResponse.STOP}, the remaining listeners in the list won't be called.<br>
+     * Events:
+     * <ul>
+     *   <li>{@link SyncTrainingListener#onTrainingStart(SyncTrainingEvent) onTrainingStart()} is called once when the training starts.</li>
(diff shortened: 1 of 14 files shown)


rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/listener/SyncTrainingListener.java
```diff
@@ -0,0 +1,31 @@
+package org.deeplearning4j.rl4j.learning.sync.listener;
+
+import org.deeplearning4j.rl4j.util.IDataManager;
+
+/**
+ * A listener interface to use with a descendant of {@link org.deeplearning4j.rl4j.learning.sync.SyncLearning}
+ */
+public interface SyncTrainingListener {
+    /**
+     * Called once when the training starts. The training can be aborted by calling event.setCanContinue(false)
```
saudet: @AlexDBlack Are you OK with using a listener for that?

rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/qlearning/discrete/QLearningDiscrete.java
```diff
@@ -68,24 +66,29 @@
     private int lastAction;
     private INDArray history[] = null;
     private double accuReward = 0;
-    private int lastMonitor = -Constants.MONITOR_FREQ;
-
 
+    /**
+     * Obsolete. Use QLearningDiscrete(MDP, IDQN, QLConfiguration, int) with addListener() instead.
```
saudet: Let's mark those with `@Deprecated`.

rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/SyncLearning.java
```diff
@@ -35,50 +41,93 @@
...
+        listeners.add(listener);
+    }
 
-            getDataManager().writeInfo(this);
+    public void train() {
 
+        log.info("training starting.");
 
+        boolean canContinue = notifyTrainingStarted();
+        if(canContinue) {
```
saudet: The formatting looks off in this file. Let's be as consistent as possible. :)

rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/listener/SyncTrainingListener.java
AlexDBlack: Yes, we're actually looking at a similar design in SameDiff listeners.
Though we're going with something like onEpochEnd returning an enumeration - `ListenerResponse.STOP` or `ListenerResponse.CONTINUE` sort of thing.

rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/util/DataManagerSyncTrainingListener.java
```diff
@@ -0,0 +1,105 @@
...
+        this.lastMonitor = -builder.monitorFrequency;
+    }
+
+    @Override
+    public void onTrainingStart(SyncTrainingEvent event) {
+        try {
+            dataManager.writeInfo(event.getLearning());
+        } catch (Exception e) {
+            log.error("Training failed.", e);
+            e.printStackTrace();
```
AlexDBlack: Don't also print stack trace, log.error will already print it.
Also other instances below.

aboulang2002:
Summary of changes:
- Added/changed javadocs
- Make listeners return a response instead of changing a canContinue property (also changed SyncLearning and unit tests)
- removed all e.printStackTrace() in DataManagerSyncTrainingListener

saudet:
Looks good to me! Tested it out a bit, and it doesn't appear to break anything.

Issue #3022, opened 2017-03-12
OSGi compliant jars for containers such as Equinox or Felix. 


carldea:
There are many OSGi containers with bundle context class loaders capable of dynamically load services. It would be nice for DL4J and its dependencies to be OSGi friendly.  

Some helpful links:
http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html
http://blog.sonatype.com/2009/09/maven-tips-and-tricks-creating-an-osgi-project-with-maven

streetturtle:
Hi @carldea, did you manage to OSGIfy DL4J?

jckautzmann:
Hi @carldea, any update on this one?

agibsonccc:
Hey folks - we want to do this at some point as part of the eclipse foundation. We'd still accept contributions for this.

laeubi:
I have used ND4J successfully inside an OSGi-Container, I'll try to contribute a patch to add OSGi-Metadata.

tuf22191:
Hi, what is the status of this?

laeubi:
My current problem is, that it is not easy to run a full ND4J build so I can't test my contribution right now :-\

treo:
@laeubi feel free to join us in gitter and ask questions about it. If you've got all the prerequisites installed, it is as simple as `mvn clean install -pl '!pydatavec,!jumpy,!pydl4j,!nd4s' -Dmaven.test.skip=true`

laeubi:
@treo Thanks for the hint, with this command line I was now able to build the project. I still got lots of warnings from maven but the build completes with a success state.

timothyjward:
I will take a look at this, and feed back on options for packaging things so that they work reliably

damadux:
I have this issue as well, when trying to launch an Eclipse application. I suspect this has to do with multiple artifacts having the same package (e.g. org.nd4j.nativeblas being a package in org.nd4j.native-api and org.nd4j.native.linux-x86_64 in my case, causing a NoAvailableBackendException). However, keep in mind that the JUnit Test works perfectly. Here is the pom that I use to generate all dependencies, using Reficio's p2-maven-plugin (https://github.com/reficio/p2-maven-plugin) 

https://gist.github.com/damadux/a199ab600338ca6e62698dcad01f2ddf

timothyjward:
[...]

PR #8083, opened 2019-08-12
Create OSGi bundles for ND4J

timothyjward:

As described in #3022 it is necessary for Deeplearning4J to be provided as OSGi bundles in order for it to be used embedded in Eclipse. OSGi has a very different class loader structure which impacts how DL4J loads its backend implementation and its various plugins.

This PR makes the necessary additions/fixes to provide ND4J as OSGi bundles (the first step on the road to running the full DL4J on OSGi. Ideally the existing ND4J jars would have been enhanced with OSGi metadata, but this is not possible because:

- The ND4J API has split packages between its various modules, these cannot work in separate OSGi bundles
- The ND4J API is tightly coupled to JavaCPP, which further requires a flat class space to load native libraries

The PR therefore creates an nd4j-osgi bundle which wraps up the "frontend api" modules into a single OSGi bundle JAR. This has a requirement for a "backend". There is also an nd4j-native-osgi bundle providing the backend - this is an OSGi fragment which attaches to the frontend API to provide a working backend.

The overall approach is minimally invasive to the existing ND4J code, there are no API changes and all of the existing jars are still produced as before. 

## How was this patch tested?

There is an automated OSGi sniff test which has been put in the nd4j-tests-osgi module

## Quick checklist


saudet:
I don't see these changes are "minimally invasive". What prevents us from creating an nd4j-osgi module like the nd4j-uberjar module that doesn't require any changes to the other modules?

timothyjward:
> I don't see these changes as "minimally invasive". What prevents us from creating an nd4j-osgi module like the nd4j-uberjar module that doesn't require any changes to the other modules?

So in this PR:

* The ND4J repackaging of Jackson is *enhanced* with OSGi metadata in the manifest, but there are no other changes
* There are some internal ND4J changes to allow `ServiceLoader` to find implementations of things when running in OSGi, vanilla Java usage is unaffected
* There is a new ND4J "frontend" Maven Module which packages up the ND4J core, common, context and buffer
* There is a new ND4J "native backend" Maven Module which packages up the native-api and native libraries for the CPU backend
* There is a new ND4J "OSGi test" project which runs tests in the OSGi environment

Of these the only ones which affect existing modules are the internal `ServiceLoader` changes and Jackson packaging changes, both of which are necessary if ND4J is to work in OSGi without forcing a choice of CPU/GPU at build time.

saudet:
Both backends can be built in the same uber JAR, that's not an issue.

If we need to add a dependency on OSGi like that, we'll need to make sure it doesn't break anything anywhere, on Android and Spark among things... Is this something you would have the time to test? :)

timothyjward:
> If we need to add a dependency on OSGi like that, we'll need to make sure it doesn't break anything anywhere, on Android and Spark among things... Is this something you would have the time to test? :)

I'm happy to help, but I'm not sure I have the appropriate infrastructure available. Unless of course you have solved this problem already?

saudet:
@timothyjward No, not really unfortunately. It's pretty manual running examples from dl4j-examples for most such integration tests.

saudet:
BTW, how did this PR get moved here?

agibsonccc:
@timothyjward @saudet with java 17 LTS forcing module-info I added a semi related PR: https://github.com/eclipse/deeplearning4j/pull/9626

I think I've at least solved part of the problem for modularization. I'm playing with how to add a meta backend that allows 1 or more backends to be present. We also have classloader overriding now. It might allow us to introduce a better way of introducing OSGI support. Any suggestions even if you dont' have time to work on this?

timothyjward:
> @timothyjward @saudet with java 17 LTS forcing module-info I added a semi related PR: #9626
> 
> I think I've at least solved part of the problem for modularization. I'm playing with how to add a meta backend that allows 1 or more backends to be present. We also have classloader overriding now. It might allow us to introduce a better way of introducing OSGI support. Any suggestions even if you dont' have time to work on this?

The biggest issues for OSGi were:

1. Packages split across multiple jar files.
2. Highly coupled API packages (API packages recieving/returning types from other packages)
3. Expectations that you could find implementation classes from the API jar classloader

I'm guessing that item 1 has been solved now as JPMS has a similar requirement. Item 2 has hopefully improved, but isn't necessarily a deal breaker. Item 3 is a big deal, but depending on what the *"classloader overriding"* support you've added offers it may be sufficient.

I'd suggest adding the `bnd-maven-plugin` to your build so that you start creating OSGi metadata for each of your jar files. You could then add an OSGi "integration test" project using the `bnd-testing-maven-plugin` to try some simple scenarios. This will probably not work at first, but will help to figure out:

1. Problems with resolving (i.e. missing dependencies, unexported API packages)
2. Problems at runtime (typically expectations of a flat classpath)

These issues can then be worked through one at a time as they are identified until the basic OSGi tests start passing. My guess is that there probably won't be many issues left if the split packages are already gone.

agibsonccc:
@timothyjward thanks for the comments that helps a lot. I'll try to address what I can in the upgrade.

Regarding  2 and 3:

2. When splitting things up one problem definitely was the packages things were in. That could stand to be cleaned up. I might do that after releasing the 1.0. For now I'll probably just leave the packages mostly intact and just changing what we need. I agree this should be revisited.

3.  We've added the ability to override everything now with ND4JClassLoading/DL4JClassloading I think with proper injection those issues are mostly solved now.

I'll use this PR for reference and see what I can do about adding proper OSGI metadata. Thanks a lot for your comments.

PR #8104, opened 2019-08-16
fix quickstart

RobAltena:
Update template with fix.
Signed-off-by: Robert Altena <Rob@Ra-ai.com>


saudet:
Does this generate the same file as the one in pull https://github.com/eclipse/deeplearning4j-docs/pull/188?

RobAltena:
It should, but without the error.

RobAltena:
Ok, this went horribly wrong after pointing my fork back to skymondIO.

PR #8106, opened 2019-08-16
RL4J - VideoRecorder

aboulang2002:

Currently in RL4J, the HistoryProcessor can be configured to capture a video of the input. On the conceptual side, a 'HistoryProcessor' shouldn't be responsible of handling the video creation. Also, when we'll start using the DataSet for observations, the HistoryProcessor will receive observations that are transformed to be used by the Learning classes instead of the raw observations of the game.

In this PR, I isolated the video recording code in a simple class, the VideoRecorder.

## How was this patch tested?

Manual Tests. Can it be unit tested?

## Quick checklist

### DIFF
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/HistoryProcessor.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/HistoryProcessor.java
@@ -21,6 +21,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.queue.CircularFifoQueue;
 import org.bytedeco.javacv.*;
 import org.datavec.image.loader.NativeImageLoader;
+import org.deeplearning4j.rl4j.util.VideoRecorder;
 import org.nd4j.linalg.api.buffer.DataType;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.factory.Nd4j;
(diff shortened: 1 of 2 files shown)


saudet:
Looks good, but I think we could put it in the relevant places right away to actually record images?

aboulang2002:
That's the thing; video recording doesn't have a place anymore since I'll be removing/deprecating it from the HistoryProcessor. It should go somewhere in the MDP, but the MDP is only an interface in RL4J; it's up to the user to implement it.
Maybe we should add a base class for MDP for everything that is video/image-sequence based?

saudet:
I'd kind of see it along with the rest of the logging stuff, around DataManager maybe?

aboulang2002:
The file is located in the same directory as DataManager. Is it what you meant?

WIP: I'd like to change the MDPs in ALE, GYM, VizDoom and Malmo to use a base ImageSequenceMDP that will handle the video recording.

saudet:
I mean, the DataManager contains the path where the videos are going to be recorded, so it makes sense if that one does the recording too. I doesn't feel like it's something that the environment should be responsible for.

aboulang2002:
I agree, it does not belong to the environment. But, remember that the learning classes will receive a DataSet with all the transforms already applied, and that's what the DataManager has access to. Plus, I don't like the fact that the DataManager tries to do a lot of things. 

What I had in mind is that something will have to handle the observations given by the MDP (a byte array) and convert it to a DataSet. As you know, that byte array can represent anything like a video feed or the positions of pieces on a gaming board. I'm not 100% sure how to do that elegantly. It's easy if what you have is a single video feed or a 1D signal, but I'd like RL4J to be able to easily support advanced cases such as multiple video, audio or sensor feeds all at once. Anyways, for now, what I had in mind is introduce a handler (interface) between the MDP and Learning do the conversion and implement a video handler that can also records the video.

What do you think?

I'll be away for a few days but I'll work on this when I return.

saudet:
I see what you mean. We need access to data before transformation, so the environment has to provide some information about that explicitly. Sounds good, I'll leave you think of something! :)

aboulang2002:
I changed HistoryProcessor to use VideoRecorder. As I wrote in a previous comment, I think its place will be in the class that will handle observations but, for now, the goal of this PR is just to introduce VideoRecorder.

saudet (approved):

PR #8202, opened 2019-09-05
Removed unused declarations that seemed to bother ncc

yquemener:
The tests `concat_test18` and `concat_test19`, declared in `DeclarableOpsTests9.cpp` initializes an array that is never used, except to reference its size. It looks like it is a remain of an earlier version of the tests. For some reason yet to be investigated, `ncc`'s optimizer does not like these declarations.

Removing the declarations and replacing the references to `arrays.size()` by a constant makes the test compile and run correctly.

### DIFF
--- a/libnd4j/tests_cpu/layers_tests/DeclarableOpsTests9.cpp
+++ b/libnd4j/tests_cpu/layers_tests/DeclarableOpsTests9.cpp
@@ -615,12 +615,11 @@ TEST_F(DeclarableOpsTests9, concat_test17) {
 
 //////////////////////////////////////////////////////////////////////
 TEST_F(DeclarableOpsTests9, concat_test18) {
-    std::array<NDArray*, 2000> arrays;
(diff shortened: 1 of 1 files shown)


raver119:
Ye, first version of these tests used array, but later it was refactored to use context instead.

saudet:
Oops, I merged this too fast. @yquemener Please sign the ECA as detailed here:
https://deeplearning4j.org/eclipse-contributors

yquemener:
Signed it. I guess I need to sign the commits, as they indicate in the doc as well?

yquemener:
Here is the signed commit: https://github.com/yquemener/deeplearning4j/commit/ec8aa04d12d28142f97fd70343c233ecfbfa68d9

saudet:
Yeah, but it's a bit late, I've already merged the PR :) I don't think this is going to cause problems, but yes do it next time, thanks!

PR #8216, opened 2019-09-08
Platform helpers

raver119:
This PR improves the way how we deal with platform-specific optimizations/helpers. I.e. mkl-dnn or cuDNN.

Now we directly specify at build time, which helpers we're going to build into this binary, effectively making possible combination of generic computation devices + specialized devices.

I.e. cpu + mkldnn, or cpu + nnapi, or cuda + cudnn, or cuda + tpu/fpga implementations of specific operations.

Also, this PR moves existing mkl-dnn implementations into separate folder, and provides clear way to define optimized implementation for specific platform.

Also:
- this PR updates mkl-dnn version to 1.0.2
- this PR adds support for runtime CPU features check
- conv ops now use convolution_auto algorithm for mkldnn


nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuNDArrayFactory.java
```diff
@@ -90,6 +90,55 @@ public void createBlas() {
...
+        if (!nativeOps.isMinimalRequirementsMet()) {
+            // this means cpu binary was built for some arch support, we don't have on this box
+
+            val binaryLevel = nativeOps.binaryLevel();
+            val optimalLevel = nativeOps.optimalLevel();
+
+            switch (binaryLevel) {
+                case 3: {
+                        // avx-512 binary
+                        log.error("Trying to use AVX-512 binary without actual CPU support for it");
```
saudet: It would probably be nicer to use JavaCPP to load the right binaries instead of only giving warnings like that.

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
```diff
@@ -257,19 +257,15 @@
                     <linkPath>${libnd4jhome}/blasbuild/cpu/blas</linkPath>
                     <buildResources>
                         <buildResource>/${javacpp.platform.library.path}/</buildResource>
-                        <buildResource>/org/bytedeco/mkldnn/${javacpp.platform}/</buildResource>
```
saudet: This should be replaced by the full version of MKL instead. There's no reason to not use it by default.

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuNDArrayFactory.java
raver119: RIght, that's exactly the plan. We want jcpp to manage that, but in case user explicitly uses, say, avx512 on one box, and tries to run on avx2 box - instead of SIGILL crash he'll see nice message that explains what happens and what to do.

libnd4j/CMakeLists.txt
```diff
@@ -50,25 +50,6 @@ if(NOT CUDA_BLAS)
...
-        message("We have manually provided MKLDNN: ${MKLDNN_PATH}")
-        set(HAVE_MKLDNN 1)
-        include_directories(${MKLDNN_PATH}/include/)
-        link_directories(${MKLDNN_PATH} ${MKLDNN_PATH}/lib/)
-        IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-            set(MKLDNN_LIBRARIES mkldnn mklml_intel)
-        else()
-            set(MKLDNN_LIBRARIES mkldnn mklml)
-        endif()
-    elseif (NOT "${OPENBLAS_PATH}" STREQUAL "")
```
saudet: Don't remove OpenBLAS

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
```diff
@@ -204,17 +204,6 @@
...
-                        <groupId>org.bytedeco</groupId>
-                        <artifactId>mkl-dnn</artifactId>
-                        <version>${mkl-dnn.javacpp.version}</version>
-                    </dependency>
-                    <dependency>
-                        <groupId>org.bytedeco</groupId>
-                        <artifactId>mkl-dnn</artifactId>
-                        <version>${mkl-dnn.javacpp.version}</version>
-                        <classifier>${dependency.platform2}</classifier>
-                    </dependency>
```
saudet: Like I mentioned previously, replace those with the full version of MKL. There is usually no reason to prefer OpenBLAS over MKL!

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
```diff
@@ -68,17 +68,6 @@
...
-            <groupId>org.bytedeco</groupId>
-            <artifactId>mkl-dnn</artifactId>
-            <version>${mkl-dnn.javacpp.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.bytedeco</groupId>
-            <artifactId>mkl-dnn</artifactId>
-            <version>${mkl-dnn.javacpp.version}</version>
-            <classifier>${dependency.platform2}</classifier>
-        </dependency>
```
saudet: Again, don't just remove it. Put MKL there instead!

libnd4j/CMakeLists.txt
raver119: It's not removed, if OPENBLAS_PATH is provded - it'll be used

saudet: Why did you remove those lines though? I'm pretty sure OpenBLAS isn't going to get linked

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
raver119: so, something like this?

```
<artifactId>mkl</artifactId>
<version>${mkl.javacpp.version}</version>
<classifier>${dependency.platform2}</classifier>
```

libnd4j/CMakeLists.txt
raver119: Mmm, let me double check then. The plan was to keep OPENBLAS_PATH if it was provided

raver119: Yep, you're right. Missed else part. Restored it back.

nd4j/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
saudet: Yes, something like that. Current version is `2019.4-1.5.1`. Not too sure if we need all the same dependencies, but just replacing everything mechanically for now should give something that works.

raver119: Ok, I'll open new PR with MKL tomorrow morning, and get back to you.

Issue #8107, opened 2019-08-17
The output from the current DQN and target DQN


flyheroljg:
#### Issue Description

When I study the file ‘QLearningDiscrete.java’, I find a question in the function ‘setTarge’.
Function setTarge should be prepare for updating the current DQN which approximates the values of behavior functions. Therefore, in the return value pair(obs, dqnOutputAr), the first is the input for training and the second is the input for labeling. 
The standard DQN algorithm requires that obs and dqnOutputAr should be outputted by the current DQN and the target DQN, respectively. However, in your code, the variable ‘yTar’, which should be the target value of TD, is inferred from the current DQN, rather than the target DQN. Because the variable ‘dqnOutAr’ is outputted by the current DQN, both obs and dqnOutAr are outputted by the current DQN. Why is it? Did I not understand the standard DQN algorithm correctly?

#### Version Information

Please indicate relevant versions, including, if relevant:
RL4J, file:  QLearningDiscrete.java

#### Additional Information

Where applicable, please also provide:

protected Pair<INDArray, INDArray> setTarget(ArrayList<Transition<Integer>> transitions) {
        if (transitions.size() == 0)
            throw new IllegalArgumentException("too few transitions");

        int size = transitions.size();

        int[] shape = getHistoryProcessor() == null ? getMdp().getObservationSpace().getShape()
                        : getHistoryProcessor().getConf().getShape();
        int[] nshape = makeShape(size, shape);
        INDArray obs = Nd4j.create(nshape);
        INDArray nextObs = Nd4j.create(nshape);
        int[] actions = new int[size];
        boolean[] areTerminal = new boolean[size];

        for (int i = 0; i < size; i++) {
            Transition<Integer> trans = transitions.get(i);
            areTerminal[i] = trans.isTerminal();
            actions[i] = trans.getAction();

            INDArray[] obsArray = trans.getObservation();
            if (obs.rank() == 2) {
                obs.putRow(i, obsArray[0]);
            } else {
                for (int j = 0; j < obsArray.length; j++) {
                    obs.put(new INDArrayIndex[] {NDArrayIndex.point(i), NDArrayIndex.point(j)}, obsArray[j]);
                }
            }

            INDArray[] nextObsArray = Transition.append(trans.getObservation(), trans.getNextObservation());
            if (nextObs.rank() == 2) {
                nextObs.putRow(i, nextObsArray[0]);
            } else {
                for (int j = 0; j < nextObsArray.length; j++) {
[...]

PR #8250, opened 2019-09-22
RL4J - Fix for setTarget() (issue #8107)

aboulang2002:

See issue #8107 

After reading again "_Playing Atari with Deep Reinforcement Learning_" (http://arxiv.org/abs/1312.5602), I still think there is a problem with the current implementation of standard DQN, but I don't think that the target-network should be used for `dqnOutputAr`.

I have also read the section 3 "_Understanding Deep Q-Network_" of "A _Theoretical Analysis of Deep Q-Learning_" (https://arxiv.org/pdf/1901.00137.pdf).
And also https://medium.com/@qempsil0914/deep-q-learning-part2-double-deep-q-network-double-dqn-b8fc9212bbb2 which has a good overview of DQN.
Plus I reviewed a few DQN implementations on github. For example, Google's TensorLayer (https://github.com/tensorlayer/tensorlayer/blob/72666dbf23718e8710dc1069dbcf3231d131dff2/examples/reinforcement_learning/tutorial_DQN_variants.py).

First, the target-network should be used to compute `tempQ` in the standard DQN algorithm.
This can be seen in the equation 3.1 (and the one above in the text) of https://arxiv.org/pdf/1901.00137.pdf and also from lines 6-8 of `tutorial_DQN_variants.py`:
```
	The max operator in standard DQN uses the same values both to select and to evaluate an action by
	Q(s_t, a_t) = R_{t+1} + \gamma * max_{a}Q_{tar}(s_{t+1}, a)
```
We can see that the target-network (`Q_{tar}`) should be used.

Second, the target-network is only used as a stable base to compute the loss function, and not "to obtain the labels" as @flyheroljg says. This means that the Q-Network should used for `dqnOutputAr`. This can be seen on line 8 of `tutorial_DQN_variants.py` and also on lines 322-325:
```
                # calculate loss
                with tf.GradientTape() as q_tape:
                    b_q = tf.reduce_sum(qnet(b_o) * tf.one_hot(b_a, out_dim), 1)
                    loss = tf.reduce_mean(huber_loss(b_q - (b_r + reward_gamma * b_q_)))
```
The above snippet is used for Double-DQN, but only the td-target part (`(b_r + reward_gamma * b_q_)` above) is different between standard and double DQN in the computation of the loss function.
This can also be seen in the section "_Double Q-Learning Algorithm_" of https://medium.com/@qempsil0914/deep-q-learning-part2-double-deep-q-network-double-dqn-b8fc9212bbb2, as well as equation 3.3 of https://arxiv.org/pdf/1901.00137.pdf

## How was this patch tested?

With cartpole-v0 (re-written in java because couldn't make gym's version to work; can add it to this PR if requested)

### DIFF
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/qlearning/discrete/QLearningDiscrete.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/qlearning/discrete/QLearningDiscrete.java
@@ -232,15 +232,14 @@ public abstract class QLearningDiscrete<O extends Encodable> extends QLearning<O
         INDArray dqnOutputAr = dqnOutput(obs);
 
         INDArray dqnOutputNext = dqnOutput(nextObs);
-        INDArray targetDqnOutputNext = null;
+        INDArray targetDqnOutputNext = targetDqnOutput(nextObs);
 
         INDArray tempQ = null;
         INDArray getMaxAction = null;
         if (getConfiguration().isDoubleDQN()) {
-            targetDqnOutputNext = targetDqnOutput(nextObs);
             getMaxAction = Nd4j.argMax(dqnOutputNext, 1);
         } else {
-            tempQ = Nd4j.max(dqnOutputNext, 1);
+            tempQ = Nd4j.max(targetDqnOutputNext, 1);
         }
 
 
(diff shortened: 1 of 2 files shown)


saudet (approved):
That looks good! Please do add your version of Cartpole, and let's merge this :) Thanks

aboulang2002:
I added CartpoleNative, a native java port of GYM's cartpole-v0. The rendering has not been ported. I may add it later but it's lower on my list of todos.

PR #8273, opened 2019-10-08
nd4j-api cleanup.

RobAltena:
Cleans up unused code from nd4j-api

## How was this patch tested?
Unit tests run with same results as master.


AlexDBlack (changes requested):
Let's keep the WeightInitScheme instances.
The fact that they are unused represents a test coverage issue, not that they should be removed. They can be used when creating a variable via SameDiff.var(String, WeightInitScheme, DataType, long...) etc

AlexDBlack (approved):
LGTM, thanks!

PR #8282, opened 2019-10-09
RL4J - Changed to use Nd4j Random instead of java.util.Random

aboulang2002:

Several places in RL4J uses java.util.Random and does not use the seed. This makes it impossible to have identical results with identical runs. Plus java.util.Random makes it very difficult to unit test components that uses random.

## How was this patch tested?

Unit tests and manual tests

## Quick checklist

### DIFF
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/ILearning.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/ILearning.java
@@ -42,7 +42,7 @@ public interface ILearning<O extends Encodable, A, AS extends ActionSpace<A>> ex
 
     interface LConfiguration {
 
-        int getSeed();
+        Integer getSeed();
 
         int getMaxEpochStep();
 
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/Learning.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/Learning.java
@@ -29,8 +29,6 @@ import org.deeplearning4j.rl4j.space.Encodable;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.factory.Nd4j;
 
-import java.util.Random;
-
 /**
  * @author rubenfiszel (ruben.fiszel@epfl.ch) on 7/27/16.
  *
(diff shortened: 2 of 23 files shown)


saudet:
@AlexDBlack Should we be using `Nd4j.getRandomFactory().getNewRandomInstance()` for that? I don't see it being used a lot in ND4J or DL4J...

aboulang2002:
Nd4j.getRandom() put and/or get the instance from a ThreadLocal storage. I was afraid that if Nd4j.getRandom().setSeed(...) is called multiple times, a user might think he is setting the seed on different Random instances but in fact is re-setting the seed on the same instance. This might cause unwanted patterns in the random sequences. This can be solved easily if we just create a new Random with every instance. 

On the other hand, I could make sure the seed is only set once and use Nd4j.getRandom(0 everywhere.

saudet:
I see, that sounds like a prudent thing to do. I'm wondering if we should be using that factory like that though. I'll let Alex decide.

AlexDBlack:
Throughout DL4J and SameDiff, we have tended to use the thread local random (Nd4j.getRandom()) pretty much everywhere, which works well enough most of the time.
The downside of course is that the seed/state is attached to the thread, not to the model. Which may or may not matter.

The problem with using getNewRandomInstance is that anything internally to the network won't be using it that random instance.
So, if you have say a model with dropout, that will continue to use the thread local random. Which means that your model training won't be made deterministic just by creating the new instance with the specified seed. I'm not 100% following what's the intended behavior here when setting the seed, and if that's an issue or not...

aboulang2002:
I decided to use `Nd4j.getRandom()` because it does the job just as well as using the factory and I feel the use of Random in RL4J needs to be similar to the rest of DeepLearning4j. As for my earlier concern that a user calling `Nd4j.getRandom().setSeed(mySeed)` might unknowingly cause side effects, it's nothing that reading the docs can't clear up.

So these are the specs I implemented:
 - If no seed is set, the learning is completely non-deterministic (i.e. `setSeed()` should never called)
 - If the seed is set, the learning should be completely deterministic (i.e. every run should always give the same results for a given seed value)
 - The classes that need `Random` should carry an internal reference. This is easier in the unit tests when we can inject every instance with different mocks that each contains precise 'random' values. (Not all ctors presently support injection of `Random` but I will add them when required)

saudet:
@AlexDBlack The thing is, RL4J "uses" DL4J so whatever happens in DL4J w.r.t to seeding comes in too late, and not all threads interact with DL4J either. How do you think we should handle this?

saudet (approved):
This looks good to me, and I'm getting reproducible results with Cartpole.

AlexDBlack (approved):
LGTM :+1:

PR #8301, opened 2019-10-19
Samediff import doc

RobAltena:
First version of samediff import documentation

### DIFF
--- a/docs/samediff/templates/model-import.md
+++ b/docs/samediff/templates/model-import.md
@@ -1,32 +1,31 @@
 ---
-title: Getting started: importing TensorFlow and ONNX models into SameDiff
+title: Getting started: importing TensorFlow models into SameDiff
 short_title: Model import
-description: importing TensorFlow and ONNX models into SameDiff
+description: importing TensorFlow models into SameDiff
 category: SameDiff
 weight: 3
 ---
 
 # Getting started: importing TensorFlow models into SameDiff
 
-## What models can be imported into samediff
+## What models can be imported into SameDiff
 
-Currently samediff supports the import of Tensorflow frozen graphs through the various Samediff.importFrozenTF methods. 
-Tensorflow documentation on frozen models can be found [here](https://www.tensorflow.org/guide/saved_model#the_savedmodel_format_on_disk). 
+Currently SameDiff supports the import of TensorFlow frozen graphs through the various SameDiff.importFrozenTF methods. 
+TensorFlow documentation on frozen models can be found [here](https://www.TensorFlow.org/guide/saved_model#the_savedmodel_format_on_disk). 
 
-    import org.nd4j.autodiff.samediff.SameDiff;
+    import org.nd4j.autodiff.SameDiff.SameDiff;
     
     SameDiff sd = SameDiff.importFrozenTF(modelFile);
     
  ## Finding the model input/outputs and running inference
  
- After you import the Tensorflow model there are 2 ways to find the inputs and outputs. The first method is to look at the output of
+ After you import the TensorFlow model there are 2 ways to find the inputs and outputs. The first method is to look at the output of
  
      sd.summary();
      
- Where the input variables are the output of no ops, and the output variables are the input of no ops.  The other way to find the inputs and outputs is
+ Where the input variables are the output of no ops, and the output variables are the input of no ops.  Another way to find the inputs is
  
       List<String> inputs = sd.inputs();
-      List<String> outputs = sd.outputs());
     
  To run inference use:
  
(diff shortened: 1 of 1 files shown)


docs/samediff/templates/model-import.md
```diff
@@ -1 +1,62 @@
-# Getting started: importing TensorFlow and ONNX models into SameDiff
\ No newline at end of file
+---
+title: Getting started: importing TensorFlow and ONNX models into SameDiff
```
AlexDBlack: No ONNX for now. It's only planned at this point.

docs/samediff/templates/model-import.md
```diff
@@ -1 +1,62 @@
...
+description: importing TensorFlow and ONNX models into SameDiff
+category: SameDiff
+weight: 3
+---
+
+# Getting started: importing TensorFlow models into SameDiff
+
+## What models can be imported into samediff
+
+Currently samediff supports the import of Tensorflow frozen graphs through the various Samediff.importFrozenTF methods.
```
AlexDBlack: Capitalization: `samediff` -> `SameDiff`
`Tensorflow` -> `TensorFlow`
`Samediff.importFrozenTF` -> `SameDiff.importFrozenTF`

Same things in a few other places in the doc, I won't flag others...

docs/samediff/templates/model-import.md
```diff
@@ -1 +1,62 @@
...
+ ## Finding the model input/outputs and running inference
+ 
+ After you import the Tensorflow model there are 2 ways to find the inputs and outputs. The first method is to look at the output of
+ 
+     sd.summary();
+     
+ Where the input variables are the output of no ops, and the output variables are the input of no ops.  The other way to find the inputs and outputs is
+ 
+      List<String> inputs = sd.inputs();
+      List<String> outputs = sd.outputs());
```
AlexDBlack: inputs() is good - that just tells you placeholders. Let's note placeholders, and maybe link to other SameDiff page that explains this.
As for outputs() - that will be going away. Yes it's in 1.0.0-beta5, but it's not robust enough - it's basically "variables that don't feed into any other ops" - which often won't be the real predictions that users want - but rather the loss function or some irrelevant unused output.

So users should just look at the summary instead and infer the outputs from that.

docs/samediff/templates/model-import.md
```diff
@@ -1 +1,62 @@
...
+    
+ To run inference use:
+ 
+    INDArray out = sd.batchOutput()
+        .input(inputs, inputArray)
+        .output(outputs)
+        .execSingle();
+
+##  Node skipping, import overrides and model validation.
```
AlexDBlack: Let's split this up into two sections: the former many users might consider using. The latter: not many users should need.
`## Import Validation` + provide simple example
`## Advanced: Node Skipping and Import Overrides`

docs/samediff/templates/model-import.md
```diff
@@ -1 +1,62 @@
...
+
+It is possible to remove nodes from the network. For example Tensorflow 1.x models can have hard coded dropout layers. 
+See the [BERT Graph test](https://github.com/eclipse/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/imports/TFGraphs/BERTGraphTest.java#L114-L150) for an example.
+We have a Tensorflow graph analyzing utility which will report any missing operations (operations that still need to be implemented) [here](https://github.com/eclipse/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/imports/tensorflow/TensorFlowImportValidator.java)
+
+
+## List of models known to work with Samediff.
```
AlexDBlack: For all of these, let's link to the actual model files, not the test resources
These can be found in the tf_model.txt files - for example `http://download.tensorflow.org/models/tflite_11_05_08/mobilenet_v2_1.0_224.tgz`

Also we'll be using https://github.com/KonduitAI/dl4j-test-resources from here on out

AlexDBlack (changes requested):
So far so good, other than a few changes flagged.

docs/samediff/templates/model-import.md
```diff
@@ -1 +1,61 @@
...
+ Where the input variables are the output of no ops, and the output variables are the input of no ops.  Another way to find the inputs is
+ 
+      List<String> inputs = sd.inputs();
+    
+ To run inference use:
+ 
+    INDArray out = sd.batchOutput()
+        .input(inputs, inputArray)
+        .output(outputs)
+        .execSingle();
```
AlexDBlack, line 35: "For multiple outputs, use exec() instead of execSingle(), to return a `Map<String,INDArray>` of outputs instead.
Alternatively, you can use methods such as `SameDiff.output(Map<String, INDArray> placeholders, String... outputs)` to get the same output."

docs/samediff/templates/model-import.md
```diff
@@ -1 +1,61 @@
...
+- [labels](https://github.com/KonduitAI/dl4j-test-resources/tree/master/src/main/resources/tf_graphs/zoo_models/labels)
+- [mobilenet_v1_0.5_128](http://download.tensorflow.org/models/mobilenet_v1_2018_02_22/mobilenet_v1_0.5_128.tgz)
+- [mobilenet_v2_1.0_224](http://download.tensorflow.org/models/tflite_11_05_08/mobilenet_v2_1.0_224.tgz)
+- [nasnet_mobile_2018_04_27](https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/nasnet_mobile_2018_04_27.tgz)
+- [resnetv2_imagenet_frozen_graph](http://download.tensorflow.org/models/official/resnetv2_imagenet_frozen_graph.pb)
+- [squeezenet_2018_04_27](https://storage.googleapis.com/download.tensorflow.org/models/tflite/model_zoo/upload_20180427/squeezenet_2018_04_27.tgz)
+- [temperature_bidirectional_63](https://deeplearning4jblob.blob.core.windows.net/testresources/temperature_bidirectional_63_frozenmodel.pb)
+- [temperature_stacked_63](https://deeplearning4jblob.blob.core.windows.net/testresources/temperature_stacked_63_frozenmodel.pb)
+- [text_gen_81](https://deeplearning4jblob.blob.core.windows.net/testresources/text_gen_81_frozenmodel.pb)
```
AlexDBlack: Let's add a section with this, which should be good enough for now. We'll get proper coverage info up at a later date.

```
## Operations Coverage

SameDiff's TensorFlow import is still being developed, and does not yet have support for every single operation and datatype in TensorFlow.
Almost all of the common/standard operations are importable and tested, however - including almost everything in the tf, tf.math, tf.layers, tf.losses, tf.bitwise and tf.nn namespaces. The majority of existing pretrained models out there should be importable into SameDiff.

If you run into an operation that can't be imported, feel free to open an issue here: https://github.com/eclipse/deeplearning4j/issues
```

AlexDBlack (changes requested):
Couple of small tweaks, otherwise this can go up soon. :+1:

AlexDBlack (approved):
LGTM :+1:

PR #8368, opened 2019-11-08
RL4J - Added Observable & LegacyMDPWrapper

aboulang2002:

In this PR, I will remove all the processing done on the observations in the Learning classes. In my mind, the observation should be 'opaque' to the Learning, that is the Learning classes should not know any detail about it.

I also plan to replace the `Encodable` with an `Observation` class. The `Encodable` works if the observation is simple, for example a video frame. But suppose a use case where additional observations could be useful or necessary, for example a video game in which an audio channel carry the info on an imminent enemy attack, the `Encodable` is not good enough.
Eventually, I'd like the Observation class to be able to handle multiple observation channels, but since the goal of this PR is only to remove the processing done to the observation from the learning classes, I will limit for now the Observation class as a dummy container that output a single `INDArray`.

Also, Existing MDPs should continue to work. So, I am adding a `LegacyMDPWrapper` as the glue between Encodable-MDPs and Learning classes that will use the Observation class.

I changed a refac test (`refac_QLearningDiscrete.trainStep`) to use `trainEpoch()` instead of `trainStep()`, a few things surfaced (not yet fixed):
1) step from `initMdp` is used to set the variable `step` in `QLearning.trainEpoch()`, but it is never used to set `stepCounter` (which is used in `QLearningDiscrete.trainStep()`). The result is that if `initMdp` returns a frame that should be skipped, `trainStep()` will not skip it since `stepCounter` will be 0. In other terms, if we use skipFrame = 2 and history length = 5, when `initMDP()` returns we will have: history = { 0, 2, 4, 6, 8 }, step = 9 and stepCounter = 0. trainStep() uses stepCounter to know if it's a skipped frame and thinks it's not. So history becomes: { 2, 4, 6, 8, 9 } which is wrong.

2) In `QLearningDiscrete`, we create the Transition this way:
`Transition<Integer> trans = new Transition(history, action, accuReward, stepReply.isDone(), nhistory[0]);`
`nhistory` is an array and `nhistory[0]` is the oldest element. The statement above would create a Transition with the oldest element in the nhistory as the next observation which is incorrect. Again in other words, if history is { 0, 2, 4, 6, 8 } and the next observation is 9, the nhistory will be { 2, 4, 6, 8, 9 } and 2 will be used as the next observation.

## Quick checklist

### DIFF
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/ILearning.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/ILearning.java
@@ -26,7 +26,7 @@ import org.deeplearning4j.rl4j.space.Encodable;
  *
  * A common interface that any training method should implement
  */
-public interface ILearning<O extends Encodable, A, AS extends ActionSpace<A>> extends StepCountable {
+public interface ILearning<O, A, AS extends ActionSpace<A>> extends StepCountable {
 
     IPolicy<O, A> getPolicy();
 
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/Learning.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/Learning.java
@@ -39,7 +39,7 @@ import org.nd4j.linalg.factory.Nd4j;
  *
  */
 @Slf4j
-public abstract class Learning<O extends Encodable, A, AS extends ActionSpace<A>, NN extends NeuralNet>
+public abstract class Learning<O, A, AS extends ActionSpace<A>, NN extends NeuralNet>
                 implements ILearning<O, A, AS>, NeuralNetFetchable<NN> {
 
     @Getter @Setter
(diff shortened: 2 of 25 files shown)


aboulang2002:
In the last commit, I moved most of the observation processing out of QLearningDiscrete. Next step is use DataSets and to clean the mess in LegacyMDPWrapper.

aboulang2002:
Summary of the previous commit:
- Observation is now using a DataSet, although all the processing is still in LegacyMDPWrapper.
- Cleaned up the shape/observation manipulation from BaseTDTargetAlgorithm and QLearningDiscrete.
- Added methods to build observation and nextObservation in Transition, plus unit tests.

saudet:
Try to finalize these changes and let's merge this before it goes over 1000 lines, thanks!

aboulang2002:
Agreed. It wouldn't  be reasonable to put everything in this PR.

aboulang2002:
Forgot to remove the [WIP] tag from the title. Sorry about that.

saudet (approved):

PR #8414, opened 2019-11-18
Verify that the slice and stack are correct

zdl010:
Verify that the gradient calculation is correct after slicing


AlexDBlack:
Looks like these 2 tests are already in master:
https://github.com/eclipse/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/samediff/SameDiffTests.java#L3470-L3506

PR #8422, opened 2019-11-19
fix pydatavec for python 3

maxpumperla:
... and python2 install problems


ShamsUlAzeem:
@maxpumperla is it simple to add the chaining mechanism that Philip mentioned here? https://stackoverflow.com/c/konduit/questions/11

maxpumperla:
@ShamsUlAzeem buddy, I already did that a week ago :D

PR #8471, opened 2019-11-30
RL4J Remove processing done on observations in Policy & Async

aboulang2002:

In this PR, I move the processing done on the observations in `Policy.play()`, `AsyncThread` and `AsyncThreadDiscrete`, in the same way to what has been done in `QLearning` and `QLearningDiscrete`.

## How was this patch tested?

Manual and unit tests

## Quick checklist

### DIFF
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/async/AsyncGlobal.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/async/AsyncGlobal.java
@@ -57,6 +57,7 @@ public class AsyncGlobal<NN extends NeuralNet> extends Thread implements IAsyncG
     final private NN current;
     final private ConcurrentLinkedQueue<Pair<Gradient[], Integer>> queue;
     final private AsyncConfiguration a3cc;
+    private final IAsyncLearning learning;
     @Getter
     private AtomicInteger T = new AtomicInteger(0);
     @Getter
@@ -64,10 +65,11 @@ public class AsyncGlobal<NN extends NeuralNet> extends Thread implements IAsyncG
     @Getter
     private boolean running = true;
 
-    public AsyncGlobal(NN initial, AsyncConfiguration a3cc) {
+    public AsyncGlobal(NN initial, AsyncConfiguration a3cc, IAsyncLearning learning) {
         this.current = initial;
         target = (NN) initial.clone();
         this.a3cc = a3cc;
+        this.learning = learning;
         queue = new ConcurrentLinkedQueue<>();
     }
 
@@ -106,11 +108,14 @@ public class AsyncGlobal<NN extends NeuralNet> extends Thread implements IAsyncG
     }
 
     /**
-     * Force the immediate termination of the AsyncGlobal instance. Queued work items will be discarded.
+     * Force the immediate termination of the AsyncGlobal instance. Queued work items will be discarded and the AsyncLearning instance will be forced to terminate too.
      */
     public void terminate() {
-        running = false;
-        queue.clear();
+        if(running) {
+            running = false;
+            queue.clear();
+            learning.terminate();
+        }
     }
 
 }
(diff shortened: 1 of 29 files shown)


aboulang2002:
Last commit:
 - Removed the processing done on observations in `Policy.play()`
 - Fix: Made sure both `QNetwork` and `QTargetNetwork` were initialized in `QLearning`, plus unit tests

aboulang2002:
Last commit:
 - Fixed the logic in run() that would initialize the MDP just prior to terminating.
 - In `AsyncThreadDiscrete.trainSubEpoch`, fixed a problem where the current observation would always be the last non-skipped observation instead of the current one.
 - Adjusted the async unit tests to test the whole `run()` instead of just a `trainSubEpoch()`.
 - Added (refac) unit tests to `A3CThreadDiscrete.calcGradient()` and `AsyncNStepQLearningThreadDiscrete.calcGradient()`

There is something I don't like in `AsyncThreadDiscrete` which, in the code, is described by this:
`        //a bit of a trick usable because of how the stack is treated to init R`

Explanation:
The async calcGradient() method needs to compute the discounted reward for each transition.
To do this, the strangely named parameter `rewards` (an ordered list of `MiniTrans` from most-recent to oldest) is used.
The MiniTrans, as stated on the javadoc, is similar to the Transition except that it doesn't have the next observation.

This is the case for all elements of `rewards` except the last. If you read the code after the 'bit of a trick' comment, `hstack` will be:
 - With my fix (2nd line in the "Last commit" section above): the next observation
 or
 - Before my fix: the next observation if it's a non-skipped frame or, if it's a skipped frame, the observation of the last non-skipped one.

Then, if you read `calcGradient()`, you can see that this last element is only used to get the initial value for the discounted reward; fortunately the observation of this special element is ignored.
But, from the AsyncThreadDiscrete's point of view, there is no guarantee that someone that is unaware of this trick won't use the observation of that last element in his implementation of `calcGradient()`.
I will leave it as-is in this PR because I want to remain focused on the removal of the observation processing, but I'd like to revisit this in the future.

saudet:
There's a lot of little "tricks" like that need to be improved alright.
Just make sure to stay within 1000 lines per PR :)

/cc @rubenfiszel

rubenfiszel:
Happy to see some changes on RL4J! I aim to take some time this weekend to refresh my memory about the codebase and its changes.

On a less code related note but since I see that the changes here touch Async and A3C: I have the distinct memory that my implementation of A3C when I wrote it was incorrect/not useful as you would have needed lock-free memory structures so that you would not be IO bound and that I hacked around a queued writer agent. Has that changed since?

Also curious about the state of RL4J. There has been many new techniques and innovations in RL since afaik. Is there a project/plan to implement them in RL4J?

aboulang2002:
Last commit:
I have removed the processing from` AsyncThreadDiscrete`. I won't be adding anything new to this PR, but I'm not ready for a review; I'd like to do more testing.

@rubenfiszel
I don't think that there currently is any official plan to change RL4J. As for me, I plan to refac the existing code and add unit tests at first. I'd like to add new techniques to RL4J but not before the code base is easier to work with. I don't mean this as a criticism of your work; I am really impressed by RL4J.

agibsonccc:
@saudet  mind updating @rubenfiszel

aboulang2002:
Last commit
I fixed a few problems:
- Fixed a problem where stopping `AsyncGlobal` would not stop `AsyncLearning`
- Fixed a problem where data would not be properly normalized when using the history processor.
- Fixed a problem where the observation would not have the correct shape when using the history processor (the size of the dimension 0 should be 1 for the stacked observation)

saudet:
@aboulang2002 I'm getting a crash with A3CALE. Is this happening for you? It's also happening with 1.0.0-beta6, so I don't think it's related to this pull request, but it didn't crash last time I tested 1.0.0-SNAPSHOT...

aboulang2002:
@saudet That's odd. For me, everything runs fine with 1.0.0-SNAPSHOT and 1.0.0-beta6
Checked with A3CALE, ALE and Cartpole

If it still crashes for you, please add a stack trace and I will investigate.

saudet:
Yeah, let's look at that at some point... I know the video recording has been broken for a while now. It's supposed to save images as returned by, for example, ALE, but it's just showing garbage these days. If you can take a look at that, that would probably solve everything :)

saudet (approved):

PR #8502, opened 2019-12-05
Update CMake toolchains for more recent versions of Android NDK

saudet:

The build now passes without crashing with at least Android NDK r18b.

## How was this patch tested?

Only checked that the build passes for all supported architectures of Android.


sshepel (approved):
Changes were tested on Jenkins, everything looks good!
Thanks @saudet

PR #8595, opened 2020-01-06
RL4J: Replace gym-java-client with JavaCPP

saudet:

[gym-http-api](https://github.com/openai/gym-http-api) is slow, buggy, hard to use, deprecated, and is no longer maintained by OpenAI, so here we remove the gym-java-client module that relies on it (moving any generic classes to rl4j-api), and instead use the new [JavaCPP Presets for Gym](https://github.com/bytedeco/javacpp-presets/tree/master/gym).

I also fixed a couple of issues that showed up when coupled with changes from @aboulang2002. (This is why I'm opening this PR here because the Konduit fork doesn't contain these changes yet.)

## How was this patch tested?

New unit test passes, and the cartpole samples from rl4j-examples work just as well as with gym-java-client, but twice as fast.

/cc @aboulang2002 @rubenfiszel


rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/policy/ACPolicy.java
```diff
@@ -72,20 +72,23 @@ public Integer nextAction(Observation obs) {
...
-        float rVal = rnd.nextFloat();
-        for (int i = 0; i < output.length(); i++) {
-            //System.out.println(i + " " + rVal + " " + output.getFloat(i));
-            if (rVal < output.getFloat(i)) {
-                return i;
-            } else
-                rVal -= output.getFloat(i);
-        }
+        // TODO: Figure out a nicer way to deal with thread-safety issues of Deeplearning4j
+        synchronized (ACPolicy.class) {
```
aboulang2002: I don't think the `synchronized` is necessary here. Each thread uses it's own copy of the neural net (`actorCritic`). See `AsyncThreadDiscrete.trainSubEpoch()`:
```
    public SubEpochReturn trainSubEpoch(Observation obs, int nstep) {

        synchronized (getAsyncGlobal()) {
            current.copy(getAsyncGlobal().getCurrent());
        }
        ...
```

saudet: The networks return NaN if we don't do that. There's some issue with DL4J and the way the networks are used...

saudet: @AlexDBlack @raver119 Would you guys would have any idea why that happens?

saudet: We could remove this workaround for now, just to get the rest merged, and revisit the issue later. 

@aboulang2002 Would you have any other comments on this commit before we merge this?

aboulang2002: I agree that we can merge and revisit later. But, I'm worried that if it is an obscure DL4J issue, it will be very difficult to implement _Hogwild!_ which require simultaneous, don't-care-if-it-collides read and writes to a common neural net. 

Everything else looks good to me.
Good work with PR; I never got GYM's cartpole to work (I'm on Windows). This will be really helpful to me.

saudet: > I agree that we can merge and revisit later. But, I'm worried that if it is an obscure DL4J issue, it will be very difficult to implement _Hogwild!_ which require simultaneous, don't-care-if-it-collides read and writes to a common neural net.

Yes, DL4J needs a bit love in that department, but don't worry, these kinds of issues are not a low priority.

> Everything else looks good to me.
> Good work with PR; I never got GYM's cartpole to work (I'm on Windows). This will be really helpful to me.

But it works fine for you on Windows with this PR now, right?

aboulang2002: I haven't tried. I will checkout your branch and try it.

aboulang2002: I have trouble compiling the project. I propose you merge this PR and I will try again this weekend when I'll have more time.

aboulang2002: Well, after spending most of the day trying to run the tests I think I corrupted my setup. I can compile your branch just fine but, when I try to run GymEnvTest.testCartpole(), I get exceptions saying `no jnind4jcpu in java.library.path` and `no nd4jcpu in java.library.path`. It _should_ find jnind4jcpu and nd4jcpu; I even tried to copy them in the source base directory and made sure there were no unmet dependency with lucasg/Dependencies. Now my own branch also cause these exceptions.

I will need to de-install / reinstall everything but this time I'll do it on linux, like everybody else.

aboulang2002: I was able to make it work on linux and all tests of RL4J passes. But I give up trying to make it work on windows.
If someone comes here through a Google search, trying to compile RL4J, here is how I did it:
From a fresh install of Ubuntu 18.04 Desktop (Desktop version is important because otherwise you will have this error: `no jniopencv_highgui in java.library.path`):
```
sudo add-apt-repository ppa:natecarlson/maven3
sudo apt-get update
sudo apt-get install maven build-essential cmake libgomp1 git libopenblas-dev

git clone --depth 1 --branch master https://github.com/deeplearning4j/dl4j-test-resources
cd dl4j-test-resources
mvn install

cd ..
git clone https://github.com/eclipse/deeplearning4j.git
cd deeplearning4j/rl4j
mvn clean install -P test-nd4j-native
```

In my case, I checked-out the branch `sa_gym` after the second `git clone`

saudet: What about A3CCartpole? Does it work even without the synchronized ACPolicy workaround?

saudet: I've tested it on Windows, and it builds fine for me with `mvn -U clean install -Dmaven.test.skip`. Interestingly, A3CCartpole also runs fine without doing anything special, so let's get this merged, and figure out later what's wrong with DL4J/ND4J on Linux.

raver119 (approved):

PR #8596, opened 2020-01-07
RL4J Changes to frame skipping logic

aboulang2002:

In this PR, the frame skipping logic is removed from `QLearning`, `Policy` and `AsyncThread`. The learning classes shouldn't be the ones deciding when an observation should be skipped. There should be an observation handler for that but in the mean time, the logic has been merged/moved to `LegacyMDPWrapper`.

Note: With this PR, a skipped observation still contains data. This, and few other related issues are not trivial enough to be fixed in this PR.

## How was this patch tested?

unit tests and manual tests

## Quick checklist


aboulang2002:
Last commit:
 - Changed all `refacInitMdp()` methods to use `Observation.isSkipped()`
 - Fixed the problem where observation added to the history would be incorrect after init (e.g. `[0, 2, 4, 6, 8, 9, 11, 13, ...]` instead of `[0, 2, 4, 6, 8, 10, ...]`)
 - Fixed a problem in `QLearningDiscrete` where a skipped observation would be used as the `nextObservation` of the `Transition`
 - Fixed step counting: `stepCounter` is absolute total number of steps (i.e. not reset between epochs) and `currentEpochStep` is the number of steps since the start of the epoch (maybe it would be a good idea to rename `stepCounter` and `StepCountable` to something more meaningful)

saudet:
Please update with master! Thanks

aboulang2002:
Done!

rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/util/LegacyMDPWrapper.java
```diff
@@ -32,9 +32,12 @@ public LegacyMDPWrapper(MDP<O, A, AS> wrappedMDP, IHistoryProcessor historyProce
     }
 
     private IHistoryProcessor getHistoryProcessor() {
-        return historyProcessor;
+        if(historyProcessor != null) {
+            return historyProcessor;
+        }
+        
+        return null;
```
saudet: Maybe we don't need to do this? :) In any case, please make sure that A3CCartpole isn't crashing!

aboulang2002: A3CCartpole seems to be running without crashing. I tried it several times because with async, it may be a 'timing' problem that is hard to reproduce. But I think we're fine.

On the other hand, I just noticed that A3CCartpole is not learning. Please do not merge, I will investigate what is wrong.

saudet: A3CCartpole never really learned that much. It's quite hard to tweak, so don't worry about it...

saudet: Please simplify (again) the getHistoryProcessor() method though.

aboulang2002:
Last commit: I have changed the `getHistoryProcessor()` to a `@Getter`. This is only temporary because the `HistoryProcessor` will soon be completely removed from the `LegacyMDPWrapper`.

I can reproduce the NaN problem on Windows relatively easily now with the snapshot builds.
At first, I thought that it might be a case of exploding gradients (cartpole observations are not normalized (A PreProcessor will handle this in the near future)). But changing the timing (adding `log.info()` or setting breakpoints) makes it much harder to reproduce. So it is very likely to be a synchronization problem. I suspect that the `synchronized()` in `ACPolicy` may have just made the problem harder to reproduce and not fixed it; they are notoriously slow and would definitely change the timing.

In any ways, that doesn't prevent us from merging this PR.

rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/qlearning/QLearning.java
```diff
@@ -136,49 +139,47 @@ protected StatEntry trainEpoch() {
...
             obs = stepR.getStepReply().getObservation();
             incrementStep();
-            step++;
         }
 
         meanQ /= (numQ + 0.001); //avoid div zero
 
 
-        StatEntry statEntry = new QLStatEntry(getStepCounter(), getEpochCounter(), reward, step, scores,
+        StatEntry statEntry = new QLStatEntry(getStepCounter(), getCurrentEpochStep(), reward, currentEpochStep, scores,
```
saudet: Were those two always effectively equal?

saudet: I mean, it looks like the number of epochs and the number of steps have been confounded here and in the rest of the changes. Could you make sure this isn't the case?

aboulang2002: You're right, this is wrong. There are other places where I mixed epoch step and epoch number. I'll fix this asap. Really sorry!

aboulang2002: Done.

aboulang2002:
Last commit: Fixed getEpochCounter() incorrectly changed to getCurrentEpochStep() calls.
Sorry again for the mixup.

rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/sync/qlearning/QLearning.java
saudet: Do you think we still need StepCountable.java? Or can we replace it with the new interfaces?

aboulang2002: That `StepCountable`/`stepCounter` is bothering me too. The way steps are handled is a mess. I think that in the async case, it's not even the right count. Presently, it's the total number of steps per thread (not reset between episodes but not shared between threads). But in literature, the closest concept is the number of iterations which is the grand total of all steps of all episodes taken by all agents. 

We shouldn't even need an interface. I think this problem is part of a bigger one. The learning classes and `Policy.play()` are not simply doing something similar; they all have their separate implementation of what is an agent. And this agent concept is missing in RL4J.

I am still thinking on the details but there should be a separate Agent class. The Policy shouldn't be used to `play()` and the learning classes should contain an agent instead of being one. And specific behaviors should be injected instead of being implemented through inheritance.

aboulang2002: I have removed StepCountable. You are right, lets remove it now since it's not used.

thomas-trendsoft:
Vielen Dank für ihre Mail. Leider bin ich bis zum 05.02. aus gesundheitlichen Gründen nicht im Büro, 

Bei dringenden Angelegenheiten rufen Sie bitte meine Kollegen im Büro unter:

0591 - 97778468 an.

Mit freundlichem Gruß

Thomas Krieger

saudet:
With these commits, I'm not able to train successfully the Cartpole example. I'm afraid there's something somewhere in the logic that changed... @aboulang2002 Could you fix this before we merge?

aboulang2002:
Great find @saudet! There was actually two issues. The first problem was that the last transition was discarded if the episode ended with `mdp.isDone()`. In the case of cartpole, when `mdp.isDone()` is the only transition when the reward is not 1... The only transition that mattered was discarded!
The other problem is in `CartpoleNative` where `mdp.isDone()` was always false. I use `CartpoleNative` for testing sync learning so I never got the first problem.

Again thanks for finding this!

Last commit:
Fixed `CartpoleNative`.
Updated `QLearningDiscreteTest` to check the experience replay memory after the test.
Added the method `finishEpisode()` in `QLearning`, called after the episode finishes.
Fixed the problem with the last transition being dropped in `QLearningDiscrete`.

saudet:
Hum, I'm still not able to train Cartpole successfully. There is something else broken in there...

saudet:
Actually, no, it seems to be fine. It just looks like sequences of random numbers changed a bit somewhere. I am able to train successfully, and although there are some random sequences that fail to train, that's nothing new.

saudet (approved):

PR #8636, opened 2020-01-24
RL4J - Added ObservationHandler

aboulang2002:

What we need now is a way to build the Observations. In this PR, I add a few classes that I'll need. 
Note: They're not connected to the rest of RL4J in this PR in order to stay below the 1000 lines limit.

What we'll need ultimately is an Observation able to contain several data 'channels'.
A use case of this would be an observation that contains a video channel with a left and right audio channels and a LIDAR channel.
Another use case would be a multi-player environment where each player's observation would be a channel with some shared observation as an additional channel.

For now, I am adding a single-channel `ObservationHandler` class, to be used in the environment to build and process the observations.
It's usage looks something like this:
```
Observation observation = observationHandler.newObservation()
      .addChannelData(legacyChannelDataFactory.create(rawData))
      .isFinalObservation(isFinalObservation)
      .build();
```
The `ObservationHandler` will handle the raw data this way: Recording -> Filtering -> Processing -> Assembly.

During the recording step, the `DataChannels` are passed to instances of `DataRecorder` that can record the data. For example, a LidarRecorder could render a point cloud channel to a video, while another recorder could create a video from the video and audio channels.

The filtering step determines if an observation should be skipped or not. With this, different approaches can be used, for example a 3/4 frame skipping, or a random frame skipping. This can also be used to skip the first few frames of an episode if needed.

The processing step is where normalization, cropping, resizing, RGBtoGrayscale, etc... is done

And finally the assembly step, where the final packaging of the observation is done. Currently, this step is not fully implemented and is just calling `HistoryProcessor.getHistory()` (if set). It will need to change to support other ways to track motion. For example, instead of stacking the last X observations, one can flatten the stack by using the max of corresponding pixels in the stack or adding (or subtracting) the corresponding pixel values of the frames. 

## How was this patch tested?

unit tests and manual tests

## Quick checklist

### DIFF
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/HistoryProcessor.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/HistoryProcessor.java
@@ -107,6 +107,13 @@ public class HistoryProcessor implements IHistoryProcessor {
         return array;
     }
 
+    public boolean isHistoryReady() {
+        return history.size() == conf.getHistoryLength();
+    }
+
+    public void reset() {
+        history.clear();
+    }
 
     private INDArray transform(INDArray raw) {
         long[] shape = raw.shape();
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/IHistoryProcessor.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/learning/IHistoryProcessor.java
@@ -46,6 +46,9 @@ public interface IHistoryProcessor {
 
     boolean isMonitoring();
 
+    boolean isHistoryReady();
+    void reset();
+
     /** Returns the scale of the arrays returned by {@link #getHistory()}, typically 255. */
     double getScale();
 
--- a/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/observation/Observation.java
+++ b/rl4j/rl4j-core/src/main/java/org/deeplearning4j/rl4j/observation/Observation.java
@@ -16,22 +16,36 @@
 
 package org.deeplearning4j.rl4j.observation;
 
+import lombok.Getter;
+import lombok.Setter;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.dataset.api.DataSet;
 import org.nd4j.linalg.factory.Nd4j;
 
 /**
- * Presently only a dummy container. Will contain observation channels when done.
+ * Represent an observation from the environment
+ *
+ * @author Alexandre Boulanger
  */
 public class Observation {
-    // TODO: Presently only a dummy container. Will contain observation channels when done.
+
+    /**
+     * A singleton representing a skipped observation
+     */
+    public static Observation SkippedObservation = new Observation((DataSet) null);
 
     private final DataSet data;
 
+    // FIXME: only use DataSet.IsEmpty instead
+    @Getter @Setter
+    private boolean skipped;
+
+    // FIXME: to be removed once ObservationHandler is plugged
     public Observation(INDArray[] data) {
         this(data, false);
     }
 
+    // FIXME: to be removed once ObservationHandler is plugged
     public Observation(INDArray[] data, boolean shouldReshape) {
         INDArray features = Nd4j.concat(0, data);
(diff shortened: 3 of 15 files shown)


saudet:
Would using DataVec to create pipelines with ImageTransform make sense here? 
https://github.com/eclipse/deeplearning4j/blob/master/datavec/datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ImageTransform.java
It might not be general enough, but is there a reason why we shouldn't also extend DataVec?

aboulang2002:
I am unfamiliar with DataVec but it looks like it's exactly what we need for image processing. And as you say, we could even extend DataVec for audio processing and other signal processing. 
I also see that there is a `SchemaBuilder` and a `TransformProcess` in DataVec. Do you think we could use these when we'll be ready to support more than one data channel? These classes seem to expect scalar data for the columns and we need something that can get any shape as input.

saudet:
The API around `Schema` and `TransformProcess` actually came from code that was developed independently for tabular data, which was then merged into DataVec at a later time:
https://deeplearning4j.org/docs/latest/datavec-transforms
I don't think that would be useful to reinforcement learning, but what do you think?

Assuming we decide to extend the `ImageTransform` API though, is there anything in this PR that you would do differently? If not, please update your fork, and I'll proceed to a review for merging!

aboulang2002:
The more I think about this PR and look at DataVec, the more I think this PR going the wrong way. We can't use directly DataVec's `Schema` or `TransformProcess`, but I like how it's done there. RL4J has a similar requirement: bring the raw data into a usable form (for a Neural Network). So I think it would be a good idea to implement something that looks similar in RL4J.

The environment need a way to describe the observation that it "publishes" and something inspired by `Schema` would do nicely. Also, the user must be able to define how the raw observation should be filtered and transformed, a bit like `TransformProcess`. And to answer your question directly yes, I prefer a generic `ImageTransform` way to what I was aiming for (DataSetPreProcessor).

So, I will cancel this PR, this one is not good enough.

saudet:
Ah, wait a minute, we talked about DataSetPreProcessor before in pull #7958. I remember now that @AlexDBlack was saying that we were aiming to have everything done through that API with libnd4j, but would that apply to any kind of data? What's your take on this @AlexDBlack?

aboulang2002:
I think we could support both DataVec's transforms and libnd4j's DataSetPreProcessors or even any 3rd party lib a user might want to use. The difficulty is how to handle all the different types of observation an environment can generate. I think the correct way to solve this to let the user handle it. 

For example, suppose I have a ALE environment that outputs 160x192x3 video frames in an observation channel called "video":
```
TransformProcess tp = TransformProcess.builder()
	.filter(new UniformFrameSkiping(3, 4))                            // Skip 3 frames out of every 4
	.transform("video", new ConvertToImageWritable())                 // Turn ALE's screen (BGR byte array) into DataVec's ImageWritable
	.transform("video", new ResizeImageTransform(80, 96))             // Scale down
	.transform("video", new ColorConversionTransform(COLOR_BGR2GRAY)) // Turn to grayscale
	.transform("video", new ConvertImageWritableToDataSet())          // Convert ImageWritable to DataSet
	.preProcess("video", new MyDataSetPreProcesor())                  // The channel must be a DataSet to use libnd4j's DataSetPreProcessors
	.transform("video", new HistoricalStacking(5))                    // Stack the last 5 non-skipped frames
	.build();
```
The operations are applied in the same order as they are added to the builder. The user can do pretty much what he wants. At the end of the TransformProcess, we only require that either:
The observation is skipped
or
All observation channels are DataSets

What do you think?

/cc @Bam4d

AlexDBlack:
At present we've got 2 options for image type preprocessing ops
datavec and libnd4j/nd4j/samediff (we've added a bunch to libnd4j/nd4j recently, but still not as substantial for image augmentation as datavec image stuff, but then there's other stuff that nd4j/libnd4j has that datavec doesn't)
without diving into the details here, in the near term, adding a layer of abstraction that allows use of either/both types of ops (datavec and nd4j/samediff) sounds good to me.

saudet:
@AlexDBlack Sounds like a plan!

@aboulang2002 So let's keep working on the `Observation` like you intended to do it at first, but as an interface to other stuff underneath. We can still extend and modify DataVec if it makes it easier though, as long as we keep backward compatibility, for now anyway.

Issue #8779, opened 2020-03-12
Arbiter UI NPE


eraly:
#### Issue Description

NPE in the arbiter basic example 
```
Exception in thread "main" java.lang.NullPointerException
	at org.deeplearning4j.ui.VertxUIServer.stop(VertxUIServer.java:306)
	at org.deeplearning4j.arbiterexamples.quickstart.BasicHyperparameterOptimizationExample.main(BasicHyperparameterOptimizationExample.java:179)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
```

#### Version Information
beta6, macos

AlexDBlack:
So far I've been unable to reproduce this (5 runs through on master, CPU backend).
Was this on current master also?

eraly:
This was on beta6. I can try with snapshots and see if it is still an issue.

eraly:
Still seeing this on snapshots:

```
Exception in thread "main" java.lang.NullPointerException
	at org.deeplearning4j.ui.VertxUIServer.stop(VertxUIServer.java:306)
	at org.deeplearning4j.arbiterexamples.quickstart.BasicHyperparameterOptimizationExample.main(BasicHyperparameterOptimizationExample.java:179)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
```

AlexDBlack:
thanks. Though based on the line numbers, that doesn't make a whole lot of sense, as it would require the server field to be null, which it can't be

[...]

PR #8856, opened 2020-04-20
Fix UIServer features in multi-session mode, synchronous start and stop

### LINKED ISSUE #8725: UI in snapshot is broken
#### Issue Description

When creating UI, it should be available, and
` Deeplearning4j UI server started at: http://localhost:9000 `
should be in the log.

In current snapshot, UI is not available, and log entry is not generated.

#### Version Information

Current snapshot.

#### Additional Information

`
2020-02-22 10:51:41.190 - SEVERE: The FreeMarker version requested by "incompatibleImprovements" was 2.3.29, but the installed FreeMarker version is only 2.3.23. You may need to upgrade FreeMarker in your project.
java.lang.IllegalArgumentException: The FreeMarker version requested by "incompatibleImprovements" was 2.3.29, but the installed FreeMarker version is only 2.3.23. You may need to upgrade FreeMarker in your project.
        at freemarker.template._TemplateAPI.checkVersionNotNullAndSupported(_TemplateAPI.java:49)
        at freemarker.core.Configurable.<init>(Configurable.java:309)
        at freemarker.template.Configuration.<init>(Configuration.java:634)
        at org.deeplearning4j.ui.module.train.TrainModule.<init>(TrainModule.java:140)
        at org.deeplearning4j.ui.VertxUIServer.start(VertxUIServer.java:197)
        at io.vertx.core.AbstractVerticle.start(AbstractVerticle.java:106)
        at io.vertx.core.Verticle.start(Verticle.java:66)
        at io.vertx.core.impl.DeploymentManager.lambda$doDeploy$8(DeploymentManager.java:556)
        at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:369)
        at io.vertx.core.impl.EventLoopContext.lambda$executeAsync$0(EventLoopContext.java:38)
        at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
        at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:510)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:518)
        at io.netty.util.concurrent.SingleThreadEventExecutor$6.run(SingleThreadEventExecutor.java:1044)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:834)
`
Forcing FreeMarker version to 2.3.29 in application pom.xml seem to be workaround.

### LINKED ISSUE #8794: DL4J UI - System information not loading
Hi DL4J team.

With both beta6 and SNAPSHOT versions, the 'System' tab on the DL4J Training UI page appears to be broken - none of the labels progress from the 'Loading...' state.

image

The network tab in Chrome Dev Tools reveals that it's trying to load resources and getting a 404 response:

image

image

We are using multi-session. The rest of the UI seems to work fine.

printomi:

fix Freemarker version mismatch: change version requested in `TrainModule` to 2.3.23 (freemarker.version in pom.xml) (fixes #8725)
fix checking multi-session mode in `VertxUIServer.getInstance`. Tested multiple calls in `TestVertxUIMultiSession`.
fix `UIServer.getInstance()` to return existing instance
extend timeout for manual UI tests from 30 to 600 seconds
fix for auto-attaching `StatsStorage` given in `VertxUIServer.getInstance(Integer, boolean, Function<String,StatsStorage>)`, test improvements
start and stop UI server synchronously (wait until complete), tests (should fix issue #8779)
exception handling, test improvements
add asynchronous method to start UI server
fix UI server language setting in multi-session mode
fix UI server system tab not loading data in multi-session mode (fixes #8794)
fix async stopping of UIServer.stopAsync(Promise<Void>), added test
restore the daemon thread style behaviour of UIServer: don't keep the process alive just because the UI is running 
speed up and don't Ignore tests in TestVertxUI and TestVertxUIMultiSession, put longer tests to separate class TestVertxUIManual

## How was this patch tested?

Tests in `TestVertxUI` and `TestVertxUIMultiSession`, manual tests in `TestVertxUIManual`

## Quick checklist

### DIFF
--- a/deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-vertx/src/main/java/org/deeplearning4j/ui/VertxUIServer.java
+++ b/deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-vertx/src/main/java/org/deeplearning4j/ui/VertxUIServer.java
@@ -19,6 +19,8 @@ package org.deeplearning4j.ui;
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
 import io.vertx.core.AbstractVerticle;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.impl.MimeMapping;
@@ -75,24 +77,15 @@ public class VertxUIServer extends AbstractVerticle implements UIServer {
 
     private TrainModule trainModule;
 
-    public static VertxUIServer getInstance(Integer port, boolean multiSession, Function<String, StatsStorage> statsStorageProvider){
+    public static VertxUIServer getInstance(Integer port, boolean multiSession,
+                                            Function<String, StatsStorage> statsStorageProvider)
+            throws RuntimeException, InterruptedException {
         if (instance == null || instance.isStopped()) {
             VertxUIServer.multiSession.set(multiSession);
             VertxUIServer.setStatsStorageProvider(statsStorageProvider);
             instancePort = port;
-            Vertx vertx = Vertx.vertx();
-
             //Launch UI server verticle and wait for it to start
-            CountDownLatch l = new CountDownLatch(1);
-            vertx.deployVerticle(VertxUIServer.class.getName(), res -> {
-                if (res.failed()) {
-                    log.error("UI server failed to start.", res.cause());
-                }
-                l.countDown();
-            });
-            try {
-                l.await(5000, TimeUnit.MILLISECONDS);
-            } catch (InterruptedException e){ } //Ignore
+            deploy();
         } else if (!instance.isStopped()) {
             if (multiSession && !instance.isMultiSession()) {
                 throw new RuntimeException("Cannot return multi-session instance." +
@@ -106,6 +99,50 @@ public class VertxUIServer extends AbstractVerticle implements UIServer {
         }
 
         return instance;
+    }
+
+    /**
+     * Deploy (start) {@link VertxUIServer}, waiting until starting is complete.
+     * @throws RuntimeException if UI server failed to start
+     * @throws InterruptedException if interrupted while waiting for completion
+     */
+    private static void deploy() throws RuntimeException, InterruptedException {
+        CountDownLatch l = new CountDownLatch(1);
+        Promise<String> promise = Promise.promise();
+        promise.future().compose(
+                success -> Future.future(prom -> l.countDown()),
+                failure -> Future.future(prom -> l.countDown())
+        );
+        deploy(promise);
+        // synchronous function
+        l.await();
+
+        Future<String> future = promise.future();
+        if (future.failed()) {
+            throw new RuntimeException("Deeplearning4j UI server failed to start.", future.cause());
+        }
+    }
+
+    /**
+     * Deploy (start) {@link VertxUIServer},
+     * and handle successful or failed completion of deployment with given promise
+     * @param startCallback promise that will handle success or failure of deployment
+     */
+    public static void deploy(Promise<String> startCallback) {
+        log.debug("Deeplearning4j UI server is starting.");
+        Promise<String> promise = Promise.promise();
+        promise.future().compose(
+                success -> Future.future(prom -> {prom.complete();}),
+                failure -> Future.future(prom -> {
+                    prom.fail(new RuntimeException(failure));
+                })
+        ).compose(
+                success -> Future.future(prom -> { startCallback.complete(); }),
+                failure -> Future.future(prom -> { startCallback.fail(failure); })
+        );
+
+        Vertx vertx = Vertx.vertx();
+        vertx.deployVerticle(VertxUIServer.class.getName(), res -> {promise.handle(res);});
     }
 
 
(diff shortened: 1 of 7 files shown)


printomi:
@AlexDBlack please, review my changes.

I have signed off the commits. [edit]

printomi:
I have signed off the commits.

printomi:
I have run all the examples using `UIServer`, without error. 

I note that most [examples](https://github.com/eclipse/deeplearning4j-examples) don't stop the server, so VertxUIServer continues to run after the example is complete. We don't have to use `Thread.sleep(...)` any more, because the server keeps running even after the main thread exits. 
In contrast, PlayUIServer stopped when we exited main in 1.0.0-beta5.
Shall we return to the 1.0.0-beta5 behavior of stopping server when main exits?

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-vertx/src/main/java/org/deeplearning4j/ui/VertxUIServer.java
```diff
@@ -302,11 +401,28 @@ private void modulesViaServiceLoader(List<UIModule> uiModules) {
...
+        // synchronous function should wait until the server is stopped
+        l.await();
     }
 
+    @Override
+    public void stop(Promise<Void> stopCallback) {
+        shutdown.set(true);
+        stopCallback.complete();
+        log.info("Deeplearning4j UI server stopped.");
+    }
```
AlexDBlack, line 445: This method looks off to me.
If I (as a user) call `stop()` - everything is fine. If I call `stop(Promise<Void>)` - the vertx instance isn't actually stopped?

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-vertx/src/test/java/org/deeplearning4j/ui/TestVertxUIMultiSession.java
```diff
@@ -54,19 +57,25 @@
  */
 @Ignore
 public class TestVertxUIMultiSession extends BaseDL4JTest {
+    private static final Logger log = LoggerFactory.getLogger(TestVertxUIMultiSession.class);
```
AlexDBlack: Not a big deal - but we can use Lombok `@Slf4j` annotation on the class to get the same thing

AlexDBlack (commented):
Just one question, otherwise looks great, thanks

AlexDBlack:
@printomi 
> I note that most examples don't stop the server, so VertxUIServer continues to run after the example is complete. We don't have to use Thread.sleep(...) any more, because the server keeps running even after the main thread exits.
In contrast, PlayUIServer stopped when we exited main in 1.0.0-beta5.
Shall we return to the 1.0.0-beta5 behavior of stopping server when main exits?

Good question, pros and cons to both options.
I think on balance it'll be better to exit - users could be surprised by the process not exiting after training completes - which is worse (and maybe harder to realize it's the UI at fault?) than an early exit. So yes, if possible let's return to the daemon thread style behaviour (i.e., don't keep the process alive just because the UI is running.

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-vertx/src/main/java/org/deeplearning4j/ui/VertxUIServer.java
printomi: Well, I did not test this scenario, i.e. calling directly `UIServer.stop(Promise<Void>)`. This method is called by `vertx.close()`.
The documentation of `Verticle.stop(Promise<Void>)` tells that:
> Vert.x calls this method when un-deploying the instance. You do not call it yourself.

Adding a separate `UIServer.stopAsync(Promise<Void>)` function and removing `UIServer.stopAsync(Promise<Void>)` would clarify this.

printomi:
> So yes, if possible let's return to the daemon thread style behaviour (i.e., don't keep the process alive just because the UI is running.

I agree, I will work on restoring the daemon thread style behaviour.

printomi:
@AlexDBlack I would like to remove the `@Ignore` annotation from `TestVertxUI` and `TestVertxUIMultiSession` classes, so that they get tested automatically. 
These contain a handful of tests that can be run within 30 seconds specified in `BaseDL4JTest`. I would move the manual tests into a separate class.
Do you agree?

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-vertx/src/main/java/org/deeplearning4j/ui/VertxUIServer.java
printomi: I fixed async stopping of UIServer.

deeplearning4j/deeplearning4j-ui-parent/deeplearning4j-vertx/src/test/java/org/deeplearning4j/ui/TestVertxUIMultiSession.java
printomi: Now using Lombok `@Slf4j` annotation.

printomi:
Changes:
- I fixed async stopping of UIServer.
- I restored the daemon thread style behaviour of UIServer: don't keep the process alive just because the UI is running.
- Additionally, I improved tests: speed up and don't `@Ignore` tests in TestVertxUI and TestVertxUIMultiSession, put longer tests to separate class TestVertxUIManual.
@AlexDBlack I hope, you agree with these changes.

AlexDBlack (approved):
LGTM, thanks! :+1:

Issue #8712, opened 2020-02-16
WordVectorSerializer calls Class.forName without a way to specify classloader


dominics:
WordVectorSerializer attempts to dynamically load a tokenizer factory configured by the user, via `Class.forName`: https://github.com/eclipse/deeplearning4j/blob/4b46aaedd3cc4bcffc878cb75ea2c558ccdf99d9/deeplearning4j/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/models/embeddings/loader/WordVectorSerializer.java#L2698

But when it does, it hard-codes the use of the default classloader. This assumes that all classes are available from a single classloader, which is not a good assumption for a library to make when it doesn't know the structure of the application.

This assumption fails, for example with [Play Framework](https://github.com/playframework/playframework/issues/4159)'s dev mode - Play uses a dynamic classloader for application classes, so they can be hot reloaded. (There's some other evidence that the Play team might be right in asking libraries to fix the problem: see also comments like https://issues.apache.org/jira/browse/LOG4J2-604: "The idiom `Class.forName` is almost always a bad idea if it's called without a classloader to go along with it.")

Finally, there are no exception details included in the `log.error` a few lines later, and the exception is swallowed after that log, so users need to breakpoint to find the exception message.

agibsonccc:
Closing — references a removed/rewritten subsystem (DataVec, NLP/Word2Vec, Model Zoo, Spark, UI, or Arbiter). These have been fundamentally rewritten or removed in the upcoming release. Please open a new issue if you experience similar problems with the new version.

PR #8972, opened 2020-05-21
FEATURE: possibility to define global Classload for ND4J

hosuaby:
This PR is related to https://github.com/eclipse/deeplearning4j/issues/8712. This issue is talking about the need to have a possibility to provide class loader for method `Class.forName`. Author mention only class `WordVectorSerializer` from `deeplearning-nlp`. But the problem is more general. We need to have possibility to chose classloader used by ND4J and Deeplearning4j.

This PR is a first part of the work. I added a new class `ND4JClassLoading` that gives possibility to used to provide his own classloader used by `Class.forName` and `ServiceLoader.load`. And I replaced all call to those methods by calls to this new class. I hope I forgot nothing.

There will be a second part of this work in anouther PR, concerning Dl4j. Both PR will be inderpendent.


agibsonccc:
@hosuaby hey sorry about this. I'm just getting to some of these, do you mind updating? I understand if you can't, otherwise we can get this merged.

agibsonccc:
Hey, sorry this is late. Could you modify it to use the nd4j class loader? Otherwise, this seems legit. Sorry for the late response!

hosuaby:
Hello, @agibsonccc :)
Thanks for review. First, I made a rebase of branch and solved conflicts with current master.
I don't understood your demend. If need any modification, don't hesitate to ask. What exactly should I modify ?

hosuaby:
Hello, @agibsonccc .

I think I undestood what you wanna say. My mistake, I have better to explain my plan.

This PR is the first part of work concerning https://github.com/eclipse/deeplearning4j/issues/8712. It provides ability to define classloader for ND4J. For DL4J I would like to have a separate class `DL4JClassLoading`, who will behaivour the same way, but for Dl4j packages.

The reasons I wanna to keep two classloaders separate: 
1) Nd4j and Dl4j are semantically two distinct projects
2) In my use case, I only need to redefine classloader for ND4j.
3) If I decide to rely on `ND4JClassLoading` from dl4j packages, it's will not be clear for user why should he define classloader for nd4j instead of dl4j.

So, my plan is, if this PR is approved, I make immediately the second one with `DL4JClassLoading` class.

Tell me what you think about.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4jBackend.java
```diff
@@ -156,14 +157,12 @@ public static Nd4jBackend load() throws NoAvailableBackendException {
         String logInitProperty = System.getProperty(ND4JSystemProperties.LOG_INITIALIZATION, "true");
         boolean logInit = Boolean.parseBoolean(logInitProperty);
 
-        List<Nd4jBackend> backends = new ArrayList<>(1);
-        ServiceLoader<Nd4jBackend> loader = ServiceLoader.load(Nd4jBackend.class);
+        List<Nd4jBackend> backends = new ArrayList<>();
+        ServiceLoader<Nd4jBackend> loader = ND4JClassLoading.loadService(Nd4jBackend.class);
```
agibsonccc, line 162: @hosuaby you might want to also change the loadLibrary call in Nd4jBackend. Could you go through and look through all the uses of classloaders and make sure whatever you do is propagated so the classloader takes effect everywhere?

agibsonccc (changes requested):

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4jBackend.java
hosuaby: Done. PR updated and rebased on master

hosuaby:
Hello, @agibsonccc 
Any updates about this PR ?

agibsonccc (changes requested):
Minor nit

agibsonccc:
@hosuaby minor nit then I think we can merge this. Thanks!

hosuaby:
@agibsonccc Ok. Do not hesitate to ask any change!

agibsonccc:
@hosuaby  think you could get this over the line? I think we can merge this soon.

hosuaby:
@agibsonccc done

agibsonccc (approved):

agibsonccc:
Great thanks @hosuaby  I want to ping @saudet for 1 review here. @saudet  do you see any major issues here?

saudet:
I didn't test it, but it looks fine.

hosuaby:
@agibsonccc Thanks!
Now I will take care of #8712 that is very similar.

PR #9114, opened 2020-10-25
FIX: ND4J tests

hosuaby:
Signed-off-by: hosuaby <alexei.klenin@gmail.com>

Fixes test in ND4J project.
Current state:
- tests on CPU backend run and pass
- tests on CUDA run but fail (probably bug)

## How was this patch tested?

Tests that was not running, are running and pass now.

## Quick checklist


hosuaby:
@agibsonccc I allow myself to ask you a question, while I make this PR.

I am looking to implement a pure JVM backend for ND4J, using one of many Linear Algebra libraries. Java is pretty fast those days, so using a right library that is able to store arrays off-heap, I think, it's possible to build a portable and performant backend. 

But I need to validate this backend using some tests. I suppose, the project `nd4j-tests` is what I need. I just to need add my backend on the classpath of this projet and run the tests. Am I right ?

agibsonccc (approved):

agibsonccc:
@hosuaby we would accept a PR for that  but would need to work on the maintenance. Our main use case is definitely c/c++. We had a java backend at one point, but no matter what we did the performance just wasn't there compared to c/c++. If you want, I would suggest running some benchmarks seeing if you can beat c/c++ or if we do knowingly spend time on that, we accept a performance penalty.

hosuaby:
@agibsonccc thanks. I have no intention to push that backend as a part of DL4J. I was thinking to keep it as separate project. I just wanna know, how do you test your backends to assert that I implemented everything right.
If you have a benchmark, I will be glad to try it.

My work concerns integration of machine learning models into Elasticsearch. Currently, I made a plugin that comes with DL4j et a lot of native code from C++ backend. It creates serious security concerns. And I don't really need it. I am not doing learning within Elasticsearch, only run pre trained models. So I don't need the best performance, but acceptable one.

agibsonccc:
@hosuaby generally, for nd4j tests all we do is make sure that every op outputs the same results. If you do that, then the rest falls in to place. 

Some things to consider here is in java, you'll only be able to support certain data types and a lot of math operations will break. In c++, we have quite a few data types outside of fp32 (fp16 comes to mind here) that you can't really do much with as is in java.


As for embedding nd4j in to elastic search, I would look at what the solr guys did there.: https://github.com/eclipse/deeplearning4j/pull/5429

PR #9126, opened 2020-11-15
Fixed #9125, make check agains the correct variable

### LINKED ISSUE #9125: SD_FORBID_HELPERS not doing anything
#### Issue Description

Ran into some problem with DNN on an old AMD Turion processor so wanted to test without the native mkldnn helper. Tried various way to set SD_FORBID_HELPERS to disable the helper but couldn't get it to work.

#### Version Information

Running deeplearning4j-1.0.0-beta7

#### Additional Information

I found in libnd4j/include/legacy/impl/Environment.cpp what looks like a typo in the check. It pulls in the environment variable into a local variable and then performs a check against another local variable. 

```
        /**
         * If this env var is defined - we'll disallow use of platform-specific helpers (mkldnn, cudnn, etc)
         */
        const char* forbid_helpers = std::getenv("SD_FORBID_HELPERS");
        if (max_master_threads != nullptr) {
            _allowHelpers = false;
        }

```

The forbid_helpers variable is not checked anywhere else in the code.

Ulrhol:

Check **forbid_helpers** instead of **max_master_threads** when SD_FORBID_HELPERS env var is set.

### DIFF
--- a/libnd4j/include/legacy/impl/Environment.cpp
+++ b/libnd4j/include/legacy/impl/Environment.cpp
@@ -111,7 +111,7 @@ namespace sd {
          * If this env var is defined - we'll disallow use of platform-specific helpers (mkldnn, cudnn, etc)
          */
         const char* forbid_helpers = std::getenv("SD_FORBID_HELPERS");
-        if (max_master_threads != nullptr) {
+        if (forbid_helpers != nullptr) {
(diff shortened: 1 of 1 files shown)


Ulrhol:
Can't get the commit signoff to work. Closing this one and make a clean attempt

PR #9186, opened 2021-02-19
Ada belief

quickwritereader:

AdaBelief updater: it was agreed to modify changes on the copy of AdamUpdater. This way we can improve it later.

## How was this patch tested?

unit test

## Quick checklist

### DIFF
--- a/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/config/AdaBelief.java
+++ b/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/config/AdaBelief.java
@@ -40,7 +40,7 @@ import java.util.Map;
 public class AdaBelief implements IUpdater {
 
     public static final double DEFAULT_LEARNING_RATE = 1e-3;
-    public static final double DEFAULT_EPSILON = 1e-8;
+    public static final double DEFAULT_EPSILON = 1e-14;
     public static final double DEFAULT_BETA1_MEAN_DECAY = 0.9;
     public static final double DEFAULT_BETA2_VAR_DECAY = 0.999;
 
--- a/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/config/AdaBelief.java
+++ b/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/config/AdaBelief.java
@@ -1,0 +1,132 @@
+/*
+ *  ******************************************************************************
+ *  *
+ *  *
+ *  * This program and the accompanying materials are made available under the
+ *  * terms of the Apache License, Version 2.0 which is available at
+ *  * https://www.apache.org/licenses/LICENSE-2.0.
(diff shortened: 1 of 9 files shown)


quickwritereader:
I just added one test case in cpp:
here is the python code that was used to get test numbers:
```
import numpy as np
alpha=0.001
beta1=0.9
beta2=0.999
epsilon=1.e-8
#https://arxiv.org/pdf/2010.07468.pdf
def update( t, w, gradW,   mt, st):
        mt =  beta1* mt + (1- beta1)*gradW 
        st =  beta2* st + (1- beta2)*((gradW-mt)**2) + epsilon
        mt_corr =  mt/(1- beta1**t) 
        st_corr =  st/(1- beta2**t)
        upW= alpha*(mt_corr/(np.sqrt(st_corr)+epsilon)) 
        w = w -  upW 
        return ( w, upW,  mt, st )
#if you want to test with more precision np.set_printoptions(precision=9)
grad = np.array([1,2,3,4,5], dtype = np.float32)
w=np.zeros(5, dtype = np.float32)
mt=np.zeros(5, dtype = np.float32)
st = np.zeros(5, dtype = np.float32)
for t in range(1,4):
    w, upW,    mt, st = update(t,w,grad, mt,st )
    print(f"---{t}----")
    print(f"update {upW}")
    print(f" s state {st} ")    
    print(f" m state {mt} ")
```

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/config/AdaBelief.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * AdaBelief 
+ * https://arxiv.org/pdf/2010.07468.pdf
+ */
+@Data
+@Builder(builderClassName = "Builder")
+public class AdaBelief implements IUpdater {
+
+    public static final double DEFAULT_LEARNING_RATE = 1e-3;
+    public static final double DEFAULT_EPSILON = 1e-8;
```
treo: The paper authors have found that a smaller default epsilon works better:

> Epsilon in AdaBelief is different from Adam (typically eps_adabelief = eps_adam*eps_adam)
> ( eps of Adam in Tensorflow is 1e-7, in PyTorch is 1e-8, need to consider this when use AdaBelief in Tensorflow)

So I guess we might want to use either 1e-14 or 1e-16 here.

treo:
Other than the epsilon thing, it looks good to me. 

I've only read through the code though and didn't run it. But the test looks reasonable.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/config/AdaBelief.java
quickwritereader: changed to 1e-14

quickwritereader:
> Other than the epsilon thing, it looks good to me.
> 
> I've only read through the code though and didn't run it. But the test looks reasonable.

let's not rush then. it needs to be checked again properly.

agibsonccc (approved):

PR #9286, opened 2021-05-02
Update LayerValidation.java

agibsonccc:
Updates LayerValidation to avoid ConcurrentModificationException by 
copying the results and setting the new result on the layer conf instead.

## How was this patch tested?

## Quick checklist

- [ ] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.org/eclipse-contributors) page for details
- [ ] Reviewed the [Contributing Guidelines](https://github.com/eclipse/deeplearning4j/blob/master/CONTRIBUTING.md) and followed the steps within.

### DIFF
--- a/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
+++ b/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java
@@ -1155,6 +1155,7 @@ public class NeuralNetConfiguration implements Serializable, Cloneable {
                     sl.setConvolutionMode(convolutionMode);
                 }
             }
+
             LayerValidation.generalValidation(layerName, layer, idropOut, regularization, regularizationBias,
                     allParamConstraints, weightConstraints, biasConstraints);
         }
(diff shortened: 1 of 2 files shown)


treo (changes requested):
While this solves the symptom, I'm not sure it is the actual solution to the problem.

The situation when the list being iterated over being modified by that code should only ever happen when they are identical, and if they are identical, that entire part is not necessary. 

I think you need to dig deeper to understand why the situation happens in the first place, and then fix the root cause.

agibsonccc:
@treo  the change was introduced here: https://github.com/eclipse/deeplearning4j/pull/9252/commits/819f3b8c9d5377ed8c3031b4c519f0a3c13e65d3 - the fix was connected to: https://github.com/eclipse/deeplearning4j/issues/9050 0 it's only invoked here: https://github.com/eclipse/deeplearning4j/issues/9050

It appears to be due to sharing references to the same list across layers, similar to what is done with dropout, I added a copy constructor here: https://github.com/eclipse/deeplearning4j/blob/ba10a09c19c94973b15a73a3cbf39bf75888d2ee/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/NeuralNetConfiguration.java#L1175

treo:
With the copy constructor added, I think the other "fix" you introduced earlier isn't necessary anymore.

agibsonccc:
@treo Ran this with GradientCheckTests#testGradientWeightDecay as the failing test case and it seems to work now. There's a subtle issue when using for each when modifying the regularizations. I'm going to keep those in. The original fix was for an actual real problem we had with regularizations.

treo (approved):

PR #9290, opened 2021-05-05
Nd4j versioning change ADR

agibsonccc:
ADR for version change allowing helpers to be introduced as optional enhancements for users on different platforms.
These helpers include components like armcompute, onednn, cudnn

## How was this patch tested?

## Quick checklist

### DIFF
--- a/ADRs/0007 - Nd4j classifiers.md
+++ b/ADRs/0007 - Nd4j classifiers.md
@@ -22,7 +22,7 @@ ids to the project, the following javacpp platform extensions are used:
 compiled type (avx etc or blank if normal) - software linked against (cudnn, onednn, armcompute) - version 
 
 
-An example for onednn could be:
+An example for the one dnn platform extension could be:
 dnnl-2.2
 avx256-dnnl-2.2
 
--- a/ADRs/0007 - Nd4j classifiers.md
+++ b/ADRs/0007 - Nd4j classifiers.md
@@ -18,7 +18,7 @@ in terms of performance and dependencies. This presents complexity in exposing t
 ## Decision
 
 In order to allow users to pick which configuration they would like to use, while avoiding adding a lot of different artifact
-ids to the project, the following versioning is proposed:
+ids to the project, the following versioning is used:
 compiled type (avx etc or blank if normal) - software linked against (cudnn, onednn, armcompute) - version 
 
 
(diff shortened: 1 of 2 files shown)


saudet:
I'm not sure I understand how that's better than JavaCPP extensions. We
only need to set javacpp.platform anyway with the way things are right now.

agibsonccc:
@saudet  FWIW this is why it's an ADR :) the point is to discuss the trade offs before writing code. For now, I want to flesh out a better way of handling the javacpp.platform standardization. A big (at least perceived, correct me if I'm wrong) issue is when working with javacpp.platform to reduce the number of dependencies works well, but is kind of an edge case when you blend something with extensions with something that doesn't. The goal is to make it a bit easier for people to use 1 workflow.
I like the way the presets does it where it shows the software being provided + the javacpp version that the bindings were done with. I figured piggybacking the extension on that made a bit of sense.

saudet:
I'm not aware of any "edge case" w.r.t to extensions. You'll need to explain what you're referring to.

agibsonccc:
@saudet  sure I'll be excited if this is a mis understanding. When we specify -Djavacpp.platform=windows-x86_64, does this exclude the classifiers with avx? Say: windows-x86_64-avx or something similar. If we  want to only include dependencies for windows, but also include anything that might optionally be windows avx, is that possible or does it have to be both? The main concern is uniformly being able to bundle the dependencies for platforms you want while also allowing the extensions you want. It seems like it might be difficult to specify both. (FWIW, I'll update the ADR with the discussion notes here so we can file this away under the docs and record the decision as well)

saudet:
Yes, that works just fine. Like I said, that's how it works right now, or actually we'd have to add nd4j-native-platform-avx/pom.xml, etc files that looks like the ones for GPU and MKL in the case of TF Java here: 
https://github.com/tensorflow/java/tree/master/tensorflow-core

agibsonccc:
Ok then I'll update the ADR to use extensions then. Thanks!

agibsonccc:
@saudet  updated the ADR to use extensions.

saudet (dismissed):
The decision hasn't been updated and there are unrelated changes in the PR, but if that's alright with you, feel free to merge.

agibsonccc:
@saudet  I updated the decision to accepted. Regarding the decision, it was updated to use extensions instead of the version. I clarified my wording there to make sure that was more explicit. Thanks for highlighting that.

saudet (approved):
There's still some unrelated changes, but looks harmless enough.

agibsonccc:
@saudet  ah yeah, that's related to another PR. Intellij has been doing some auto spacing/indent. That will be cleaned up in another PR. Thanks!

PR #9319, opened 2021-05-23
libnd4j. Ctc beam search decoder.

quickwritereader:

Ctc beam search decoder.

## How was this patch tested?

unit tests

## Quick checklist


libnd4j/include/ops/declarable/generic/decoder/ctc_beam_op.cpp
```diff
@@ -0,0 +1,111 @@
...
+    auto logitShapeInfo =  inputShape->at(0);
+    auto sequenceShapeInfo =  inputShape->at(1);
+    auto arg_size =  block.getIArguments()->size();
+
+    int nbest_len = arg_size>2? INT_ARG(2): 1;
+
+    REQUIRE_TRUE(logitShapeInfo[0] ==3 , 0, "Ctc Beam Search: logit Input fails to meet rank requirement {BATCH_LEN, MAX_FRAME_LEN, CLASS_LEN }: %i == 3",
+                 logitShapeInfo[0]);
+
+    int batch_size = shape::shapeOf(logitShapeInfo)[0] ;
```
agibsonccc: Shouldn't this be Nd4jLong?

libnd4j/include/ops/declarable/helpers/cpu/ctcLoss.cpp
```diff
@@ -197,23 +175,17 @@ namespace sd
...
                         auto currentProb = element<IsLogPStrided>(logP, currentInd, elwiseP);
                         // if blank or the same as previous
                         if (s < lenSB - 2 && currentInd != blankIndex && currentInd != element<IsLblStrided>(lbl, ind + 1, elwiseS))
                         {
                             Type bettaS_2 = bettaPrevPtr[s + 2];
-                            cMax = std::max(cMax, bettaS_2);
-                            if (cMax == negInf)
-                                cMax = 0;
-                            bettaPtr[s] = std::log(std::exp(bettaS - cMax) + std::exp(bettaS_1 - cMax) + std::exp(bettaS_2 - cMax)) + cMax + currentProb;
+                            bettaPtr[s] = log_sum_exp(bettaS, bettaS_1, bettaS_2) + currentProb;
```
agibsonccc, line 184: Shouldn't log_sum_exp be an op call? I believe we have this somewhere laying around, that or at least part of a standard helper method?

agibsonccc (changes requested):
Only one other comment is the usual cuda version. Can we map a cudnn equivalent in the same PR?

libnd4j/include/ops/declarable/helpers/cpu/ctcLoss.cpp
quickwritereader: its inline [function](https://github.com/eclipse/deeplearning4j/blob/6ecf944a692498e443b083797bc90d55f32d3c1b/libnd4j/include/ops/declarable/helpers/ctc.h#L67-L90) and also for 2 arguments I'm using substituted implementation. using op call or its helper does not suit. I can use logsumexp as an op probably in beam search for doing normalization outside, but I already added it as fused inner call when user wants to normalize their input as well.

libnd4j/include/ops/declarable/generic/decoder/ctc_beam_op.cpp
quickwritereader: > Only one other comment is the usual CUDA version. Can we map a CUDNN equivalent in the same PR?

I believe there is not any function in CUDNN for that task. plus due to the nature of the algorithm, I believe other libraries did not implement CUDA version, as well. I'll look for this again.

agibsonccc (approved):

PR #9353, opened 2021-06-07
Add support for keras 1d pooling nwhc

### LINKED ISSUE #9349: M1: Parameters of Keras MaxPooling1D layer are incorrectly read
#### Issue Description

I trained a model using Tensorflow/Keras version 2.4.1 in Python. This model contains a MaxPooling1D layer. I then saved the model as the (default) .hdf5 file. Loading this file in Keras and saving its config as json, reveals the following layer:

``` json
{
  "class_name": "MaxPooling1D",
  "config": {
      "name": "max_pooling1d_3",
      "trainable": true,
      "dtype": "float32",
      "strides": [
          836
      ],
      "pool_size": [
          836
      ],
      "padding": "valid",
      "data_format": "channels_last"
  },
  "name": "max_pooling1d_3",
  "inbound_nodes": [
      [
          [
              "conv1d_3",
              0,
              0,
              {}
          ]
      ]
  ]
},
```
Please note that the **data_format** is set to "channels_last" which is the default value in Keras. This corresponds to NHWC. The stride is loaded as `(836,)` in Keras - a 1-tuple holding only the value (836). 

If I now load the model in 1.0.0-M1 using `importKerasModelAndWeights()`, this specific layer is loaded as a `Subsampling1DLayer`. Unfortunately, the parameter values are incorrectly parsed. The `Cnn2dDataFormat` is set to `NCHW`. The stride is loaded as `[836,2]` which leads to incorrect dimensions in the following layer. I believe that `[836,1]` is the correct value for the stride.

When I overwrite these two parameter values in my loaded network, then everything runs as expected.

#### Version Information

Please indicate relevant versions, including, if relevant:

* Deeplearning4j version: 1.0.0-M1
* Tensorflow version 2.4.1 (same error exists for models trained in Tensorflow 1.14)
* Platform information: Ubuntu 20.04

agibsonccc:
Add support for keras 1d pooling nwhc
Fixes #9349 

## How was this patch tested?
Adds test model from end user and asserts that NWHC is propagated.
NWHC was supported by the 2d version but not configured for 1D.

## Quick checklist

- [X ] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.org/eclipse-contributors) page for details

### DIFF
--- a/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/SubsamplingLayer.java
+++ b/deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/SubsamplingLayer.java
@@ -102,7 +102,9 @@ public class SubsamplingLayer extends NoParamLayer {
         this.convolutionMode = builder.convolutionMode;
         if (builder instanceof Builder) {
             this.dilation = ((Builder) builder).dilation;
-            this.cnn2dDataFormat = ((Builder) builder).dataFormat;
+            this.cnn2dDataFormat = builder.cnn2DFormat;
+        } else {
+            this.cnn2dDataFormat = builder.cnn2DFormat;
         }
         this.pnorm = builder.pnorm;
         this.eps = builder.eps;
@@ -239,7 +241,6 @@ public class SubsamplingLayer extends NoParamLayer {
          * Dilation for kernel
          */
         private int[] dilation = new int[] {1, 1};
-        protected CNN2DFormat dataFormat = CNN2DFormat.NCHW;
 
         public Builder(PoolingType poolingType, int[] kernelSize, int[] stride) {
             super(poolingType, kernelSize, stride);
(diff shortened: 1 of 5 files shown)


deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/SubsamplingLayer.java
```diff
@@ -102,7 +102,9 @@ protected SubsamplingLayer(BaseSubsamplingBuilder builder) {
         this.convolutionMode = builder.convolutionMode;
         if (builder instanceof Builder) {
             this.dilation = ((Builder) builder).dilation;
-            this.cnn2dDataFormat = ((Builder) builder).dataFormat;
+            this.cnn2dDataFormat = builder.cnn2DFormat;
+        } else {
+            this.cnn2dDataFormat = builder.cnn2DFormat;
```
treo: why an if/else here? you are doing the same thing, put it before this if clause.

treo (changes requested):

deeplearning4j/deeplearning4j-nn/src/main/java/org/deeplearning4j/nn/conf/layers/SubsamplingLayer.java
agibsonccc: Ah, updated. Originally there was an if for pooling 2d since it was missing for the subsampling 1d layer. Now I just added it as a general property. That was more just an artifact. Updated.

treo (approved):

PR #9360, opened 2021-06-09
[fix]: bugfix choice(INDArray source, INDArray probs, int numSamples)

Romira915:
## Summary

Fixed a bug that occurred when using the `choice(INDArray source, INDArray probs, int numSamples)` function (For example, when you declare an `INDArray` with double type.) because the data type of `source` was not taken into account.


agibsonccc:
@raymondlam  looks good! Do you mind signing the eclipse ECA? See more here: https://deeplearning4j.konduit.ai/getting-started/contribute/eclipse-contributors

Romira915:
The signature seems to have been processed correctly. Thank you.

agibsonccc (approved):

agibsonccc:
Merged, great job!

PR #9372, opened 2021-06-15
cudnn lstm forward inference and other fixes

quickwritereader:

ctcLoss: refactored to use new scoped Descriptors in cudnn call

Cudnn Lstm:
Forward inference implemented using v6, and v8 (when version > 8.0.1) api calls.
As our Cuda Lstm implementation has 1 layer. Cudnn implementation was implemented for 1 physical layer
Cudnn helper restrictions:
 - all NDArrays should be the same type
 - dataFormat should be 0 or 1
 - only unidirectional (directionMode == 0) and bidirectional concat (directionMode == 3)
 - no peephole connection
 - Clipping is allowed for cudnn version >= 7.2.1
 - SeqLen array is allowed for cudnn version >= 8.0.1
 - gateActivation: sigmoid, cellActivation and outputActivation: tanh
 - NDArrays (excluding weight ones, as we have to transpose or permute it) should follow 'c' order and ews()==1

 
ctcBeam fix: run when NdArray allocated on the device 

## How was this patch tested?

unit tests

## Quick checklist

### DIFF
--- a/libnd4j/include/ops/declarable/helpers/impl/ctcBeam.cpp
+++ b/libnd4j/include/ops/declarable/helpers/impl/ctcBeam.cpp
@@ -624,7 +624,6 @@ beamSearch_(const NDArray& logit, const NDArray& sequence_length, NDArray& resul
     const auto strides = logit.stridesOf();
     const auto rank = logit.rankOf();
 
-    const IndexType* len_t_ptr = nullptr;
     uint64_t element_stride_t = 1;
 
     //checks before
@@ -636,19 +635,11 @@ beamSearch_(const NDArray& logit, const NDArray& sequence_length, NDArray& resul
     if (len_c < 1 || max_len_t < 1) return;
     //defaulting blankIndex to the last class if its incorrect or -1
     if (blank_index > len_c || blank_index < 0) blank_index = static_cast<int>(len_c) - 1;
-    if (sequence_length.rankOf() == 1 && sequence_length.shapeOf()[0] == batch_len)
-    {
-        sequence_length.syncToHost();
-        len_t_ptr = sequence_length.bufferAsT<IndexType>();
-        element_stride_t = sequence_length.stridesOf()[0];
-    }
 
     //strides
     auto batch_stride = rank > 2 ? strides[0] : 0;
     auto inc_p = strides[rank - 2];
     auto element_stride = logit.stridesOf()[rank - 1];
-    logit.syncToHost();
-    auto logits_ptr = logit.bufferAsT<Type>();
 
 #if defined(ASSERT_INNER)
     //result_probs should be [batch_len, nbest_len]
@@ -657,12 +648,19 @@ beamSearch_(const NDArray& logit, const NDArray& sequence_length, NDArray& resul
     assert(result_sequences.ews() == 1 && result_sequences.rankOf() == 3 && result_sequences.shapeOf()[0] == batch_len && result_sequences.shapeOf()[1] == nbest_len
             && result_sequences.shapeOf()[2] == max_len_t);
 #endif
+    //as ctcBeam search runs on Cpu we should make NdArray buffers available on the host side as well
+    NDArray::preparePrimaryUse({&result_sequences, &result_probs, &result_sequences_length}, {&sequence_length, &logit});
 
-    NDArray::preparePrimaryUse({&result_sequences, &result_probs, &result_sequences_length}, {});
+    auto logits_ptr = logit.bufferAsT<Type>();
     auto result_seq_ptr = result_sequences.bufferAsT<IndexType>();
     auto result_probs_ptr = result_probs.bufferAsT<Type>();
     auto result_seq_length_ptr = result_sequences_length.bufferAsT<IndexType>();
-
+    const IndexType* len_t_ptr = nullptr;
+    if (sequence_length.rankOf() == 1 && sequence_length.shapeOf()[0] == batch_len)
+    {
+        len_t_ptr = sequence_length.bufferAsT<IndexType>();
+        element_stride_t = sequence_length.stridesOf()[0];
+    }
     const auto  batch_stride_res = result_sequences.stridesOf()[0];
     const auto  inc_res = result_sequences.stridesOf()[1];
     const auto  batch_stride_res_prob = result_probs.stridesOf()[0];
@@ -708,7 +706,7 @@ beamSearch_(const NDArray& logit, const NDArray& sequence_length, NDArray& resul
     };
     samediff::Threads::parallel_for(func, 0, batch_len, 1);
 
-    NDArray::registerPrimaryUse({&result_sequences, &result_probs, &result_sequences_length}, {});
+    NDArray::registerPrimaryUse({&result_sequences, &result_probs, &result_sequences_length}, {&sequence_length, &logit});
     return;
 }
 
--- a/libnd4j/include/ops/declarable/helpers/impl/ctcBeam.cpp
+++ b/libnd4j/include/ops/declarable/helpers/impl/ctcBeam.cpp
@@ -638,6 +638,7 @@ beamSearch_(const NDArray& logit, const NDArray& sequence_length, NDArray& resul
     if (blank_index > len_c || blank_index < 0) blank_index = static_cast<int>(len_c) - 1;
     if (sequence_length.rankOf() == 1 && sequence_length.shapeOf()[0] == batch_len)
     {
+        sequence_length.syncToHost();
         len_t_ptr = sequence_length.bufferAsT<IndexType>();
         element_stride_t = sequence_length.stridesOf()[0];
     }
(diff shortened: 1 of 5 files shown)


libnd4j/include/ops/declarable/helpers/impl/ctcBeam.cpp
```diff
@@ -646,7 +647,7 @@ beamSearch_(const NDArray& logit, const NDArray& sequence_length, NDArray& resul
     auto batch_stride = rank > 2 ? strides[0] : 0;
     auto inc_p = strides[rank - 2];
     auto element_stride = logit.stridesOf()[rank - 1];
```
agibsonccc, line 642: Mind adding comments to each of these so someone knows why this was called? Mainly for external contributors or people just reading the implementation.

libnd4j/include/ops/declarable/platform/cudnn/ctcloss.cu
```diff
@@ -85,17 +69,16 @@ namespace platforms {
...
-        cudnnTensorDescriptor_t gradsDesc = nullptr;
-        callCudnnIfNoErr(err, cudnnCreateCTCLossDescriptor, &ctcLossDesc);
-        callCudnnIfNoErr(err, cudnnSetCTCLossDescriptorEx, ctcLossDesc, CUDNN_DATA_FLOAT, CUDNN_LOSS_NORMALIZATION_SOFTMAX, CUDNN_PROPAGATE_NAN);
-        callCudnnIfNoErr(err, cudnnCreateTensorDescriptor, &probsDesc);
-        callCudnnIfNoErr(err, cudnnSetTensorNdDescriptor, probsDesc, cudnnDataType(probs.dataType()), probs.rankOf() , dims, strides);
-        if(!grads.isEmpty()){
+        CTCLossDesc  ctcLossDesc;
+        CudnnTensor probsDesc, gradsDesc(false);
+        bool calcGrads = !grads.isEmpty();
+        ctcLossDesc.set( err, CUDNN_DATA_FLOAT, CUDNN_LOSS_NORMALIZATION_SOFTMAX, CUDNN_PROPAGATE_NAN);
```
agibsonccc: Any reason for the hard coded data type here? Shouldn't we have a conversion for the known data type? Either way, a comment here would be good as well.

libnd4j/include/ops/declarable/platform/cudnn/cudnnUtils.h
```diff
@@ -60,9 +63,325 @@ namespace platforms {
...
+                            cudnnRNNInputMode_t inputMode,
+                            cudnnDirectionMode_t dirMode,
+                            cudnnRNNMode_t cellMode,
+                            cudnnRNNAlgo_t algo,
+                            cudnnDataType_t mathPrec,
+                            int32_t hiddenSize,
+                            int32_t numLayers,
+                            cudnnDropoutDescriptor_t dropoutDesc,  bool use_tensor_op=false){
+
+    cudnnStatus_t err = CUDNN_STATUS_SUCCESS;
```
agibsonccc, line 295: Mind adding some declarations here so we know what versions of methods are being invoked? nd4j_debug should be fine here

libnd4j/tests_cpu/layers_tests/DeclarableOpsTests13.cpp
```diff
@@ -1367,7 +1367,8 @@ TEST_F(DeclarableOpsTests13, lstmLayer_4) {
     hI({1,2, 0,0, 0,0}) = -1;
     cI({0,1, 0,0, 0,0}) = 2;
     cI({1,2, 0,0, 0,0}) = -2;
-
+    hI.printIndexedBuffer("hI");
+    cI.printIndexedBuffer("cI");
```
agibsonccc: Debugging code we should get rid of?

agibsonccc (changes requested):

libnd4j/include/ops/declarable/helpers/impl/ctcBeam.cpp
quickwritereader: my bad, I do not know why I did not put it as a derived datatype for the future. that is a lesser confusion.
 
we have a CTC loss (maybe in the future, cudnn-they consider other types as well) restricted (only for Float32 and input32). I'm thinking of refactoring ctc Loss again ( it needs if def sections and proper comments so it was not confusing ). and also adding v8 api calls .

libnd4j/include/ops/declarable/platform/cudnn/ctcloss.cu
quickwritereader: my bad, I do not know why I did not put it as a derived datatype for the future. that is a lesser confusion.

we have a CTC loss (maybe in the future, cudnn-they consider other types as well)  is restricted (only for Float32 and input32). I'm thinking of refactoring ctc Loss again ( it needs if def sections and proper comments so it was not confusing ). and also adding v8 api calls .

libnd4j/include/ops/declarable/platform/cudnn/cudnnUtils.h
quickwritereader: I already wrote nd4j_debug that informs it will use v6 but it's in the caller function.

I tried to keep Scoped descriptors as simple as I can. Plus as history shows Cudnn side when upgrading they are not hesitant to change method signatures, introduce new functions, and so on. As they had newly introduced functions in v7 which were deprecated in v8. so I had to ignore some of them to make the code cleaner.
when CUDNN < 8.0.1 it calls v6 (CUDNN version 6) API ( but still allowing some functionality (for example clipping) by comparing against some versions).
for > 8 I'm using new v8 API calls. 
What's good with v8 they simplified/merged API calls, now most argument-based arrays are allocated on the device ( probably for the Cuda graph thing).

libnd4j/include/ops/declarable/helpers/impl/ctcBeam.cpp
quickwritereader: I refactored those (syncToHost calls) by adding them inside NDArray::preparePrimaryUse.  and commented

libnd4j/include/ops/declarable/platform/cudnn/ctcloss.cu
quickwritereader: now it's determined by input type. but still, "Cudnn CTC loss" is restricted (only float32). I'll think of relaxing it and adding the conversion.

quickwritereader: I'll address all these next time when I'll refactor CUDNN again.

libnd4j/tests_cpu/layers_tests/DeclarableOpsTests13.cpp
quickwritereader: removed

libnd4j/include/ops/declarable/platform/cudnn/ctcloss.cu
```diff
@@ -72,8 +72,9 @@ namespace platforms {
         CTCLossDesc  ctcLossDesc;
         CudnnTensor probsDesc, gradsDesc(false);
         bool calcGrads = !grads.isEmpty();
-        ctcLossDesc.set( err, CUDNN_DATA_FLOAT, CUDNN_LOSS_NORMALIZATION_SOFTMAX, CUDNN_PROPAGATE_NAN);
-        probsDesc.set( err, cudnnDataType(probs.dataType()), probs.rankOf() , dims, strides);
+        auto cudnnType = cudnnDataType(probs.dataType());
```
agibsonccc, line 75: Could we still explain why we're converting the cudnn type? Why are we forced to do that? A simple link to the cudnn docs with the relevant section is fine. It could be something like: cudnn only supports xyz data types, users may try to pass in abc data types.

It's mainly because we have a trade off here of either throwing an error for it not being the exact type or automatically doing a cast. Whenever you automatically do something, that means it's ok to do so. Knowing why would be good for future reference.

agibsonccc (changes requested):

libnd4j/include/ops/declarable/platform/cudnn/ctcloss.cu
quickwritereader: I do not convert from one type to another. that conversion is from our type to CUDNN equivalent type.
 I just allow certain types  in platform_check. with ctc loss I allow float32 adn int32. I will check it in the official documentation. but it could be also undocumented limitation

libnd4j/include/ops/declarable/platform/cudnn/cudnnUtils.h
agibsonccc: The point was more to put this comment you just did in the comments so it's in the code and people don't have to search github to find context  :)

libnd4j/include/ops/declarable/platform/cudnn/ctcloss.cu
agibsonccc: Yeah just make sure to document that. What the conversion is for is important.

agibsonccc (approved):

PR #9487, opened 2021-10-13
Fix samediff-import-api pom file

treo:
This removes the space character in the very first position of the pom.xml file. 

With it present, it breaks the XML parsing capabilities of some more strict tools.


treo:
used the wrong email address and now eca is unhappy. Created another PR with the correct one.

https://github.com/eclipse/deeplearning4j/pull/9488

PR #9589, opened 2022-01-03
Add model hub SDK for interacting with different model zoos

agibsonccc:
This adds 2 ADRs plus a basic proof of concept SDK for model hub: a way of interacting with and pre processing model files from different frameworks.
Initial targets include:
1. Pytorch vision
2. TF Hub
3. Huggingface
4. Onnx model zoo
5. Keras applications

## How was this patch tested?

## Quick checklist


contrib/omnihub/src/omnihub/frameworks/pytorch.py
```diff
@@ -0,0 +1,236 @@
...
+        super().__init__(framework_name,BASE_URL)
+
+
+    def download_model(self, model_path,**kwargs) -> str:
+        model = None
+        if model_path == 'resnet18':
+            height = kwargs.get('height',224)
+            width = kwargs.get('width',224)
+            x = torch.from_numpy(np.ones((1, 3, height, width),dtype=np.float32))
+            model = models.resnet18(pretrained=True,**kwargs)
```
treo: This looks very much like you've hardcoded something that should have been a `dict`. 

In particular, I think you can probably put the default input shape into a dict indexed by the modelname, and use `getattr` on `models` to reference the actual model function.

Or have a dict that holds on to both input shape and model generation with two lambdas.

omnihub/src/main/java/org/eclipse/deeplearning4j/omnihub/api/ModelType.java
```diff
@@ -17,11 +17,21 @@
...
+            default:
+                return null;
+        }
+    }
 
-    fun displayName() = name[0].toString() + name.substring(1).toLowerCase()
-}
\ No newline at end of file
+
+}
```
treo: This change looks out of place. Have you accidentally moved that file and then renamed it or something like that?

treo (changes requested):

omnihub/src/main/java/org/eclipse/deeplearning4j/omnihub/api/ModelType.java
treo: I'm saying this because it is shown as a move and rename:
` .../kotlin/org/nd4j/codegen/api/Namespace.kt → ...deeplearning4j/omnihub/api/ModelType.java ` 

And the Namespace.kt file is no longer in place

agibsonccc: @treo actually namespace isn't used anywhere so I just deleted it. The code compiles just fine without it. The namespace generation also runs just fine without this present. The move was originally because I was trying to adapt some of the codegen infra to the omnihub SDK and just noticed what was unused.

contrib/omnihub/src/omnihub/frameworks/pytorch.py
agibsonccc: @treo ok broke this up a bit and also made the model name generation a bit better. It's still a bit messy but at least reduces the boilerplate a bit.

contrib/omnihub/src/omnihub/frameworks/keras.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        path_split = path.split('/')
+        type = path_split[0]
+        weights_file = path_split[1]
+        include_top = 'no_top' in weights_file
+        if type == 'vgg19':
+            ret = VGG19(include_top=include_top, **kwargs)
+        elif type == 'vgg16':
+            ret = VGG16(include_top=include_top, **kwargs)
+        elif type == 'resnet50':
```
treo, line 58: This still looks like it could be a `dict`.

Maybe even outside the class. That way it would allow for monkey patching to introduce additional model support if necessary.

libnd4j/include/array/NDArray.hXX
```diff
@@ -2685,6 +2685,8 @@ void NDArray::applyTrueBroadcast(sd::BroadcastOpsTuple op, const NDArray &other,
       (op.s == scalar::ReverseDivide && this->isB()))
     throw std::runtime_error("NDArray::applyTrueBroadcast method: you can't divide by bool array !");
 
+  other.printIndexedBuffer("other at apply true broadcast");
```
treo, line 2688: leftover debug message?

treo (changes requested):

PR #9624, opened 2022-02-02
Allows adding a labelsSource as part of the Builder

jcabot:

BagOfWordVectorizer was missing the option to set the labels

## How was this patch tested?

It hasn't been tested but the added code is directly copied from  ParagraphVectors.java that does offer this option.


treo:
Looks good to me. 

Can you also take care of the Eclipse ECA, so we can get this merged?

treo (approved):

PR #9688, opened 2022-05-06
Python4j performance optimizations

### LINKED ISSUE #9595: Python4J Performance Optimizations
Hi,

I took a go at Python4J and compared it to some other integrations: https://github.com/invesdwin/invesdwin-context-python/blob/master/README.md#results

Following findings:
- PythonGIL should be made reusable: https://github.com/cnuernber/libjulia-clj/issues/4#issuecomment-1008055456
- type conversions can be made a lot faster and a None/Null type could be supported: https://github.com/cnuernber/libjulia-clj/issues/4#issuecomment-1008125477
  - don't call PythonObject.isNone() because it is very slow
  - I could not find a way to create a new "None" instance, thus I used a workaround to get a new one allocated from the python side, though this is only relevant when lists have "null" elements. I guess there should be a better way to do this directly via CPython API.

With the new implementation of PythonTypes I was able to get it from ~7k calls per second to more than 100k calls per second.
See implementation here (especially the ModifiedPythonTypes class): https://github.com/invesdwin/invesdwin-context-python/tree/master/invesdwin-context-python-parent/invesdwin-context-python-runtime-python4j/src/main/java/de/invesdwin/context/python/runtime/python4j/internal

grafik

agibsonccc:
Fixes https://github.com/eclipse/deeplearning4j/issues/9595

Noteworthy:
1. Shades netty-util for its fasththreadpool implementation: https://programmer.ink/think/why-netty-s-fastthreadlocal-is-fast.html
2. Adds caching of python objects to access derived variables created from python to prevent double allocation.

## How was this patch tested?

## Quick checklist

- [X ] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.org/eclipse-contributors) page for details


treo:
Can you please add a quick explanation why we need to shade netty?

treo (dismissed):

treo (approved):

PR #9721, opened 2022-06-29
fix concurrency issue in ticketing framework

pza94:

https://github.com/eclipse/deeplearning4j/issues/9534
fix addressing concurrency problems in ticketing framework

## How was this patch tested?

Was not.

## Quick checklist

### DIFF
--- a/libnd4j/include/execution/impl/ThreadPool.cpp
+++ b/libnd4j/include/execution/impl/ThreadPool.cpp
@@ -117,6 +117,8 @@ ThreadPool::ThreadPool() {
 #endif
      */
   }
+  //add an extra ticket to minimize the risk of running out of tickets due to race conditions
+  _tickets.push(new Ticket());
 #endif
 }
 
@@ -154,7 +156,9 @@ Ticket *ThreadPool::tryAcquire(int numThreads) {
   {
     // we lock before checking availability
     std::unique_lock<std::mutex> lock(_lock);
-    if (_available >= numThreads) {
+    //test for both _available and _tickets in order to deal with race conditions caused by the
+    //fact that marking threads as available AND releasing tickets does not happen atomically
+    if (_available >= numThreads && !_tickets.empty()) {
       threaded = true;
       _available -= numThreads;
 


pza94:
this might be sub-optimal, since it doesn't increase the available thread count right away when a thread finishes its task, but at least takes care of the concurrency issues.

Possibly bit more efficient would be, to increment the _available (thread count) variable for all but the last task in the for-loop, and then increment it once more after the ticket is put back on the queue. ?

quickwritereader:
thanks @pza94 
I will recheck that ticket threading logics and also your potential fix

quickwritereader:
alright, I think this fix is feasible.  I checked the source. It seems this thread pool was implemented to achieve   a  **restricted** fork-join model unlike standard OMP

quickwritereader (dismissed):

pza94:
We started using the changes I proposed in production, because the segfaults we were getting, were critical. However, I would like to point out that this really is not a very efficient solution:

Let's say if a ticket acquired 2 threads, and one of them runs for 1 second and the other runs for an hour then both of those threads will be locked down for an hour. In our specific prod use cases this wouldn't happen, so we didn't put much effort into making this efficient, but this definitely could be improved by refactoring the `tryAcquire` method and maybe the `_tickets` queue in the `ThreadPool` class. We didn't want to commit to those changes ourselves.

quickwritereader:
I see. Mostly its used for splitting into equal parts 
.
Also could you specify the device and os in which you were getting segs 
I think maybe somehow its related with atomic variable.
assembly should be checked for atomic variable operations :

--var
++var
var+=n

pza94:
os is ubuntu16
device is x86-64 cpus (various amd and  intel CPUs).

`I think maybe somehow its related with atomic variable.`
I wouldn't think so. In my view the issue is that the 2 operations: incrementing `_available` and putting back the ticket in the queue does not happen atomically. Or in fact just putting back the ticket first and then incrementing should suffice as well, given how the `tryAcquire` method checks for thread-availability.

quickwritereader:
I see. and you mentioned that somehow the tickets are exhausted.

pza94:
That is our theory at least, based on the segfault dump and our notes in the related issue. Couldn't verify it though, feels like a nasty race condition that is not easy to reproduce.

quickwritereader:
yes let assume
we have 2 threads
2 parallel executions.
in that case
there is possibility that the current order might affect null ticket

quickwritereader:
so doubling ticket size can be pretty good solution for such cases.
otherwise there should be condition for checking the ticket queue size and return as nonavailable in the case its exhausted

quickwritereader:
Could you verify those possible fix above 
theoritically ticket being none is possible due to releasing threads before the ticket

quickwritereader:
https://github.com/eclipse/deeplearning4j/blob/573cdfc71684ffd88f1d20c278d181dac1e4f8dd/libnd4j/include/execution/impl/ThreadPool.cpp#L157

`    if (_available >= numThreads) {` 
coudl be replcaed by
`if (_available >= numThreads && _tickets.size()>0) {`

Furthermore, to make such cases happen rarely, an additional ticket could be pushed in the constructor as well. 

. the other solution could be to push twice more ticket

pza94:
We considered doubling the initial ticket count. In theory (in some crazy scenario) I think it's possible that you'll end up running into the same issue, but in practice it would never really happen.

fx. say max_thread_count=1, then you have 2 tickets initially:

* (Java thread1) acquire 1 ticket, that uses 1 thread. when that task is finished it increments `_available` and the OS unschedules the thread before giving back the queue. 
* at this point `_available==1` and ticket count is also 1
* (Java thread2) same as in first point but in a different (Java) thread AND OS hadn't rescheduled `Java thread1`.
* (Java thread3) does will segfault
You gotta be real unlucky for this to happen though.

`checking the ticket queue size and return as nonavailable` seems fair to me

quickwritereader:
@pza94 if possible could you check the above fix.

as I confirm that there can be potential ticket exhaustion.
the current fix is pretty well. that's why I accepted, as this thread pool is acting like fork-join style.
but if the above fix can work it can help with the potential sub-optimal case you mentioned

pza94:
> 

this is the best probably

pza94:
> https://github.com/eclipse/deeplearning4j/blob/573cdfc71684ffd88f1d20c278d181dac1e4f8dd/libnd4j/include/execution/impl/ThreadPool.cpp#L157
> 
> ` if (_available >= numThreads) {` coudl be replcaed by `if (_available >= numThreads && _tickets.size()>0) {`
> 
> Furthermore, to make such cases happen rarely, an additional ticket could be pushed in the constructor as well.
> 
> . the other solution could be to push twice more ticket

this is the best probably

quickwritereader:
yes, let's see if it helps.

quickwritereader:
@pza94 could you apply the discussed fix in the current Pr , if it works
otherwise just inform us. we will merge this as a fix.
thanks. well done

pza94:
sure, I'll submit it later today

pza94:
are the comments sufficient/too much/explain well enough the concurrency issues?

quickwritereader (approved):

quickwritereader:
yes, they are.
thanks. well done

quickwritereader (approved):

pza94:
note: if you want to run an extra mile with testing this you can try (on linux) forcing a context switch before releasing the ticket and have many java threads request operations that use only a few (or preferably just one) thread, like so:
```
void Ticket::waitAndRelease() {
  for (uint32_t e = 0; e < this->_acquiredThreads; e++) {
    // block until finished
    _interfaces[e]->waitForCompletion();

    // mark available
    _interfaces[e]->markAvailable();

    // increment availability counter
    ThreadPool::getInstance().release();
  }
  sched_yield();
  // return this ticket back to the pool
  ThreadPool::getInstance().release(this);
}
```
I can't really afford to play with this myself, but I think this should trigger the behavior I described. And then test that it doesn't happen with the changes proposed here. Up to you

agibsonccc (approved):

PR #9901, opened 2023-01-31
Allow user to define UDFs in samediff/nd4j

### LINKED ISSUE #9898: No way to override the standard Op BP algorithm
#### Overriding the standard Op BP algorithm
- **Expected behavior** : if the existing Op's BP is overridden with a custom one, the latter should be used in the graph, not the original one.
- **Encountered behavior** :  if the existing Op's BP is overridden with a custom one, the original BP is being used thus fully ignoring any custom overrides

#### Version Information

* Deeplearning4j version - M 2.1
* Platform information - Windows 11
* CUDA version - N/A
* NVIDIA driver version - N/A

#### Additional Information

The example is the Op `CompareAndSet`. I tried to overide the existing BP algorithm with my own in the same way as I did for Gather Op, but with `CompareAndSet `it didn’t work. Unlike `Gather Op`, `CompareAndSet`' custom Op instance is being ignored if it's inside `ImportClassMapping` `OP_NAME_MAP`. The difference could be caused by the `OpType `- Gather Op has the type `CUSTOM`, `CompareAndSet `- `TRANSFORM_`...

I tried even replacing this Op's instance with my custom one directly in `SameDiff`, but I got stuck in `FlatBuffersMapper#fromFlatNode` because it creates this Op instance based on a **hardcoded numerical value** of `OpType `thus bringing the original version to life. That one is incorrectly cloned (I have no idea why cloning is needed) and some props are missing so the execution fails. Overriding the props setup method doesn’t help because the original version is hardcoded in `LegacyOpMapper`. Changing this type to a `CUSTOM ` in my custom class also doesn’t help because the execution fails on another part then (which makes sense because this Op is the part of the `SameDiff `Ops).

_P.S. In general it would nice to have a high-level possibility to use custom Op (or their BPs) implementations in a "plug-and-play" fashion. There could be something specific to a project but not required to be present for the whole framework. Right now it’s not taken into account even architecturally (e.g. private fields in all Ops instead of protected etc.)._

agibsonccc:
Allows definition of UDFS, fixes: https://github.com/deeplearning4j/deeplearning4j/issues/9898

Users can now define custom operations while being properly integrated with samediff and serialization.
This also adds annotation scanning for registration with the samediff registered ops infrastructure.

## How was this patch tested?

## Quick checklist

- [ X] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.konduit.ai/multi-project/how-to-guides/contribute/eclipse-contributors) page for details


treo:
I'll take a closer look later today, but in the meantime, I guess you should probably take a look at the auto generated files and the now missing license headers.

treo (approved):

PR #9936, opened 2023-03-15
Ensure PCA doesn't replace input data

### LINKED ISSUE #9934: PCA.pca(INDArray A, int nDims, boolean normalize) does not return correct principal components
#### Issue Description
Commit [#9933](https://github.com/deeplearning4j/deeplearning4j/issues/9931 ) fixed the transposed factor, but didn't address the fact that input matrix is overwritten with garbage. So, now methods:
`public static INDArray pca(INDArray A, int nDims, boolean normalize)`
and
`public static INDArray pca(INDArray A, double variance, boolean normalize)`
return incorrect principal factors because they call pca_factor and immediately do:
`return A.mmul(factor);`
Since A is no longer the original input matrix (it has been modified, I suspect as a side effect of some operation), the returned pca matrix is wrong.
All this is for native backend. CUDA backend works properly.
The proper fix would be to make sure that call:
`Nd4j.getBlasWrapper().lapack().gesvd(A, s, null, VT);`
does not modify A and does really return VT, not V as it does now. The same call behaves correctly for CUDA backend.

agibsonccc:
Fixes #9934 

Ensures input data isn't modified during PCA call.

## How was this patch tested?

## Quick checklist

- [X ] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.konduit.ai/multi-project/how-to-guides/contribute/eclipse-contributors) page for details

### DIFF
--- a/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
+++ b/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
@@ -150,7 +150,7 @@ public class PCA {
      * @return the reduced parameters of A
      */
     public static INDArray pca(INDArray A, int nDims, boolean normalize) {
-        INDArray factor = pca_factor(A, nDims, normalize);
+        INDArray factor = pca_factor(A.dup(), nDims, normalize);
         return A.mmul(factor);
     }
 
--- a/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
+++ b/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
@@ -140,7 +140,7 @@ public class PCA {
      *
      * INDArray Areduced = A.mmul( factor ) ;
      * INDArray Aoriginal = Areduced.mmul( factor.transpose() ) ;
-     * 
+     *
      * }
      * </pre>
      *
@@ -150,7 +150,7 @@ public class PCA {
      * @return the reduced parameters of A
      */
     public static INDArray pca(INDArray A, int nDims, boolean normalize) {
-        INDArray factor = pca_factor(A.dup(), nDims, normalize);
+        INDArray factor = pca_factor(A, nDims, normalize);
         return A.mmul(factor);
     }
 
@@ -169,7 +169,7 @@ public class PCA {
      * @return the reduced feature set
      */
     public static INDArray pca_factor(INDArray A, int nDims, boolean normalize) {
-
+        A = A.dup();
         if (normalize) {
             // Normalize to mean 0 for each feature ( each column has 0 mean )
             INDArray mean = A.mean(0);
@@ -231,7 +231,7 @@ public class PCA {
      * To use the returned factor: multiply feature(s) by the factor to get a reduced dimension
      *
      * INDArray Areduced = A.mmul( factor ) ;
-     * 
+     *
      * The array Areduced is a projection of A onto principal components
      *
      * @see pca(INDArray, double, boolean)
@@ -242,6 +242,7 @@ public class PCA {
      * @return the matrix to mulitiply a feature by to get a reduced feature set
      */
     public static INDArray pca_factor(INDArray A, double variance, boolean normalize) {
+        A = A.dup();
         if (normalize) {
             // Normalize to mean 0 for each feature ( each column has 0 mean )
             INDArray mean = A.mean(0);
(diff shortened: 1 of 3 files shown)


nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
dpetronijevic, line 1: It is better to duplicate the input matrix (A) inside the pca_factor body and operate on the copy. This change doesn't fix  metod public static INDArray pca(INDArray A, double variance, boolean normalize). If input duplication is done inside the pca_factor, both variants of the pca method are fixed.

agibsonccc: @dpetronijevic Great point! I've pushed a commit to fix it in all the different pca_factor methods.

treo (approved):

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
```diff
@@ -169,7 +169,7 @@ public static INDArray pca(INDArray A, int nDims, boolean normalize) {
      * @return the reduced feature set
      */
     public static INDArray pca_factor(INDArray A, int nDims, boolean normalize) {
-
+        A = A.dup();
```
dpetronijevic, line 172: I haven't had the time to review this earlier, but I am afraid that this is not going to solve the issue where A is modified after the call further down in:
`Nd4j.getBlasWrapper().lapack().gesvd(A, s, null, VT);`
`A = A.dup();` just duplicates A and assigns it back to A, so after the call to gesvd A will be modified.
The only way to leave A unmodified after the call to gesvd is to do the following:
`Nd4j.getBlasWrapper().lapack().gesvd(A.dup(), s, null, VT);`
That way gesvd operates and modifies the duplicate and leaves A intact for further use.

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
```diff
@@ -242,6 +242,7 @@ public static INDArray pca(INDArray A, double variance, boolean normalize) {
      * @return the matrix to mulitiply a feature by to get a reduced feature set
      */
     public static INDArray pca_factor(INDArray A, double variance, boolean normalize) {
+        A = A.dup();
```
dpetronijevic, line 245: The same applies here as in the above comment

platform-tests/src/test/java/org/eclipse/deeplearning4j/nd4j/linalg/blas/BlasTests.java
```diff
@@ -65,6 +64,25 @@ public void pcaTest(Nd4jBackend backend) {
...
+    @MethodSource("org.nd4j.linalg.BaseNd4jTestWithBackends#configs")
+    public void pcaTest(Nd4jBackend backend) {
+        double[][] inputArray = { { 1.0, 2.0, 3.0 }, { 11.0, 12.0, 13.0 }, { 21.0, 22.0, 23.0 }, { 31.0, 32.0, 33.0 } };
+        double[][] assertion = new double[][]{
+                {-0.55332, -0.72606, 0.40825},
+                {      -0.57703 ,-0.01936 ,-0.81650},
+                {-0.60073, 0.68735, 0.40825 }
+        };
+        INDArray assertArr = Nd4j.create(assertion);
+        INDArray inputMatrix = Nd4j.create( inputArray );
```
dpetronijevic, line 77: In order to verify that PCA.pca_factor does not modify inputMatrix  it would be good to add to test the following:
`INDArray duplicateInput = inputMatrix.dup();`
before the calls to PCA.pca_factor and PCA.pca and then assert after the calls:
`assertEquals(duplicateInput, inputMatrix);`

dpetronijevic (commented):
Sorry one more time for reviewing this late, but as the saying goes, better late than never. I left the comments inside the changes at appropriate locations

nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dimensionalityreduction/PCA.java
treo: > A = A.dup(); just duplicates A and assigns it back to A, so after the call to gesvd A will be modified.

Yes, but no. As the reference to A gets changed, it means that A inside the method gets changed, but the A that was originally passed to pca_factor will not be changed. 

What you're thinking off would happen if `A.assign(A.dup())` was used instead.

dpetronijevic: You are right, my paranoia from C-like languages got better of me. I never assign to input parameters even when it is OK to do it like here.

PR #10073, opened 2024-08-02
Add nd4j-log-analyzer

agibsonccc:

Adds bytebuddy based log analyzer for comparing runs of operations across different
versions of dl4j.
## How was this patch tested?

Manually

## Quick checklist

- [X ] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.konduit.ai/multi-project/how-to-guides/contribute/eclipse-contributors) page for details


treo (dismissed):
Looks like a plan to me. 

Good idea to use the java agent capabilities for that. I'd expect that it wouldn't be used in production unless someone needs to figure out a specific problem, so the actual overhead shouldn't matter as much, unless it changes the performance profile of the application under test too much.

Edit: Oh, I see this may be supposed to be more than just the ADR? The last commit removes the entire implementation, is that intended?

agibsonccc:
@treo I readded it. I meant to just move the contents of the directory up one level.

treo (commented):
Wouldn't it make sense to use a logger instead of println here?

agibsonccc:
Mainly trying to minimize dependencies. When trying to inject this in to master and M2.1 when I was using it for baseline comparison I ran in to dependency issues. A logging implementation would just make this more error prone.

treo (approved):

PR #10162, opened 2024-12-18
Remove shape descriptor creation when caching shape buffers

agibsonccc:

This PR migrates shape buffers to use a new trie based approach for storing constant shape buffers.
This also introduces a new threadlocal based mutex for the shape buffer allowing for better local caching of shape information.
By migrating to tries from a standard map we removed all new SHapeDescriptor calls
and directly just use shape info as it is. This removes a lot of redundant code and allocations.

This substantially reduced overhead especially on smaller operations.
## How was this patch tested?

Via benchmarks
## Quick checklist

- [ X] Eclipse Contributor Agreement signed, and signed commits - see [IP Requirements](https://deeplearning4j.konduit.ai/multi-project/how-to-guides/contribute/eclipse-contributors) page for details


contrib/benchmarking_nd4j/pom.xml
```diff
@@ -85,6 +88,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.1</version>
                 <configuration>
+                    <source>11</source>
+                    <target>11</target>
```
treo: ideally you'd use the following here instead of half hard-coding values.
```
<source>${java.version}</source>
<target>${javac.target}</target>
<compilerVersion>${java.compiler.version}</compilerVersion>
```

libnd4j/include/array/impl/ShapeDescriptor.cpp
```diff
@@ -724,16 +719,89 @@ ShapeDescriptor  * ShapeDescriptor::paddedBufferDescriptor(const DataType type,
...
 size_t hash<sd::ShapeDescriptor>::operator()(sd::ShapeDescriptor k) const {
-  auto res = std::hash<char>()(k.order());
-  res ^= std::hash<int>()((int)k.dataType()) + 0x9e3779b9 + (res << 6) + (res >> 2);
-  sd::LongType * shape_strides = k.shape_strides();
-  auto ptr = shape_strides;
-  //dont include strides if its' ews==1
-  int stop =  k.rank() * 2;
-  for (int j = 0; j < stop; j++) {
-    res ^= std::hash<sd::LongType>()(ptr[j]) + 0x9e3779b9 + (res << 6) + (res >> 2);
+  // Check cache first
```
treo, line 722: You've got the hash implementation in a couple of places, if it can be reused, try to reuse it.

treo (changes requested):
An ADR explaining a bit more of the rationale behind this change would be nice. 

Also I've left a couple of comments.

contrib/benchmarking_nd4j/pom.xml
```diff
@@ -35,7 +35,10 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <jmh.version>1.37</jmh.version>
-        <javac.target>11</javac.target>
+        <source>${java.version}</source>
+        <target>${javac.target}</target>
+        <compilerVersion>${java.compiler.version}</compilerVersion>
+
```
treo: you applied the changes at the wrong place. Those variable usages should be at the maven-compiler-plugin config.

treo (changes requested):

contrib/benchmarking_nd4j/pom.xml
agibsonccc: Ah..yeah I missed what was modified. The properties were in the middle. I moved them to the top for readability so as not to miss that again..Thanks.

treo (approved):

PR #10243, opened 2025-09-27
Maven version updates for lombok for java 25 support

agibsonccc:
Updates lombok and various versions on various maven modules.


treo (changes requested):
This PR also has the mistakenly commited submodules.

treo (approved):

platform-tests/pom.xml
```diff
@@ -447,7 +457,7 @@
         </dependency>
 
         <dependency>
-            <groupId>org.eclipse.deeplearning4j</groupId>
+            <groupId>org.deeplearning4j</groupId>
```
treo, line 460: Before you merge, this change looks a bit suspicious.

agibsonccc: This was just due to me not updating test resources yet. I'll handle tests in a subsequent update.

PR #10252, opened 2025-09-27
Better type handling with random generator.

agibsonccc:
This adds better type checking in combination with selective rendering.
Only certain signatures will be output now when we actually have the types compiled in to the binary.

Also handles various type warnings when casting for certain scalar outputs.


libnd4j/include/graph/RNG.h
```diff
@@ -0,0 +1,11 @@
...
+*
+*
+* This program and the accompanying materials are made available under the
+* terms of the Apache License, Version 2.0 which is available at
+* https://www.apache.org/licenses/LICENSE-2.0.
+*
+*  See the NOTICE file distributed with this work for additional
+*  information regarding copyright ownership.
+* Unless required by applicable law or agreed to in writing, software
+* distributed under
```
treo: Something must have gone wrong with selecting the chunks to commit. This file is just an unfinished header.

treo (changes requested):

treo (approved):

PR #10285, opened 2025-09-27
NDArray flatarray serialization improvements, misc spacing changes

agibsonccc:
This introduces the changes for flatbuffers adding samediff subgraphs to the flatbuffers metadata itself.


treo (changes requested):
This PR seems to be all over the place. 

There are two submodules that I'm quite certain were added by mistake.

The Nd4j.initWithBackend is full of debugging prints, instead of using a logger. 

And it adds an entirely unrelated onnx tools file.

agibsonccc:
Let me break this one up more.

